PR #87, opened 2015-01-05
added the swift version for LegendSample and updated the folder structur...

singhgag:
...e for objC sample


dg0yal:
This seems to be crashing on an iPad when i try to display the legend VC

singhgag:
fixed the issue!

PR #96, opened 2015-01-07
added the swift version for DownloadTileCache sample and update objc fol...

singhgag:
...der structure


DownloadTileCacheSample/swift/DownloadTileCache/ViewController.swift
```diff
@@ -0,0 +1,252 @@
...
+        //Alert user of error
+        UIAlertView(title: "Error", message: error.localizedDescription, delegate: nil, cancelButtonTitle: nil).show()
+    }
+    
+    func layerDidLoad(layer: AGSLayer!) {
+        if layer == self.tiledLayer {
+            //Initialize UIStepper based on number of scale levels in the tiled layer
+            self.levelStepper.value = 0
+            self.levelStepper.minimumValue = 0
+            self.levelStepper.maximumValue = Double(self.tiledLayer.tileInfo.lods.count)
```
dg0yal: This should be count-1
See https://github.com/Esri/arcgis-runtime-samples-ios/pull/91

singhgag: Updated the code!

dg0yal:
1) Sample crashes when you attempt to download the cache for 9 scale levels at the starting extent
2) Sample crashes when you attempt to download the cache for 1 scale level after zooming in all the way

singhgag:
fixed the issue causing the crash in both scenarios.

dg0yal:
Yup, those are working now.

3) The estimate results do not include the unit of measurement (_KB_ or _MB_ for size, _Tiles_ for count)
screen shot 2015-01-07 at 4 14 29 pm. Both in the label, and in the estimate results alert view.

singhgag:
added units for both objC and swift

dg0yal:
The objc sample already has pretty byte count (properly formatted as KB or MB depending upon the size using NSByteCountFormatter). It'll be good to implement something similiar in swift

singhgag:
Done

PR #100, opened 2015-01-09
Gagan geometry sample swift

singhgag:


dg0yal:
@singhgag It doesn't work fine on 7.1 simulator. The left pane takes up all the visible area.

singhgag:
@dg0yal seemed like an apple bug.

PR #107, opened 2015-03-13
added logic for local notification permissions in DownloadTileCacheSampl...

singhgag:
...e and OfflineFeatureEditingSample


dg0yal:
@singhgag  This won't compile when building with iOS 7 SDK (which is still supported at 10.2.4)
Can you add conditional check?

singhgag:
@dg0yal the checks are already in. It was working for me when I tested. 
Is the problem specific to swift or objC?

dg0yal:
Your project must be compiling with iOS 8 SDK. It won't compile with iOS 7 SDK.
(I don't mean running on iOS 7 or 8 simulator, I mean the iOS SDK that your project builds with)

singhgag:
@dg0yal done

PR #109, opened 2015-03-25
added the TOC swift sample, now using AGSMapContentsTree class

singhgag:


dg0yal:
Image.png needs to stay in the top level folder (next to readme.md). Your pull request is moving it to the objective-c folder.

dg0yal:
And also there appears to be bug with collapsing nodes. To repro, expand Census top level layer, then expand Census Block Points sublayer. Now collapse top level Census layer. You'll notice that a couple of legends of the sublayer are left behind 

image

dg0yal:
Everything else looks good. Good job. I like the way the classes are architected. Very easy to follow. A couple of suggestions - can you reduce the size of the TOCViewController, specifically the width? It appears to cover most of the map on iPad. Also, could you change the text of the button "Content" to "Layers" ? That's it.

singhgag:
@dg0yal nice catch with the collapsing issue. I have added the changes. Should be good now.

PR #127, opened 2015-09-24
Update Xcode version support

maryharvey:
Updated the Xcode version support for Quartz beta.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,9 @@ For Objective-C :
 * iOS 8 SDK (or higher)
 
 For Swift : 
-* XCode 6.3 (or higher)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,8 +12,9 @@ For Objective-C :
 * iOS 8 SDK (or higher)
 
 For Swift : 
-* XCode 6.3 (or higher)
+* XCode 7 (or higher)
 * iOS 8 SDK (or higher)
```
dg0yal: We need to change iOS 8 SDK to iOS 9 SDK as that is what XCode 7 comes with.

PR #130, opened 2015-10-15
added the display layer view state sample

singhgag:
https://devtopia.esri.com/runtime/cocoa-common-sdk/issues/295


maryharvey:
@singhgag 
* When you zoom out past the extent of the WorldTimeZones the map goes black and you can just see the faciltites. Why does the Layer Status still say Active?
* The code is not displayed in the code window
* Can you place the sample next to the Display drawing status sample?
* Some immutable values are never used. Can you check the code?
* In the splitViewController function you have some commented code - Can you check the code here and clarify the comments if necessary

singhgag:
@maryharvey 


> When you zoom out past the extent of the WorldTimeZones the map goes black and you can just see the faciltites. Why does the Layer Status still say Active?

Can you create an issue for this?

> The code is not displayed in the code window

Fixed

> Can you place the sample next to the Display drawing status sample?

Done

> Some immutable values are never used. Can you check the code?

Fixed

> In the splitViewController function you have some commented code - Can you check the code here and clarify the comments if necessary

I did remove some commented code but not sure if that was what you meant. If not, can you provide specific file names?

maryharvey:
@singhgag I have tested the sample with the released Quartz Beta iOS SDK. An error is getting displayed every time the ```self?.tableView.reloadData()``` is called. Can you take a look at it?
```
2016-01-07 13:25:59.555 arcgis-ios-sdk-samples[11719:635906] ArcGIS Runtime Error 15 / 16. Can not call this method in this context - Can only call this method on a loaded table.
2016-01-07 13:25:59.871 arcgis-ios-sdk-samples[11719:635906] ArcGIS Runtime Error 15 / 16. Can not call this method in this context - Can only call this method on a loaded table.
```

singhgag:
@maryharvey This error is shown when I try to get the name of the feature layer and the table is not loaded yet. Its just a log and I already have a check for it in the code.
```
    func tableView(tableView: UITableView, cellForRowAtIndexPath indexPath: NSIndexPath) -> UITableViewCell {
        let cell = tableView.dequeueReusableCellWithIdentifier("LayerStatusCell")!
        cell.backgroundColor = UIColor.clearColor()
        let layer = self.map.operationalLayers[UInt(indexPath.row)] as! AGSLayer

>>>>>        cell.textLabel?.text = layer.name ?? "Layer \(indexPath.row)"
        cell.detailTextLabel?.text = self.viewStatusArray[indexPath.row]
        
        return cell
    }
```

maryharvey:
We are accepting that this Runtime message is thrown for Beta 2. This is a known Issue and we will follow this up for Final.

PR #142, opened 2016-02-18
Added two samples for routing

singhgag:


maryharvey:
@singhgag tested and comments added to repo

PR #146, opened 2016-02-28
Remove flashing of callout

dg0yal:
For offline geocoding sample while reverse geocoding on tap-hold-and-move.


singhgag:
Merged

PR #161, opened 2016-04-04
Checking layer load status before accessing the name property. The ru…

singhgag:
…ntime was printing an error otherwise.


maryharvey:
@maryharvey Is this notification reaching Mary?

PR #166, opened 2016-04-12
removed the unnecessary dispatch asyncs

singhgag:


singhgag:
PS: By removing the dispatch async from the `startWithStatusHandler`, issues with UI updates is quite visible. For example, the progress HUD won't update as and when required. And at times exceptions get thrown by apple stating that you should be updating the UI in a thread other than main (no crashes though).

PR #167, opened 2016-04-18
New sample for tiled layer render mode

singhgag:


maryharvey:
@singhgag Can you make the following changes?
The default rendermode is Scale so make the first item in the segmentcontroller Scale and make sure it is selected.
Change the label to Aesthetic

The description needs more information - I will supply that in another PR

singhgag:
>The default rendermode is Scale so make the first item in the segmentcontroller Scale and make sure it is selected.

I forgot to mention, the default mode for tiled layer is supposed to be Aesthetic. There is an issue open right now. Once thats fixed it should work as designed. Right now you have to switch to Scale then back to Aesthetic to make it work

PR #168, opened 2016-04-19
Added new sample

singhgag:
Control graphic draw index


maryharvey:
Sample works great but this is confusing...
```
        //subtracting 10, since the buttons are tagged from 10 - 13
        self.graphics[sender.tag-10].zIndex = self.drawIndex
```
Why not label the buttons as 0,1,2,3
In addGraphics add comments such as //blue marker is self.graphics[0] etc.
In buttonAction... how abut this
```
        //the button's tag value specifies which graphic to re-index
        //for example, a button tag value of 1 will move self.graphics[1] - the red marker
        self.graphics[sender.tag].zIndex = self.drawIndex
```

maryharvey:
Can the scale be more zoomed in as the layer looks a bit pixelated?

maryharvey:
@singhgag PR has conflicts can you resubmit?

singhgag:
@maryharvey Fixed

PR #170, opened 2016-04-26
Added new sample for identify layers

singhgag:


maryharvey:
Tested with Quartz Beta2

PR #192, opened 2016-07-13
Added more samples

singhgag:
- Map loaded
- Feature layer (geodatabase)


maryharvey:
@singhgag comments
Map Loaded - Info - "This sample shows you the map's load status. The LoadStatus class is considered loaded when any of the following are true:", Remove an an from second bullet, :"The sample uses Key-Value Observing on the AGSMap's loadStatus property to determine when the status has changed." Why are you updating the banner on the main thread.?

Feature Geodatabase - see comments in osx PR.

singhgag:
@maryharvey I have made the changes.

PR #201, opened 2016-08-03
Corrected draw order for layers and not adding non spatial layers

singhgag:


arcgis-ios-sdk-samples/Edit data/Offline edit and sync/OfflineEditingViewController.swift
```diff
@@ -234,9 +234,12 @@ class OfflineEditingViewController: UIViewController, AGSMapViewTouchDelegate, A
                 self?.liveMode = false
                 
                 self?.map.operationalLayers.removeAllObjects()
-                for featureTable in geodatabase.geodatabaseFeatureTables {
-                    let featureLayer = AGSFeatureLayer(featureTable: featureTable)
-                    self?.map.operationalLayers.addObject(featureLayer)
+                for (_, featureTable) in geodatabase.geodatabaseFeatureTables.enumerate().reverse() {
```
dg0yal: If you dont need the index value (_), then wouldn't it be simpler to just do `for featureTable in gdb.geodatabaseFeatureTables.reverse()` instead of using enumerate().reverse() ?

singhgag:
I made the changes.

singhgag:
@dg0yal Can you merge this?

dg0yal:
I think you'll need to load all the tables before you iterate through them. There is no guarantee that the tables will load in the order you asked them to, so you might end up with adding them in different order to the map.

look at this function - [loadObjects()](http://morphine/api-ref/ios/_a_g_s_loadable_8h.html#ad67ad80f9e2c1418b24b2636bbd0d4e3) It should make things easier.

singhgag:
@dg0yal Now using loadObjects()

PR #204, opened 2016-08-05
Added new sample

singhgag:
- Scene properties expressions


dg0yal:
Can we show the current heading and pitch values next to the slider?

singhgag:
@dg0yal Done.

PR #209, opened 2016-08-29
Added new sample

singhgag:
- Change map view background


dg0yal:
@singhgag Can you update the branch?

dg0yal:
Also, I think its clever that you initialized an empty map with just a spatial reference so that the customized grids would be noticeable, I'd like to stay with this spirit but I feel like there should be some data on the map to make it look like a map. How about we add a MapImageLayer for just this sublayer showing the continents: http://sampleserver6.arcgisonline.com/arcgis/rest/services/WorldTimeZones/MapServer/2
Hopefully that'll keep enough of the background grid visible to be noticeable when you change settings.

singhgag:
The service won't let me hide sublayers, because `Single Fused Map Cache: true`. So I added the layer as is. Also updated the branch to resolve the conflicts.

PR #212, opened 2016-09-15
Added new sample

singhgag:
- ArcGIS vector tiled layer (custom style)


maryharvey:
@singhgag Comments added in original Issue

PR #216, opened 2016-09-20
Simplified Identify layers sample

singhgag:


maryharvey:
Tested in Build 1464 - Comments in original Issue

PR #223, opened 2016-10-14
Added 2 new samples

singhgag:
- List AGOL basemaps
- Org basemap picker


dg0yal:
Putting this one on hold to discuss whether we *really* need two samples to show basemap

PR #244, opened 2017-01-24
Added spatial operations sample

singhgag:


arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -780,6 +798,32 @@
 			);
 			name = "Spatial operations";
 			path = "Geometry/Spatial operations";
+		3E0B58451E36BCCD00951767 /* Analysis */ = {
```
singhgag: @dg0yal you are missing following code
```
			sourceTree = "<group>";
 		};
```

PR #260, opened 2017-05-09
New sample: Update related feature’s attribute

singhgag:


dg0yal:
i find the workflow of this sample very confusing? I tap on a park, but the adjacent feature gets selected!!! I edit the details of the feature, but the changes apply to the adjacent feature? Can't really tell what is going on!

dg0yal:
I say back to the drawing board for this one

PR #261, opened 2017-05-17
New sample: Format coordinates

singhgag:


dg0yal:
- When i update the values in the textbox and hit return, the keyboard is not dismissed automatically. Can we do that?

- I'd like to be able to see the map when the graphic updates after hitting return. Should we move the map above the coordinates so that it is not obscured by the keyboard? Or will the change above make it better?

- Can we try another layout of the field labels and textboxes on ipad (it works ok on iphone)? too much white space on ipad because the textboxes are on a new line. Can we have them side-by-side, similar to the macos

singhgag:
@dg0yal I am done with the changes. Please verify.

PR #267, opened 2017-05-31
Change sublayer renderer

singhgag:


dg0yal:
The sample feels incomplete without the ability to reset the renderer back to the original. Can you add that? 

Everything else looks good.

singhgag:

dg0yal:
Looks good. Please resolve conflict and merge

PR #273, opened 2017-06-15
Generate offline map

singhgag:


dg0yal:
- [ ] Need to handle the case where a user hits cancel on the sign in page. A lot of users will do this because they won't know why they are being asked to sign in

singhgag:
I can't think of a way to handle it. Because in a split view controller I can't pop the sample view controller.

But I can show a different view controller at first, with a message that "This sample requires login. Do you wish to continue?". And when the user presses Yes I can push the map view controller and hence the OauthViewController would show. And if they cancel, I will pop to the previous view controller (one with the message).

dg0yal:
Yeah, something like that. Or even an alert view instead of a new view controller

singhgag:
I did initially have an alert. But that did not help. As it has to be a view controller to pop to. I will add a new view controller with the message.

singhgag:

dg0yal:
- [ ] why are the credentials not being remembers when i log in?
- [ ] lets change the text "This sample requires login to allow basemap to be take offline. Do you wish to continue?" -> "This sample requires you to login in order to take the map's basemap offline. If you would like to proceed, please tap on continue"

dg0yal:
- [ ] Also, I keep getting this result in the offline map when i generate the offline map at the scale where the pipes are just visible
img_0045

arcgis-ios-sdk-samples/Maps/Generate offline map/GenerateOfflineMapViewController.swift
```diff
@@ -0,0 +1,209 @@
...
+                return
+            }
+            
+            //disable cancel button
+            strongSelf.cancelButton.isEnabled = false
+            
+            strongSelf.mapView.map = result?.offlineMap
+            
+            //remove KVO observer
+            strongSelf.generateOfflineMapJob.progress.removeObserver(strongSelf, forKeyPath: "fractionCompleted", context: nil)
```
dg0yal: looks like we remove observer only when the job finishes. 

What if the user exits the sample before completion?

singhgag:
> sample is missing the "i" button to show code/readme..

It should be on the first screen (Tap to continue vc).

> why are the credentials not being remembers when i log in?

Per design. I clear the credential every time. Keeping it consistent with the `Create and save map` sample.

> lets change the text "This sample requires login to allow basemap to be take offline. Do you wish to continue?" -> "This sample requires you to login in order to take the map's basemap offline. If you would like to proceed, please tap on continue"

So we don't need a button? Just a tap gestureRecognizers on the view? Will that be intuitive?

dg0yal:
>It should be on the first screen (Tap to continue vc).

Totally missed that. We should rethink this. It doesn't work on that temporary VC. It should really be on the VC that comprises the sample. 

>So we don't need a button? 

We still do. We should rename the button to "Continue" instead of "Yes".
"Yes" didn't work as well because whenever you see Yes, you expect to see No. And it didn't make sense to put No cause the user could just move on to the next sample. I think "Continue" works better in this regard

singhgag:
> Totally missed that. We should rethink this. It doesn't work on that temporary VC. It should really be on the VC that comprises the sample.

The info button is handled by the `Content display logic` of the viewer. It puts the button on the initial view controller of the sample's storyboard. In this case its the InfoVC. So how about this, we show the actual map view controller as initialVC. On map load, the OAuthViewController will be shown. And on cancel, I can show an alert that this sample requires login. The alert will have two options: Cancel and Continue. On continue, I will reload the map and that should show the OAuthVC again. Else on cancel I will simply dismiss the alert. How does that sound?

> I don't like it. Makes it very cumbersome to explore the sample.

If we store the credential then we should also provide a way to clear it or sign out. We don't have that right now. With the current implementation, you will be required to login only if you re open the sample. You should not be asked for credential if you start the job multiple times. Like start, cancel, start and so on.

singhgag:

dg0yal:
- [ ] When the alert is display, I see this warning logged to the console : `Presenting view controllers on detached view controllers is discouraged <arcgis_ios_sdk_samples.GenerateOfflineMapViewController: 0x7fbd8b584460>.`

arcgis-ios-sdk-samples/Maps/Generate offline map/GenerateOfflineMapViewController.swift
```diff
@@ -202,8 +207,9 @@ class GenerateOfflineMapViewController: UIViewController, AGSAuthenticationManag
         return extent
     }
 
-    override func didReceiveMemoryWarning() {
-        super.didReceiveMemoryWarning()
-        // Dispose of any resources that can be recreated.
+    deinit {
+        
+        //remove observer
+        self.generateOfflineMapJob?.progress.removeObserver(self, forKeyPath: "fractionCompleted")
```
dg0yal: this causes a crash for cases when the observer is already removed when the job finishes, and then it tries to remove it again when you switch to another sample

singhgag:
@dg0yal I have made the changes. Please take yet another look. I believe I have covered all the cases wrt KVO.

dg0yal:
Looks good

PR #275, opened 2017-06-16
Mhd/search for webmap

mhdostal:
The "Search for web map by keyword" sample.


singhgag:
@mhdostal We should constraint the label in the tableView cell from top and bottom. Otherwise it will keep on growing with the larger text.
simulator screen shot jun 16 2017 2 34 43 pm

mhdostal:
@singhgag I made the changes; please take another look.

PR #276, opened 2017-06-19
New sample: Add delete related feature

singhgag:


dg0yal:
-[ ] Can we show the name of the park that the user tapped on? It's not very intuitive that I tap on a park and see a list of species. If we can add the name, then maybe we can show a banner along the top edge of the map saying something like "Tap on a park to see its species"
- [ ] Didn't realize i could swipe a cell to delete. Can we show a visual indicator or maybe a delete action in each cell
- [ ] Can we scroll to the end of the list when the add finishes? I couldn't tell if anything was happening when i hit that button

singhgag:

dg0yal:
- [ ] The sample title should be "Add/Delete related features"

Also, the panel which lists related features looks better now. But wondering if we should swap the title with the name of the feature, move everything up one level. Right now there is redundant "Related features : species" both in the title and in a sub-section
screen shot 2017-06-23 at 12 15 41 pm

singhgag:
The name of the feature might be omitted in the title. I can clear the title instead.

PR #285, opened 2017-06-26
Fixed Xcode warnings

singhgag:


arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -26,7 +26,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
 
     func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool {
```
njarecha: this is deprecated by Apple.. use the new method. This was the reason we had deprecated our delegate method.

singhgag:
@njarecha I have made the change. Please take another look.

PR #287, opened 2017-06-29
New sample: Map Scale

fozoglu:
This sample demonstrates how to show a map scale in the basemap


dg0yal:
Very nice! Thanks for contributing! 

We do plan on providing a scalebar sample in the near future, so that would dominate this use case. However, would you be willing to make this sample go beyond just "displaying" the scale. For example, allow a user to tap on the scale display and choose a well defined scale range (1:1K, 1:5K, 1:1M, 1:5M....) to zoom to? That would give this sample a reason to co-exist with the scalebar sample that is in the works

fozoglu:
@dg0yal Good idea. I will revise the example.

fozoglu:
@dg0yal I've updated my codes. Can you check it ?

fozoglu:
I am waiting for a long time and unhappy about this. For this reason, I close the my pull request.

PR #289, opened 2017-07-05
Added display grid sample

njarecha:
@singhgag Please review


singhgag:
@njarecha 
The sample overall looks good. Just a suggestion and a bug.

**Suggestion**: The content in the horizontal scroller is being omitted. May be for `Compact` size class you can display the label and scroller in two lines.
**Bug**: The horizontal scroller cell does not align properly on device rotation. We had talked about this. If you can fix the issue then fix has to go in before this branch gets merged. Otherwise create an issue and mark this as blocked on that issue.

njarecha:
Created an issue for the above mentioned bug. Merging this.

PR #305, opened 2017-12-05
Update the sample project

nixta:
Targets iOS 10.0 and updated for Xcode 9.

Includes fixes for warnings on `String.character.count` and `UIApplication.open()`.


dg0yal:
Looks good on Xcode 9.0 and 9.1

PR #307, opened 2017-12-06
New Sample: Raster Layer from GeoPackage.

nixta:


nixta:
I was basing it off the screenshot for the Qt sample which showed the edge of the raster, which I felt was possibly useful (showing it's a raster overlaid on top of a basemap rather than just filling the display, which doesn't show that).

dg0yal:
That's fine. Just need the image to show the updated raster 
Plus you'll have to fix the merge conflict in the project file now that I have merged another PR for a new sample (yikes!)

PR #308, opened 2017-12-06
raster_layer_url sample

sbaskaran:
@dg0yal please review

I was off `master` branch originally, but there were different conflicts with `v.next` when creating PR. I've now created this off `v.next`


arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -956,6 +962,16 @@
 /* End PBXFrameworksBuildPhase section */
 
 /* Begin PBXGroup section */
+		096A55DC1FD8677800D2E91B /* Raster layer (URL) */ = {
+			isa = PBXGroup;
+			children = (
+				09B7FDD21FD86FD200B4DFC1 /* RasterLayerUsingURLVC.swift */,
+				096A55DF1FD8677800D2E91B /* RasterLayerUsingURL.storyboard */,
+			);
+			name = "Raster layer (URL)";
```
dg0yal: Name should be Raster layer (service)

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -1898,6 +1914,7 @@
 				3E81C1B71D8A0C7400470AA4 /* ArcGIS vector tiled layer (custom style) */,
 				3E6294081EAFAF4100084105 /* Web tiled layer */,
 				166E8F951EE4DF6C00C57C46 /* Raster layer (file) */,
+				096A55DC1FD8677800D2E91B /* Raster layer (URL) */,
```
dg0yal: Name should be Raster layer (service)

arcgis-ios-sdk-samples/Layers/Raster layer (URL)/README.md
```diff
@@ -0,0 +1,12 @@
+# Raster layer (URL)
```
dg0yal: service

arcgis-ios-sdk-samples/Layers/Raster layer (URL)/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+This sample demonstrates how to create and use a raster layer from a URL.
```
dg0yal: Reword to say something like create a raster layer from an ArcGIS Image service

arcgis-ios-sdk-samples/Layers/Raster layer (URL)/RasterLayerUsingURLVC.swift
```diff
@@ -0,0 +1,70 @@
...
+    
+    private var map:AGSMap!
+    
+    private var rasterLayer: AGSRasterLayer!
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        //create an image service raster from an online raster service
+        let imageServiceRaster = AGSImageServiceRaster(url: URL(string:"http://sampleserver6.arcgisonline.com/arcgis/rest/services/NLCDLandCover2001/ImageServer")!)
```
dg0yal: prefer to use `https` if service supports it

arcgis-ios-sdk-samples/Layers/Raster layer (URL)/RasterLayerUsingURLVC.swift
```diff
@@ -0,0 +1,70 @@
...
+        //set map view's viewpoint to the raster layer's full extent
+        self.rasterLayer.load { [weak self] (error) in
+            
+            guard error == nil else {
+                SVProgressHUD.show(withStatus: error!.localizedDescription, maskType: .gradient)
+                return
+            }
+            
+            if let center = self?.rasterLayer.fullExtent?.center {
+                self?.mapView.setViewpoint(AGSViewpoint(center: center, scale: 40000000))
```
dg0yal: can we not zoom to a viewpoint based on the full extent?

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
sbaskaran: There are other samples where we follow `Some layer (URL)` format for online services. Would this still be `(service)`?

arcgis-ios-sdk-samples/Layers/Raster layer (URL)/RasterLayerUsingURLVC.swift
sbaskaran: The layer actually doesn't display when `fullExtent` is used. We then need to zoom in a tiny bit in order for it to render.

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
dg0yal: Yeah, I think service is the way to go. You're right, we have used URL in the past, in many of those cases you do pass a URL to the constructor, so it does kind of fit. But not this one. If we had to re-think all previous ones, I think a strong case could be made for using service in all

arcgis-ios-sdk-samples/Layers/Raster layer (URL)/RasterLayerUsingURLVC.swift
nixta: What's the scale when at fullExtent? Looks like the service should display (based off looking in the Web Viewer).

sbaskaran: There is a design limitation when using the full extent of the image as geometry for setting the viewpoint for the full extent of the service in portrait mode, we do not permit RasterLayer currently to zoom out beyond 4x published LOD. I've tagged you on that.

sbaskaran:
@dg0yal please re-review. I've also changed the file names.

PR #312, opened 2017-12-08
New Sample: Symbolize a shapefile

nixta:


dg0yal:
@nixta A couple of things -

1) I'm not confident the on-demand resources is properly set for the shapefile. I don't see a "Downloading...." pane at first launch of the sample as I do with others

2) There are 3 sets of shapefiles in this PR - Subdivisions, Public_Art, and TrailBikeNetwork. Do we need to include all 3? The sample seems to use only Subdivisions

nixta:
Yeah, I've sometimes not seen the loading too (try deleting the app from your simulator or resetting the simulator), but I will double-check the tags.

I wasn't sure what to include. You're correct that the other Shapefiles aren't being used, they're part of the same resource stored in ArcGIS Online, so the other platforms' samples will get the entire zip of 3 Shapefiles. Happy to pare it down if we're OK with being different (but someone working with the sample would then not be able to test points and lines without their own data).

dg0yal:
Ah, good call on deleting the app. Looks better now.
yes, let's pare it down to only what we need. If we end up writing another sample that needs it, we can include it

nixta:
@dg0yal Shapefiles removed.

PR #314, opened 2017-12-08
New Sample: Read Geopackage.

nixta:


arcgis-ios-sdk-samples/Maps/Read a geopackage/ReadGeopackageViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+                return AGSFeatureLayer(featureTable: featureTable)
+            }) ?? []
+            
+            // Create raster layers for each raster in the geopackage.
+            let rasterLayers = self?.geoPackage?.geoPackageRasters.map({ raster -> AGSLayer in
+                return AGSRasterLayer(raster: raster)
+            }) ?? []
+
+            // Keep an array of all the feature layers and raster layers in this geopackage.
+            self?.allLayers.append(contentsOf: featureLayers)
```
dg0yal: can you add the raster layers first and then feature layers? that way in the UI the order I add layers going top to bottom will be more preferable, raster layers will get added first and feature layers on top. This will allow me to see all the layers being added. Right now, the raster layers get added last and obscure feature layers underneath

arcgis-ios-sdk-samples/Maps/Read a geopackage/ReadGeopackageViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+    
+    //MARK: - Segue to and from the Layer Control viewcontroller.
+    
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        if segue.identifier == "LayersSegue",
+            let navigationController = segue.destination as? UINavigationController,
+            let controller = navigationController.viewControllers.first as? GPKGLayersViewController {
+            // Provide the map and all layers to the layer controller UI.
+            controller.map = mapView.map
+            controller.allLayers = allLayers
```
dg0yal: maybe set a preferred content size to something like 300,300 which we use in other samples. Makes it easy to view the map behind and see the changes as you add/remove/reorder layers

nixta: I stole this code from the Map->Manage Operational Layers example which does the same thing and covers the map (even with preferred content size). Will investigate some other samples - I'd rather see the map at the same time too.

arcgis-ios-sdk-samples/Maps/Read a geopackage/ReadGeopackageViewController.swift
nixta: Good idea. Rasters are now listed first before features.

arcgis-ios-sdk-samples/Maps/Read a geopackage/ReadGeopackageViewController.swift
dg0yal: That sample works fine on an iPad. Maybe the preferred size doesn't apply to phones

nixta:
@dg0yal See how this works for you… Popover in Phone or iPad.

arcgis-ios-sdk-samples/Maps/Read a geopackage/ReadGeopackageViewController.swift
nixta: Should work as popover on iPhone or iPad now.

dg0yal:
Nice.

PR #317, opened 2017-12-10
Refactors Node to be Decodable

philium:
This allows ContentPList to be decoded using a PropertyListDecoder. To
facilitate implementation of Decodable, I broke Node up into two parts:
Category and Sample. This more closely resembles the structure of the
plist.


philium:
I wanted to include unit tests, but there isn’t a unit testing target in the project. Shall I add one?

esreli:
@philium is there an issue filed for this PR? If not, would you please do so and then link to this PR? This is so we can prioritize PR reviews in this sprint.

arcgis-ios-sdk-samples/Content Display Logic/Model/Category.swift
```diff
@@ -0,0 +1,71 @@
...
+    
+    init(from decoder: Decoder) throws {
+        let values = try decoder.container(keyedBy: CodingKeys.self)
+        name = try values.decode(String.self, forKey: .name)
+        samples = try values.decode([Sample].self, forKey: .samples)
+    }
+}
+
+extension Category: Equatable {
+    static func ==(lhs: Category, rhs: Category) -> Bool {
```
philium: This is unnecessary. It can be synthesized by the compiler as of Swift 4.1.

arcgis-ios-sdk-samples/Content Display Logic/Model/Category.swift
```diff
@@ -0,0 +1,71 @@
...
+        let values = try decoder.container(keyedBy: CodingKeys.self)
+        name = try values.decode(String.self, forKey: .name)
+        description = try values.decode(String.self, forKey: .description)
+        storyboardName = try values.decode(String.self, forKey: .storyboardName)
+        dependencies = try values.decodeIfPresent([String].self, forKey: .dependencies) ?? []
+    }
+}
+
+extension Category.Sample: Equatable {
+    static func ==(lhs: Category.Sample, rhs: Category.Sample) -> Bool {
```
philium: This is unnecessary. It can be synthesized by the compiler as of Swift 4.1.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -159,11 +156,11 @@ class ContentTableViewController: UITableViewController, CustomSearchHeaderViewD
             self.bundleResourceRequest?.endAccessingResources()
             
             //show view controller
-            self.showSample(indexPath: indexPath, node: node)
+            self.showSample(indexPath: indexPath, node: sample)
         }
     }
     
-    func downloadResource(for node:Node, at indexPath:IndexPath) {
+    func downloadResource(for node: Sample, at indexPath:IndexPath) {
```
esreli: please change `node` input parameter to `sample`.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -215,14 +212,14 @@ class ContentTableViewController: UITableViewController, CustomSearchHeaderViewD
         }
     }
     
-    func showSample(indexPath: IndexPath, node: Node) {
+    func showSample(indexPath: IndexPath, node: Sample) {
```
esreli: please change `node` input parameter to `sample`.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -30,7 +30,8 @@ class ContentCollectionViewController: UIViewController, UICollectionViewDataSou
     
     private var headerView:CustomSearchHeaderView!
     
-    var nodesArray:[Node]!
+    /// The categories to display in the collection view.
+    var categories = [Category]()
```
esreli: This variable is set twice. Once here and once in the App Delegate. Consider reducing this to once. As a result, this member could be made static.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -116,15 +77,8 @@ class ContentCollectionViewController: UIViewController, UICollectionViewDataSou
```
esreli: Could you remove this TODO? If this is something we'd like to implement, this belongs as an issue and out of version control.

I recognize this is not a part of the original PR, I just noticed it.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -116,15 +77,8 @@ class ContentCollectionViewController: UIViewController, UICollectionViewDataSou
...
-        var nodes = [Node]()
-        for node in self.nodesArray {
-            let children = node.children
-            if let matchingNodes = children?.filter({ return names.contains($0.displayName) }) {
-                nodes.append(contentsOf: matchingNodes)
-            }
-        }
-        return nodes
+    func samplesByNames<C: Collection>(_ names: C) -> [Sample] where C.Element == String {
+        return categories.flatMap({ $0.samples }).filter({ names.contains($0.name) })
```
esreli: nice and succinct!

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
philium: I took a look at the TODO and I am all for removing it, but I think it should be a separate PR. Removing it could be done before or after this PR is merged without affecting anything. That's kind of my rule of thumb for making unrelated changes in a PR.

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -41,6 +41,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
         splitViewController.minimumPrimaryColumnWidth = 320
         splitViewController.maximumPrimaryColumnWidth = 320
         
+        // Decode and populate Categories.
+        let contentCollectionViewController = (splitViewController.viewControllers.first as! UINavigationController).viewControllers.first as! ContentCollectionViewController
```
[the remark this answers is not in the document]
philium: I think I'll leave it the way it is. It is fairly common for view controllers to be given a reference to the data they need to display. `ContentTableViewController`, for example, is given an array of samples to display.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
philium: My goal was to remove the source (location) of the categories/samples as a dependency from `ContentCollectionViewController`. You are correct that the variable is set twice (once on initialization and once when it is given the data to prevent), but creating an empty array is a very inexpensive operation.

I could update the variable rather than assign a new array:
```swift
let categories = try PropertyListDecoder().decode([Category].self, from: data)
contentCollectionViewController.categories.append(contentsOf: categories)
```
But I'm not sure that makes any practical difference.

arcgis-ios-sdk-samples/AppDelegate.swift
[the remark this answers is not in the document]
nixta: I agree with @philium here, but I would prefer to see the `[Category]` building code refactored to either a function or property on `AppDelegate` that is used to set the `contentCollectionViewController.categories` property. It would make this code a bit cleaner and it's purpose clearer.

Right now we have:

1. Get the view controller
2. Try a couple of things that could throw exceptions.
3. Now back to the view controller. Let's set its `categories` property to something else that could throw an exception.

Would be much cleaner to refactor that away to:
1. Get the view controller.
2. Set its categories to `AppDelegate.readCategoriesFromConfiguration()`.

And of course `readCategoriesFromConfiguration()` will return a `[Category]` or else will `fatalError()`.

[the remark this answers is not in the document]
nixta:
``` Swift
lazy var configuredCategories: [Category] = {
    do {
        let url = Bundle.main.url(forResource: "ContentPList", withExtension: "plist")!
        let data = try Data(contentsOf: url)
        return try PropertyListDecoder().decode([Category].self, from: data)
    } catch {
        fatalError("Error decoding categories: \(error)")
    }
}()
```
and
``` Swift
// Decode and populate Categories.
let contentCollectionViewController = (splitViewController.viewControllers.first as! UINavigationController).viewControllers.first as! ContentCollectionViewController
contentCollectionViewController.categories = configuredCategories
```
seems to work nicely.

@philium - what do you think? I think this cleans up the logic a bit. Rest of the PR looks good to me.

philium:
@esreli @nixta Please re-review.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
esreli: Following your goal, there's also the of declaring categories as an implicit unwrapped optional:

```swift
var categories: [Category]!
```

PR #323, opened 2017-12-15
Viewshed GeoElement sample

mhdostal:
This is the Viewshed (GeoElement) sample.

A viewshed shows the visible and obstructed areas from an observer's vantage point. This sample demonstrates how to attach a viewshed to a moving GeoElement.


mhdostal:
...wrong branch; should be v.next

PR #334, opened 2018-01-09
Update label constraint

mhdostal:
Move the instruction label to immediately below the navigation bar, as opposed to leaving a small gap.  Required removing the original top constraint (top of label to top of superview top margin) and adding one for the top of label to top of safe area.


mhdostal:
Needs to go into v.next; closing

PR #335, opened 2018-01-09
Constrain top of label to top of Safe Area, not the superView top margin

mhdostal:
Move the instruction label to immediately below the navigation bar, as opposed to leaving a small gap. Required removing the original top constraint (top of label to top of superview top margin) and adding one for the top of label to top of safe area.


mhdostal:
@nixta @dg0yal This addresses feedback from @saratchandrakarumuri regarding the positioning of the label.

nixta:
Hey Mark. Couple of items that struck me (the first two of which might have been true before this change):

1. The label is a child of the scene view. It should be a sibling, ordered to appear above.
2. The label itself is a bit cramped. It might be better to place it in a parent view with padding of 8pt on each edge.
3. When the phone is in landscape orientation, the `Licensed for Developer Use` watermark can appear behind the label and make it very hard to read. Embedding the label in an opaque parent UIView or Visual Effect View should solve this.

I'll submit a PR for you to take a look at that addresses this so you can see if you like the look of it.

mhdostal:
@nixta I've merged the PR you created (updating the label).  Please re-review.

PR #339, opened 2018-01-19
Modify to make use of Safe Area for iOS 11 and iPhone X

nixta:
Hey @gdhakal - could you take a look at this please? It's meant to make the sample compatible with Safe Area and suitable for iPhone X.

Could you let me know if this looks OK for you in Xcode and if it runs OK? I think it might be a bug in Xcode, but playing with constraints makes the instructions label disappear in the storyboard sometimes, though it seems to run properly.


gdhakal:
@nixta I hadn't worked with Safe Area before. Learnt something new today. Thanks for putting this together. I tested it on simulator and it works great!

PR #341, opened 2018-02-13
Suggestion for constraining popover to safe area.

nixta:
Hey Suganya,

In case this helps, but actually I can't reproduce what you're seeing. If I use commit 66bb7d9f83b6044b9987ccd10737807037af0081 then the popover appears just fine. Anyway, this embeds the popover in a stack view and the stack view is constrained to the safe area.

But since I can't reproduce what you're seeing (which I didn't try to do until after I "fixed" things :)) perhaps this won't help you…


sbaskaran:
@nixta I ran the sample from your branch. Unfortunately, I still see that the popover is behind the safe area (I've been able to get the content positioned correct). 

screen shot 2018-02-13 at 10 58 59 am

PR #346, opened 2018-02-21
Viewshed location popover display ios10 compatibility

nixta:
Restructured to work with iOS 10.

@njarecha Could you confirm that this fixes the issue you were seeing please?


njarecha:
@nixta Yes, it fixes the issue I was seeing. However, I would suggest increasing the margin on left and right. The controls are too close to the edge.

nixta:
Thanks @njarecha - does this look better?

PR #353, opened 2018-04-23
Fix Xcode typos

stevemoser:


nixta:
Good catch. Thank you @stevemoser!

PR #355, opened 2018-04-26
New Sample: Open a scene (portal item)

nCastle1:
This sample demonstrates how to open a web scene from a portal item.


mhdostal:
@nCastle1 One more thing... the storyboard, the constraint for the "Scene View.bottom" should equal the SuperView's bottom (so `bottom = Scene View.bottom`).  This will match the other samples; see the "Display a Scene" storyboard for an example.  To change it, select the constraint and in the inspector panel on the right, and choose "Superview".

constraint

arcgis-ios-sdk-samples/Scenes/Open a scene (portal item)/OpenScene.storyboard
```diff
@@ -0,0 +1,43 @@
...
+    <scenes>
+        <!--Open Scene View Controller-->
+        <scene sceneID="qMc-Za-MJD">
+            <objects>
+                <viewController id="oI0-pg-r9z" customClass="OpenSceneViewController" customModule="arcgis_ios_sdk_samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <view key="view" contentMode="scaleToFill" id="Wvg-jB-teh">
+                        <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <subviews>
+                            <sceneKitView clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="ryt-GZ-85l" customClass="AGSSceneView">
```
nixta: @nCastle1 This appears in the storyboard editor as an `SCNView` (even though the custom class is `AGSSceneView`) which seems undesirable. I think you must have dragged a SceneKit View into the storyboard from the Object Browser. You should instead drag in a blank View.

PR #356, opened 2018-04-30
fix issue that resulted in white space above ContainerView in iOS 10.x

sbaskaran:
@dg0yal please review 

cc: @saratchandrakarumuri


saratchandrakarumuri:
I have tested this branch on iOS 10 devices and simulators, and I can see the issue is fixed

PR #363, opened 2018-05-15
Updated basemaps in samples

saratchandrakarumuri:
Updated basemaps in change basemap and vector tiled layer samples to match with other sample viewers


saratchandrakarumuri:
@dg0yal I have already updated the screenshots for those two samples. I have updated the URLs to use https

dg0yal:
Is it possible to change the text size of the segmented control so that  basemap names fit properly?

saratchandrakarumuri:
@dg0yal I have changed the size of text in segmented control and I have updated the screenshot accordingly

dg0yal:
Looks good. Just one more thing to resolve. The `Change Basemap` sample doesn't work right on iPhone 8 Plus in landscape mode 

image

saratchandrakarumuri:
Fixed issue with landscape mode and extended blur view beyond safe area to cover whole screen on iPhoneX in landscape mode

PR #367, opened 2018-05-30
Removes an extra space

philium:
Removes a space that was inserted before the source code, causing the first line to always be offset.

Before: 
simulator screen shot - iphone x - 2018-05-30 at 08 42 15

After: 
simulator screen shot - iphone x - 2018-05-30 at 09 01 20


philium:
@esreli @nixta Could one of you please review this PR?

PR #368, opened 2018-05-30
Fixes source code view on iOS 10

philium:
Fixes an issue where the source code view underlaid the navigation bar on iOS 10.


philium:
@esreli @nixta Please review.

PR #371, opened 2018-06-04
Style WMS layer

philium:
This sample demonstrates how to select from one of the defined styles for a WMS sublayer.


arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+        let alertController = UIAlertController(title: nil, message: "Failed to load WMS layer", preferredStyle: .alert)
+        let okayAction = UIAlertAction(title: "OK", style: .default)
+        alertController.addAction(okayAction)
+        alertController.preferredAction = okayAction
+        present(alertController, animated: true)
+    }
+    
+    /// Enables the segmented control if the view has loaded and there is more
+    /// than one style.
+    func enableSegmentedControlIfPosasible() {
```
esreli: typo!

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+    /// The map displayed in the map view.
+    let map: AGSMap
+    let layer: AGSWMSLayer
+    var styles = [String]()
+    
+    /// The map view managed by the view controller.
+    @IBOutlet weak var mapView: AGSMapView!
+    @IBOutlet weak var segmentedControl: UISegmentedControl!
+    
+    required init?(coder aDecoder: NSCoder) {
```
esreli: Can we follow the same pattern as we do in other `UIViewControllers` where variables are declared as unconditionally unwrapped optionals and then handle all setting in `viewDidLoad`?

```swift
var map: AGSMap!
var layer: AGSWMSLayer!
```

I think what you did in overriding the `init` is totally fine but it doesn't match the pattern we have laid out in other View Controllers.

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+        super.viewDidLoad()
+        
+        (self.navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["StyleWebMapServiceLayerViewController"]
+        
+        // Assign the map to the map view.
+        mapView.map = map
+        
+        // The segmented control needs to be enabled here if the layer is
+        // already loaded.
+        enableSegmentedControlIfPosasible()
```
nixta: Would it be posasible to fix this typo please?

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+        let okayAction = UIAlertAction(title: "OK", style: .default)
+        alertController.addAction(okayAction)
+        alertController.preferredAction = okayAction
+        present(alertController, animated: true)
+    }
+    
+    /// Enables the segmented control if the view has loaded and there is more
+    /// than one style.
+    func enableSegmentedControlIfPosasible() {
+        guard isViewLoaded, styles.count > 1 else { return }
```
esreli: 1. should this `guard else` statement disable the segment control if the conditions aren't met? It would encourage a strong bind between the condition and the UI

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+    
+    @IBAction func changeStyle(_ sender: UISegmentedControl) {
+        guard let sublayer = layer.sublayers.firstObject as? AGSWMSSublayer else {
+            preconditionFailure()
+        }
+        sublayer.currentStyle = styles[sender.selectedSegmentIndex]
+    }
+    
+    override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
```
nixta: Is this override necessary?

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
esreli: 2. what happens if there's more than 2 styles? Should the segment control update the number of segments?

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -483,6 +483,10 @@
 		3EFDE0F81E36ABB700CBCD92 /* FeatureCollectionLayerQueryVC.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3EFDE0F41E36ABB700CBCD92 /* FeatureCollectionLayerQueryVC.swift */; };
 		3EFDE0F91E36AE3100CBCD92 /* FeatureCollectionLayerVC.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = 3EFDE0F11E36ABB700CBCD92 /* FeatureCollectionLayerVC.swift */; };
 		3EFDE0FA1E36AE3100CBCD92 /* FeatureCollectionLayerQueryVC.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = 3EFDE0F41E36ABB700CBCD92 /* FeatureCollectionLayerQueryVC.swift */; };
+		4CC00F7520C1FAE5004244EF /* StyleWebMapServiceLayerViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4CC00F7320C1FAE5004244EF /* StyleWebMapServiceLayerViewController.swift */; };
+		4CC00F7620C1FAE5004244EF /* StyleWebMapServiceLayer.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 4CC00F7420C1FAE5004244EF /* StyleWebMapServiceLayer.storyboard */; };
+		4CEA34FD20C59D640087F93A /* README.md in Resources */ = {isa = PBXBuildFile; fileRef = 4CEA34FC20C59D640087F93A /* README.md */; };
```
nixta: Let's not include the README in the project for now.

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+    /// Enables the segmented control if the view has loaded and there is more
+    /// than one style.
+    func enableSegmentedControlIfPosasible() {
+        guard isViewLoaded, styles.count > 1 else { return }
+        segmentedControl.isEnabled = true
+    }
+    
+    @IBAction func changeStyle(_ sender: UISegmentedControl) {
+        guard let sublayer = layer.sublayers.firstObject as? AGSWMSSublayer else {
+            preconditionFailure()
```
esreli: is this better than an `assert`?

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
esreli: 3. moving the setting of variables from the `init?` to `viewDidLoad` will mean that you no longer need to check if `isViewLoaded`

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
nixta: Do we really want the app to bomb out if this happens? Strikes me that an `if let` would be more friendly.

esreli:
@dg0yal can you specify Mike Wilburn as a contributor? I would like to add him as a reviewer for this PR, specifically so that he may review the README.md

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
esreli: Actually, I would prefer a graceful failure like @nixta suggests. So that the app published to the store won't bomb out.

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
philium: Good catch! Thank you.

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
philium: @esreli Good question. The advantage of `precondition()`/`preconditionFailure()` is twofold:
1. It prevents forward progress in both debug (non-optimized) and release (optimized) builds.
2. It returns `Never`. If this were an assert, a return statement would also be needed.

`preconditionFailure()` is a good choice here because we already know that the layer exists. If the layer didn't exist, then there wouldn't be any styles and the segmented control wouldn't be enabled. The fact that the handler fired is proof that the the sublayer does exist. If the control somehow became enabled without getting any styles, we have bigger problems and should bail immediately.

I could have force unwrapped `firstObject`, but that leaves ambiguity about whether the `nil` case has been considered. Using the `guard`/`preconditionFailure()` combo makes it explicit that I have considered the `nil` case and essentially deemed it impossible.

I also could have subscripted into `sublayers`, but that really isn't any different than `firstObject!`.

If you'd like, I could add a message to make it even more explicit why we can't move forward:
`preconditionFailure("Styles control enabled without a sublayer")`

philium: @nixta See my response to @esreli. The `preconditionFailure()` should never be reached. But if it does, it is an indication of a bigger problem and unconditionally failing makes sense.

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
philium: I realize this deviates from the style adopted by other samples, but there are good reason to do so:

1. `viewDidLoad()` is not intended to be the initializer for the view controller. `init` is a more appropriate place to initialize members that are not dependent on the view being loaded.
2. The `AGSWMSLayer` needs to load before it can be displayed. By telling it to load in the initializer, we have increased the likelihood that it will have loaded by the time the map view is displayed to the user.

I strongly believe that this is a good model for new samples to follow moving forward.

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
philium: Remember that in Swift, a subclass must override all designated initializers or none of them. Since `init(coder:)` has been overridden, `init(nibName:bundle:)` must be overridden as well.

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
philium: There are also benefits for the view controller's other properties:

1. They don't need to be `IUO`s because they are set in an initializer.
2. They can be immutable because they are never changed after initialization.

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
esreli: For an app that is published to the iTunes App store, I feel strongly about removing a forced crash for a potentially missing web asset layer. 

In the event we have bigger problems, I would encourage it be handled gracefully instead of bailing. I would rather the app's component fail than the entire app fail.

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
esreli: Yes I agree with your reasoning. I also believe in code consistency, perhaps we adopt the pattern across all samples?

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
nixta: > If the control somehow became enabled without getting any styles, we have bigger problems and should bail immediately.

Then I would ask that we either gracefully log something from the `guard` and not kill the app if we feel that this is application logic error, or that we `if let` our way to a `sublayer` variable if we feel that it'll never rightly happen.

And should we not rather/also be guarding that the following `selectedSegmentIndex` is within range for the `styles` array?

More fundamentally, I question the use of a `guard` statement whose content kills the app. `guard` is meant to help defend against issues, not trigger them (I'd consider a hard kill an issue, whether it's us or the array access pulling the trigger). By all means use a guard statement to ensure that the next bits of code will work, but surely you don't want to make the result of them not working terminal for the whole app?

arcgis-ios-sdk-samples/Layers/Style WMS layers/StyleWebMapServiceLayerViewController.swift
```diff
@@ -0,0 +1,101 @@
...
+        let okayAction = UIAlertAction(title: "OK", style: .default)
+        alertController.addAction(okayAction)
+        alertController.preferredAction = okayAction
+        present(alertController, animated: true)
+    }
+    
+    /// Enables the segmented control if the view has loaded and there is more
+    /// than one style.
+    func enableSegmentedControlIfPossible() {
+        guard isViewLoaded, styles.count > 1 else { return }
```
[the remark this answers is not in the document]
philium: I like it!

PR #373, opened 2018-06-12
Drops UIColor where possible

philium:
Takes advantage of type inference by removing explicit references to `UIColor` where possible.


arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -96,14 +96,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
     
     func modifyAppearance() {
         UINavigationBar.appearance().titleTextAttributes = [NSAttributedStringKey.foregroundColor : UIColor.white]
-        UINavigationBar.appearance().barTintColor = UIColor.primaryBlue()
-        UINavigationBar.appearance().tintColor = UIColor.white
+        UINavigationBar.appearance().barTintColor = .primaryBlue()
```
nixta: Let's take advantage of this PR to also change these funcs to vars.

PR #375, opened 2018-06-18
Adds a Distance Measurement Analysis sample

philium:
This sample demonstrates measuring 3D distances between two points in a scene.


arcgis-ios-sdk-samples/Analysis/Distance measurement analysis/README.md
```diff
@@ -0,0 +1,25 @@
+# Distance Measurement Analysis
+This sample demonstrates measuring 3D distances between two points in a scene.
+
+The distance measurement analysis allows you to add the same measuring experience found in ArcGIS Pro, City Engine, and the ArcGIS API for JavaScript to your app. You can set the unit system of measurement (metric or imperial) and have the units automatically switch to one appropriate for the current scale. The rendering is handled internally so they do not interfere with other analyses like viewsheds.
+
+![Image](image1.png)
+
+## How to use the sample
+Choose a unit system for the measurement in the segmented control. Click any location in the scene to start measuring. Move the mouse to an end location, and click to complete the measure. Click a new location to clear and start a new measurement.
```
nixta: This references how the macOS version works. I think this is more of a tap-drag workflow, and we should not mention 🐭.

arcgis-ios-sdk-samples/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // Configure the scene view.
+        sceneView.scene = scene
+        sceneView.touchDelegate = self
+        let camera = AGSCamera(lookAt: locationDistanceMeasurement.startLocation, distance: 200, heading: 0, pitch: 45, roll: 0)
```
nixta: One suggestion. In portrait orientation on the phone, focusing on the start point of the analysis leaves half the analysis clipped.

What do you think of this?
``` Swift
let targetGeom = AGSGeometryEngine.union(ofGeometry1: locationDistanceMeasurement.startLocation,
                                         geometry2: locationDistanceMeasurement.endLocation)?.extent.center
                 ?? locationDistanceMeasurement.startLocation
let camera = AGSCamera(lookAt: targetGeom, distance: 200, heading: 0, pitch: 45, roll: 0)
```

Also, I wonder how Qt, Android and .NET have handled this for phone devices in portrait? If they're not considering it, I think this might be a nice option (or at least consideration) to feed back.

arcgis-ios-sdk-samples/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -0,0 +1,161 @@
...
+    
+    func geoView(_ geoView: AGSGeoView, didMoveLongPressToScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        sceneView.screen(toLocation: screenPoint) { [weak self] mapLocation in
+            self?.locationDistanceMeasurement.endLocation = mapLocation
+        }
+    }
+    
+    // MARK: UIViewController
+    
+    override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
```
nixta: We can dispense with this here for now.

If you think all samples should have this, please open an issue and we can discuss.

arcgis-ios-sdk-samples/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
nixta: Actually, this might be a better option (as it only deals with extents and returns an extent):

``` Swift
let targetGeom = AGSGeometryEngine.combineExtents(of: locationDistanceMeasurement.startLocation,
                                         andGeometry: locationDistanceMeasurement.endLocation)?.center
                 ?? locationDistanceMeasurement.startLocation
```

arcgis-ios-sdk-samples/Analysis/Distance measurement analysis/README.md
philium: Thank you. I changed the first sentence, but completely ignored the rest of the paragraph. I'll update it.

arcgis-ios-sdk-samples/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
philium: I copied this from the macOS implementation before the override was removed there. I'll get rid of it.

PR #387, opened 2018-06-22
Ensures LayerStatusViewController updates UI from the main thread

philium:


arcgis-ios-sdk-samples/Maps/Display layer view state/LayerStatusViewController.swift
```diff
@@ -70,14 +70,15 @@ class LayerStatusViewController: UIViewController, UITableViewDataSource, UITabl
         //layer status logic
         //assign a closure for layerViewStateChangedHandler, in order to receive layer view status changes
         self.mapView.layerViewStateChangedHandler = { [weak self] (layer:AGSLayer, state:AGSLayerViewState) in
-            if let weakSelf = self {
+            DispatchQueue.main.async {
+                guard let `self` = self else { return }
```
pgruenler: I'm wondering if it would make sense to use `strongSelf` here instead of `self`.  @rolson, do you usually use 'strongSelf' in cases like this?

rolson: i've never seen that... i'm not sure about it. Is this a widely accepted pattern?

philium: @pgruenler @rolson The method I used to unwrap `weak self` is considered a compiler bug. Swift 4.2 will allow using `guard let self = self else { … }`, but for now I'll just use `strongSelf`.

PR #393, opened 2018-06-25
Removes TPKeyboardAvoiding

philium:
It wasn't being used.


philium:
Nope. My guess is that it was used at one point, but the usage was removed and the library was left in place.

PR #396, opened 2018-06-25
HorizontalColorPicker: Fixes animation when selected color changes

philium:
Before:

before

After:

after


rolson:
asking @njarecha to review, I think he wrote this component

PR #402, opened 2018-06-26
Makes MapViewDrawStatusViewController update UI on main thread.

philium:
This also fixes an issue where the draw status wasn't being displayed properly when the view first appears.

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Display drawing status/MapViewDrawStatusViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Display drawing status/MapViewDrawStatusViewController.swift
@@ -48,7 +48,7 @@ class MapViewDrawStatusViewController: UIViewController {
         
         //add observer for drawStatus on mapView
         //so we can show/hide an indicator when the status change
-        mapView.addObserver(self, forKeyPath: #keyPath(AGSGeoView.drawStatus), options: [.initial], context: nil)
+        mapView.addObserver(self, forKeyPath: #keyPath(AGSGeoView.drawStatus), options: .initial, context: nil)
     }
     
     override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey : Any]?, context: UnsafeMutableRawPointer?) {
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Maps/Display drawing status/MapViewDrawStatusViewController.swift
```diff
@@ -41,23 +41,26 @@ class MapViewDrawStatusViewController: UIViewController {
...
+    }
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
         
         //add observer for drawStatus on mapView
-        //so we can show/hide an indicator when the status change 
-        self.mapView.addObserver(self, forKeyPath: "drawStatus", options: .new, context: nil)
+        //so we can show/hide an indicator when the status change
+        mapView.addObserver(self, forKeyPath: #keyPath(AGSGeoView.drawStatus), options: .initial, context: nil)
```
philium: The initial option ensures that the view is updated correctly when it first displays. The new option was unnecessary because we consult the map view directly for the new value.

rolson:
is it better to do this in willAppear/didDisappear rather than load/dealloc?

philium:
@rolson Good question. It is generally better for a couple reasons:

- It is possible for a view controller to be initialized and then de-initialized without the view being loaded. If that were to happen, attempting to remove the observer in `deinit` would cause a crash. Using `view[will/did]appear` to register and `view[will/did]disappear` to unregister ensures that removing the observer will only happen after it has been added as an observer.
- The observation is only used to update the view. Therefore it only makes sense to observe changes while the view is displayed. If the observation affected something other than the view, then it would be best to register the observer in the initializer rather than in `viewDidLoad()`.

Granted, for this particular sample, there is no practical difference between the two approaches. We know that the view controller will always be presented after it is created. And it is likely that the map view will not update it's draw status when it isn't displayed. But that may not always be the case, so when the observation only affects the view, it is generally a good idea to add/remove the observer in the view lifecycle methods. Does that make. sense?

rolson:
Yes, perfect sense. Good explanations!

PR #411, opened 2018-07-03
Fix Line of Sight (GeoElement) sample to account for target height

nixta:


arcgis-ios-sdk-samples/Analysis/Line of sight (geoelement)/LineOfSightGeoElementViewController.swift
```diff
@@ -104,6 +104,18 @@ class LineOfSightGeoElementViewController: UIViewController {
 
         // set the initial viewpoint
         scene.initialViewpoint = AGSViewpoint(center: initialViewpointCenter, scale: 6000)
+
+        // the symbol is on the ground, but we want to do analysis against a sensible height above ground
+        taxiSymbol.load { [weak self] error in
+            guard error == nil else {
```
philium: > A guard statement is used to transfer program control out of a scope if one or more conditions aren’t met.

In my experience, it is best not to put anything else inside a `guard` statement except log statements. It is very easy for someone else to see a `guard` statement and assume it is just an early exit, overlooking other side affects it might have. If something needs to happen (other than transferring program control out of a scope) regardless of whether the condition(s) are met, then an `if` statement should be used instead of a `guard` statement.

What would you think of something like this instead?

```swift
taxiSymbol.load { [weak self] error in
    let newTargetOffsetZ: Double
    if let error = error {
        print("Error loading the taxi symbol: \(error.localizedDescription)")
        // default to a line of sight target offset of 2m above ground
        newTargetOffsetZ = 2
    } else {
        // use the model's height as the line of sight target offset above ground
        newTargetOffsetZ = taxiSymbol.height
    }
    self?.lineOfSight.targetOffsetZ = newTargetOffsetZ
}
```

nixta: Thanks. Yeah. That's a good point.

But I anyway realized late last night that the code could be improved and will make that push now.

arcgis-ios-sdk-samples/Analysis/Line of sight (geoelement)/LineOfSightGeoElementViewController.swift
```diff
@@ -104,6 +104,19 @@ class LineOfSightGeoElementViewController: UIViewController {
 
         // set the initial viewpoint
         scene.initialViewpoint = AGSViewpoint(center: initialViewpointCenter, scale: 6000)
+
+        // default to a line of sight target offset of 1.5m above ground
+        lineOfSight.targetOffsetZ = 1.5
+        sour
```
philium: Please remove this line.

nixta:
Thank you, @philium!

PR #425, opened 2018-07-10
Refactors MapLoadedViewController to update load status before the vi…

philium:
…ew is displayed.

I also took the opportunity to cleanup the code:
• Made the map non-optional.
• Conformed the code to our style guide.

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Map loaded/MapLoadedViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Map loaded/MapLoadedViewController.swift
@@ -16,60 +16,56 @@ import UIKit
 import ArcGIS
 
 class MapLoadedViewController: UIViewController {
+    /// The map displayed in the map view.
+    let map = AGSMap(basemap: .imageryWithLabels())
 
-    @IBOutlet var mapView:AGSMapView!
-    @IBOutlet var bannerLabel:UILabel!
-    
-    private var map:AGSMap?
+    @IBOutlet var mapView: AGSMapView!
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Maps/Map loaded/MapLoadedViewController.swift
```diff
@@ -16,60 +16,56 @@ import UIKit
...
     
     override func viewDidLoad() {
         super.viewDidLoad()
 
         //setup source code bar button item
-        (self.navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["MapLoadedViewController"]
-        
-        //initialize map with basemap
-        self.map = AGSMap(basemap: AGSBasemap.imageryWithLabels())
+        (navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["MapLoadedViewController"]
```
rolson: would it be safer or better in any way to do this instead:

```swift
(navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["MapLoadedViewController"]
```

philium: Maybe, but all of the other samples use a force cast. 

I'd much rather refactor how we present the source code view. It is very odd that we are giving the button an array of file names.

rolson: ok - I'm fine with entering an issue for that. However let's focus on getting our sample coverage up to par first.

PR #429, opened 2018-07-13
Mhd/list transforms

mhdostal:
The List transforms by suitability sample:  https://devtopia.esri.com/runtime/common-samples/issues/487


arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+        }
+        
+        setupTransformsList()
+    }
+
+    @IBAction func oderByMapExtentValueChanged(_ sender: Any) {
+        setupTransformsList()
+    }
+
+    //MARK: - TableView data source
```
philium: Please implement `UITableViewDataSource.numberOfSections(in:)`. I realize the default is `1`, which is correct in this case, but it is highly unusual to not implement that data source method.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+
+        // if we're missing the grid files, detail which ones
+        var files = ""
+        if transformation.isMissingProjectionEngineFiles {
+            files = "Missing grid files"
+
+            if let gt = transformation as? AGSGeographicTransformation {
+                gt.steps.forEach { (step) in
+                    if step.isMissingProjectionEngineFiles {
+                        files.append(": " + step.projectionEngineFilenames.joined(separator: ","))
```
philium: If there are multiple steps missing files, the message would be something like this (with too many colons):

> Missing grid files: a.gsb: b.gsb

You could use `flatMap` to get an array of all the missing files and then `join` them:

```swift
cell.textLabel?.text = {
    if transformation.isMissingProjectionEngineFiles,
        let geographicTransformation = transformation as? AGSGeographicTransformation {
        let files = geographicTransformation.steps.flatMap { (step) -> [String] in
            step.isMissingProjectionEngineFiles ? step.projectionEngineFilenames : []
        }
        return "Missing grid files: \(files.joined(separator: ", "))"
    } else {
        return ""
    }
}()
```

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+                    }
+                }
+            }
+        }
+        
+        cell.textLabel?.text = transformation.name
+        cell.detailTextLabel?.text = files
+        
+        if let defaultTransform = defaultTransformation {
+            cell.isSelected = transformation.isEqual(to: defaultTransform)
```
philium: Please call `UITableView.selectRow(at:animated:scrollPosition:)` after setting `defaultTransformation`. Setting the selected state here only sets the display state of the cell. The table view resets this when the cell is displayed, so the default cell doesn't actually appear selected.

FYI: You could have used `transformation == defaultTransformation`. No need to unwrap `defaultTransformation` or explicitly call `isEqual(to:)`.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+            else {
+                // add projected graphic
+                addGraphic(projectedGeometry, color: .blue, style: .cross)
+            }
+        }
+        else {
+            // If a transformation is missing grid files, then it cannot be
+            // successfully used to project a geometry, and "projectGeometry" will return nil.
+            // In that case, remove projected graphic
+            if graphicsOverlay.graphics.count > 1 {
```
philium: It might be better to prevent the user from selecting invalid transformations in the first place. One way of doing that is to disable user interaction on those cells in `tableView(_:cellForRowAt:)`:

```swift
cell.isUserInteractionEnabled = !transformation.isMissingProjectionEngineFiles
```

arcgis-ios-sdk-samples/Geometry/List transformations/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## How it works
+The sample sets the location of projection engine data on the device by calling `AGSTransformationCatalog.setProjectionEngineDirectory`.
+
+The list of `AGSDatumTransformations` is created by calling `AGSTransformationCatalog.transformationsBySuitability`, passing in the `AGSSpatialReference` of the original geometry (the input spatial reference) and that of the `AGSMap` (the output spatial reference). Depending on the state of a switch control, the current visible extent of the map is used to sort the list by suitability.
+
+When the user taps on a transformation in the list, the selected transformation is used to reproject an `AGSPoint`. The `AGSGraphic's` geometry is then updated with the new `AGSPoint`. If the selected transformation is not usable (has missing grid files) then no projected point is displayed.
```
philium: Please change "\`AGSGraphic's\`" to "\`AGSGraphic\`'s"

arcgis-ios-sdk-samples/Geometry/List transformations/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+To download projection engine data to your device:
+
+1. Log in to the ArcGIS for Developers site using your Developer account.
+2. In the Dashboard page, click 'Download APIs and SDKs'.
+3. Click the download button next to 'ArcGIS_Runtime_Coordinate_System_Data' to download projection engine data to your computer.
+4. Unzip the downloaded data on your computer.
+5. Use iTunes to copy the `PEDataRuntime` folder to your application's Documents folder.
+  
+#### Tags
```
philium: The Tags section should be at the same level as the other sections (##).

arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
```diff
@@ -837,6 +837,14 @@
 				<key>descriptionText</key>
 				<string>This sample demonstrates how to convert a map location Point to a number of different coordinate notations. Tap on the map or edit the values in textfields</string>
 			</dict>
+            <dict>
```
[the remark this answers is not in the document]
mhdostal: Done.

arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
```diff
@@ -837,6 +837,14 @@
...
+                <key>displayName</key>
+                <string>List transformations</string>
+                <key>storyboardName</key>
+                <string>ListTransformations</string>
+                <key>descriptionText</key>
+                <string>This sample demonstrates how to use the TransformationCatalog to get a list of available DatumTransformations that can be used to project a Geometry between two different SpatialReferences, and how to use one of the transformations to perform the GeometryEngine.project operation.</string>
```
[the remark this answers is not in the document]
mhdostal: They were originally class names/methods; there are many differing ways that's handled in the various samples, but the most common seems to avoid actually class names in favor of descriptive words, so instead of "TransformationCatalog", use "transformation catalog".  I've made those changes in this instance.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+        (self.navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["ListTransformationsViewController"]
+
+        // Get MapView from layout and set a map into this view
+        mapView.map = AGSMap(basemap: AGSBasemap.lightGrayCanvasVector())
+        mapView.graphicsOverlays.add(graphicsOverlay)
+        
+        //add original graphic to overlay
+        addGraphic(originalGeometry, color: .red, style: .square)
+        
+        mapView.map?.load(completion: { [weak self] (error) in
```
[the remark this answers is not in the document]
mhdostal: Is there any advantage to doing that?

The autocomplete fills it out as above.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformations.storyboard
```diff
@@ -0,0 +1,108 @@
...
+                <viewController id="tQc-7l-sMX" customClass="ListTransformationsViewController" customModule="arcgis_ios_sdk_samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <view key="view" contentMode="scaleToFill" id="eqq-a5-6uj">
+                        <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <subviews>
+                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="3bB-y6-ZSJ" customClass="AGSMapView">
+                                <rect key="frame" x="0.0" y="0.0" width="375" height="200"/>
+                                <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                                <constraints>
+                                    <constraint firstAttribute="height" constant="200" id="SBH-SM-eft"/>
```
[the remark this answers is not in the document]
mhdostal: I implemented #1.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformations.storyboard
```diff
@@ -0,0 +1,108 @@
...
+                                            <outlet property="dataSource" destination="tQc-7l-sMX" id="qA0-Of-7gd"/>
+                                            <outlet property="delegate" destination="tQc-7l-sMX" id="McY-aK-daB"/>
+                                        </connections>
+                                    </tableView>
+                                </subviews>
+                            </stackView>
+                        </subviews>
+                        <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                        <constraints>
+                            <constraint firstItem="3bB-y6-ZSJ" firstAttribute="top" secondItem="eqq-a5-6uj" secondAttribute="top" id="21B-pW-bD4"/>
```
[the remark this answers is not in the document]
mhdostal: Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformations.storyboard
```diff
@@ -0,0 +1,108 @@
...
+                            <constraint firstItem="3dc-76-dsw" firstAttribute="top" secondItem="3bB-y6-ZSJ" secondAttribute="bottom" id="KM1-Gc-r85"/>
+                            <constraint firstItem="kRG-V1-ajl" firstAttribute="trailing" secondItem="3dc-76-dsw" secondAttribute="trailing" id="NBI-dc-kK7"/>
+                            <constraint firstItem="3bB-y6-ZSJ" firstAttribute="leading" secondItem="kRG-V1-ajl" secondAttribute="leading" id="bN7-Kh-cnB"/>
```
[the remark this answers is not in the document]
mhdostal: Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+
+            guard let originalGeometry = self?.originalGeometry else { return }
+            self?.mapView.setViewpoint(AGSViewpoint(center: originalGeometry, scale: 5000), duration: 2.0, completion: nil)
+            
+            // set the url for our projection engine data;
+            self?.setPEDataURL()
+        })
+    }
+
+    override func didReceiveMemoryWarning() {
```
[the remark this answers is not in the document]
mhdostal: Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+import UIKit
+import ArcGIS
+
+class ListTransformationsViewController: UIViewController, UITableViewDelegate, UITableViewDataSource {
+
+    @IBOutlet var mapView: AGSMapView!
+    @IBOutlet var tableView: UITableView!
+    @IBOutlet var orderByMapExtent: UISwitch!
+    
+    var datatumTransformations: [AGSDatumTransformation] = []
```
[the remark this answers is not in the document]
mhdostal: Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+    var originalGeometry = AGSPoint(x: 538985.355, y: 177329.516, spatialReference: AGSSpatialReference(wkid: 27700))
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // Add the source code button item to the right of navigation bar.
+        (self.navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["ListTransformationsViewController"]
+
+        // Get MapView from layout and set a map into this view
+        mapView.map = AGSMap(basemap: AGSBasemap.lightGrayCanvasVector())
```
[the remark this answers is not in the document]
mhdostal: Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+        //add original graphic to overlay
+        addGraphic(originalGeometry, color: .red, style: .square)
+        
+        mapView.map?.load(completion: { [weak self] (error) in
+            if error != nil {
+                print("map load error = \(String(describing: error))")
+                return
+            }
+
+            guard let originalGeometry = self?.originalGeometry else { return }
```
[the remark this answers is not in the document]
mhdostal: Good call.  Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+        }
+        
+        defaultTransformation = AGSTransformationCatalog.transformation(forInputSpatialReference: inputSR, outputSpatialReference: outputSR)
+        tableView.reloadData()
+    }
+    
+    func setPEDataURL() {
+        // find the PEDataRuntime folder from the documents directory
+        // added using iTunes
+        let path = NSSearchPathForDirectoriesInDomains(FileManager.SearchPathDirectory.documentDirectory, FileManager.SearchPathDomainMask.userDomainMask, true)
```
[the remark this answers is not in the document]
mhdostal: That looks good, but doesn't that restrict the location of the "PEDataRuntime" folder to the top-level of the user's documents folder?  Just want to make sure I understand what it's doing.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+        let predicate = NSPredicate(format: "SELF MATCHES %@", "PEDataRuntime")
+        let peDataRuntimePaths = subpaths.filter({ (objc) -> Bool in
+            return predicate.evaluate(with: objc)
+        })
+
+        // use first matching path as path to PE data
+        if let documentPEDataRuntime = peDataRuntimePaths.first {
+            // found "PEDataRuntime" folder, create full url
+            let peDataURL = URL(fileURLWithPath: path[0]).appendingPathComponent(documentPEDataRuntime)
+            try? AGSTransformationCatalog.setProjectionEngineDirectory(peDataURL)
```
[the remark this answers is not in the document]
mhdostal: In a regular app, I'd say yes, but since these are self-contained samples, I think it's OK.  I'll put a comment in regarding that.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
mhdostal: Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
mhdostal: Nice.  Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
mhdostal: Very nice, thanks.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
mhdostal: Good call.

arcgis-ios-sdk-samples/Geometry/List transformations/README.md
mhdostal: Done.

arcgis-ios-sdk-samples/Geometry/List transformations/README.md
mhdostal: Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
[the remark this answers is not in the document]
philium: The advantage is readability since the parameter label can be omitted. The way you have it isn't wrong, but In my experience it is more common to use a trailing closure than not.

I just learned something new. I was seeing autocomplete suggesting a trailing closure:

screen shot 2018-07-17 at 1 52 50 pm

The difference seems to be the optional chaining. Since `load(completion:)` is being called on an optional map using optional chaining, autocomplete doesn't use a trailing closure. Fascinating!

Since you're creating the map in the same scope, I would suggest you create a variable for the map and call load from that rather than the map view's map property. That way you avoid having to unwrap the map.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
philium: @mhdostal Yes. Make sure you reload the table view before selecting the row for the default transformation:

```swift
tableView.reloadData()
if let defaultTransformation = defaultTransformation,
    let row = datatumTransformations.index(of: defaultTransformation) {
    let indexPath = IndexPath(row: row, section: 0)
    tableView?.selectRow(at: indexPath, animated: false, scrollPosition: .middle)
}
```

mhdostal:
@philium I've made the changes and some further tweaks.  Please re-review.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
mhdostal: I've refactored this a bit; when the transformation array is reset/rebuilt, the project geometry is removed, as is the table highlighting.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+        mapView.map = AGSMap(basemap: .lightGrayCanvasVector())
+        mapView.graphicsOverlays.add(graphicsOverlay)
+        
+        //add original graphic to overlay
+        addGraphic(originalGeometry, color: .red, style: .square)
+        
+        mapView.map?.load() { [weak self] (error) in
+            if error != nil {
+                print("map load error = \(String(describing: error))")
+                return
```
philium: This `return` statement is no longer necessary.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+        // Get MapView from layout and set a map into this view
+        mapView.map = AGSMap(basemap: .lightGrayCanvasVector())
+        mapView.graphicsOverlays.add(graphicsOverlay)
+        
+        //add original graphic to overlay
+        addGraphic(originalGeometry, color: .red, style: .square)
+        
+        mapView.map?.load() { [weak self] (error) in
+            if error != nil {
+                print("map load error = \(String(describing: error))")
```
philium: Using `String.init(describing:)` is only necessary to prevent a warning because `error` is `Optional`. But you've already checked that `error` is not `nil`. Why not unwrap it and avoid it altogether?

```swift
if let error = error {
    print("map load error = \(error)")
```

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+        }
+        
+        setupTransformsList()
+    }
+
+    @IBAction func oderByMapExtentValueChanged(_ sender: Any) {
+        setupTransformsList()
+    }
+    
+    func projectedGraphic() -> AGSGraphic? {
```
philium: The way this method is named and used is more like a property than a method. I also think the implementation might be simpler without the `graphic` variable:

```swift
var projectedGraphic: AGSGraphic? {
    if graphicsOverlay.graphics.count > 1 {
        return graphicsOverlay.graphics.lastObject as? AGSGraphic
    } else {
        return nil
    }
}
```

Let me know what you think.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
mhdostal: Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
mhdostal: Done.

arcgis-ios-sdk-samples/Geometry/List transformations/ListTransformationsViewController.swift
mhdostal: Yea, that works for me.  Not too much of a difference in my mind, though, but I went with your solution.

mhdostal:
@philium Done with the 2nd set of changes.  Please re-review.

mhdostal:
@njarecha Do you want to review this?

arcgis-ios-sdk-samples/Geometry/List transformations/README.md
```diff
@@ -25,7 +25,7 @@ When the user taps on a transformation in the list, the selected transformation
 * AGSGeographicTransformationStep
 * AGSGeometryEngine.project
 
-## [Offline data](id:offline_data)
```
philium: This isn't needed for the link to work? Cool!

mhdostal: The link never worked for me, so I removed it and in the next commit the link reference.

mhdostal:
> philium changed the base branch from v.next  to release/100.3  3 days ago

Was this intentional?

mhdostal:
@philium I've merged in the release/100.3 branch so we're good to go.  Please re-review.

PR #432, opened 2018-07-16
Makes Display Grid settings persist while in the sample.

philium:
Previously the settings would be reset everytime the dialogue was opened, not showing accurate values for each setting.

### DIFF
--- a/arcgis-ios-sdk-samples/Display information/Display grid/DisplayGridViewController.swift
+++ b/arcgis-ios-sdk-samples/Display information/Display grid/DisplayGridViewController.swift
@@ -46,7 +46,7 @@ class DisplayGridViewController: UIViewController, UIAdaptivePresentationControl
             fatalError()
         }
         
-        viewController.mapView = self.mapView
+        viewController.mapView = mapView
         viewController.modalPresentationStyle = .popover
         
         return viewController
--- a/arcgis-ios-sdk-samples/Display information/Display grid/DisplayGridViewController.swift
+++ b/arcgis-ios-sdk-samples/Display information/Display grid/DisplayGridViewController.swift
@@ -17,9 +17,7 @@ import UIKit
 import ArcGIS
 
 class DisplayGridViewController: UIViewController, UIAdaptivePresentationControllerDelegate {
-
     @IBOutlet weak var mapView: AGSMapView!
-    private var gridSettingsViewController:DisplayGridSettingsViewController!
     
     override func viewDidLoad() {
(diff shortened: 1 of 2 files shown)


arcgis-ios-sdk-samples/Display information/Display grid/DisplayGridViewController.swift
```diff
@@ -41,34 +39,47 @@ class DisplayGridViewController: UIViewController, UIAdaptivePresentationControl
...
-    //MARK: - UIAdaptivePresentationControllerDelegate
+    var settingsViewController: DisplayGridSettingsViewController?
     
-    func adaptivePresentationStyle(for controller: UIPresentationController, traitCollection: UITraitCollection) -> UIModalPresentationStyle {
-        //
-        // For popover or non modal presentation
-        return UIModalPresentationStyle.none
+    func makeSettingsViewController() -> DisplayGridSettingsViewController {
+        guard let viewController = storyboard?.instantiateViewController(withIdentifier: "DisplayGridSettingsViewController") as? DisplayGridSettingsViewController else {
+            fatalError()
```
pgruenler: fatalError seems kind of abrupt, is there a way that execution could continue or does a storyboard problem mean that we are in an unstable state?  I didn't find any use of fatalError in the app so far; if the philosophy is to avoid stopping execution, I'm hesitant to introduce that pattern?

philium: @pgruenler In effect we use `fatalError()` all throughout the app; it just shows up as a force unwrap (`!`). I was just reading [a post on the Swift forums](https://forums.swift.org/t/se-0217-the-unwrap-or-die-operator/14107/222) that describes this practice as "a way of more thoughtfully crashing on nil with an actionable error message." 

That being said, I didn't include an actionable message. I could have, but at least using `fatalError()` communicates that the author considered the `nil` case and deemed it nigh impossible. With a force unwrap, the author may have added it out of convenience, not considering the consequences. Does that make sense?

pgruenler: @philium, I see what you mean about `fatalError()` being a step up from forced unwrapping.  However, I would be a proponent of, as much as possible, not using either in our samples.  

@dg0yal, do you have any thoughts on this?  The code looks fine but to me it's a new pattern for our samples.  There are definitely two sides to the use of `fatalError()` and I fall more on the avoid it as much as possible side.

PR #433, opened 2018-07-18
added ArcGIS.bundle to the project's target

dg0yal:
@philium please review
I think it might have gotten dropped at some point (possibly with #416 ?) 
This was causing issues such as magnifier mask , compass location symbol , rotate sketch icon, and other assests the runtime requires not being displayed


dg0yal:
I am ok with moving in that direction. Let's get this fixed up for now, we can change the project to use dynamic framework for the next release

PR #439, opened 2018-07-20
Uses WKWebView for source code

philium:


arcgis-ios-sdk-samples/Content Display Logic/Controllers/SourceCodeViewController.swift
```diff
@@ -13,23 +13,30 @@
...
+    @IBOutlet private weak var contentView: UIView!
+    /// The web view that displays the source code.
+    @IBOutlet private weak var webView: WKWebView!
+    @IBOutlet private weak var toolbarTitleButton: UIBarButtonItem!
     
     private var listViewController:ListViewController!
     private var selectedFilenameIndex = 0
     var filenames:[String]!
     
-    private var isListViewContainerVisible = false
```
philium: I removed these properties because they aren't being used.

PR #444, opened 2018-07-23
Uses static Brest buildings service URL

philium:


philium:
@esreli Thanks. Not sure how I overlooked one but not the other.

PR #446, opened 2018-07-24
Uses Implicit Member Expressions for NSAttributedStringKey

philium:


philium:
@njarecha We can't use implicit member expressions for the dictionary values. The type of `Value` is `Any`, so the compiler doesn't know to which type the member belongs.

screen shot 2018-07-24 at 2 46 02 pm

njarecha:
@philium Ok

PR #449, opened 2018-07-25
Use new KVO API for Draw Status sample

philium:
I thought for sure this was only available on iOS 11, but I was wrong. Sweet!

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Display drawing status/MapViewDrawStatusViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Display drawing status/MapViewDrawStatusViewController.swift
@@ -21,6 +21,8 @@ class MapViewDrawStatusViewController: UIViewController {
     @IBOutlet private weak var activityIndicatorView:UIView!
     
     private var map:AGSMap?
+    /// The observation of the map view's draw status.
+    private var drawStatusObservation: NSKeyValueObservation?
     
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Maps/Display drawing status/MapViewDrawStatusViewController.swift
```diff
@@ -45,22 +47,15 @@ class MapViewDrawStatusViewController: UIViewController {
...
+                self?.activityIndicatorView.isHidden = mapView.drawStatus == .completed
+            }
         }
     }
     
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
-        
-        mapView.removeObserver(self, forKeyPath: #keyPath(AGSGeoView.drawStatus))
+        drawStatusObservation = nil
```
nixta: Shouldn't you call `invalidate()` before setting to `nil`? In iOS 11 this isn't necessary but I think it's still required for iOS 10.

philium: I don't think so. From the documentation:

> when the returned NSKeyValueObservation is deinited or invalidated, it will stop observing

And from [the implementation](https://github.com/apple/swift/blob/3f859a3ffd6ee22c765449b773b84fa767a4118f/stdlib/public/SDK/Foundation/NSObject.swift#L159):

> invalidate() will be called automatically when an NSKeyValueObservation is deinited

nixta: Given our discussion elsewhere, do you think this should include an `invalidate()` for as long as we support iOS 10?

PR #451, opened 2018-07-26
Update Line of Sight (GeoElement) 

nixta:
Match macOS sample with MeasurementFormatter.


arcgis-ios-sdk-samples/Analysis/Line of sight (geoelement)/LineOfSightGeoElementViewController.swift
```diff
@@ -160,6 +159,11 @@ class LineOfSightGeoElementViewController: UIViewController {
         }
     }
 
+    // Clean up when done with the sample
+    deinit {
+        losObserver?.invalidate()
```
philium: This is unnecessary.

PR #452, opened 2018-07-26
Generate Offline Map: Adds alert when individual layers have errors

philium:
Also updates the readme with further information.


dg0yal:
@philium code looks good. But it would be nice to add a comment to the new block of code explaining what its doing. Something like - `//Checking if any of the layers in the map encountered error while going offline`. 

Also, just fyi, wanted to point out that there is a `tableErrors` property that will hold any errors for non-spatial tables in the map. The map used in the sample probably doesn't have any such tables so I wouldn't expect it to have any errors, but that is something users will also have to check for completeness. Wanted to mention it so that you can think about whether the sample should be showing that or not

philium:
@dg0yal Thank you! I didn't realize that. I'll create another PR with your suggestions.

PR #455, opened 2018-08-02
Uses the ArcGIS framework

philium:
As opposed to the ArcGIS static library packaged to look like a framework. This prevents us from having to pull in the resources bundle separately.


dg0yal:
I looked through the changes to see if there were any signs of hardcoded paths from your machine, but didn't find any. So the project should still be portable and able to work on any machine where the SDK has been installed to the default location. Nice!

PS - you'll have to show me how you did that. Everytime I try to configure the dynamic framework it puts a specific path in like `/Users/dive....`

philium:
@dg0yal I manually edited the path in the project file to use `~` rather than `/Users/…` and to enclose it in quotation marks.

PR #462, opened 2018-08-15
Fixes build warnings generated when building in Xcode 10.

philium:
These changes are backwards compatable with Xcode 9.4.


philium:
@rolson Thank you for the fast response!

PR #464, opened 2018-08-16
Updates the Feature Layer Extrusion sample

philium:
This brings it in line with the macOS [implementation](https://github.com/Esri/arcgis-runtime-samples-macos/tree/v.next/arcgis-runtime-samples-macos/Scenes/Feature%20layer%20extrusion). Includes changes to the extrusion mode and population density expression.

### DIFF
--- a/arcgis-ios-sdk-samples/Extensions/Foundation/URL.swift
+++ b/arcgis-ios-sdk-samples/Extensions/Foundation/URL.swift
@@ -26,6 +26,10 @@ extension URL {
     
     // Map Services
     
+    /// The url of a map service containing sample census data of the United States.
+    static let censusMapService = URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer")!
+    /// The url of the States layer of the Census Map Service.
+    static let censusMapServiceStatesLayer = censusMapService.appendingPathComponent("3")
     /// The url of a map service containing sample data of the United States.
     static let unitedStatesMapService = URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/USA/MapServer")!
     
--- a/arcgis-ios-sdk-samples/Scenes/Feature layer extrusion/FeatureLayerExtrusionViewController.swift
+++ b/arcgis-ios-sdk-samples/Scenes/Feature layer extrusion/FeatureLayerExtrusionViewController.swift
@@ -54,6 +54,9 @@ class FeatureLayerExtrusionViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         
+        // Adds the source code button item to the right of navigation bar.
+        (self.navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["FeatureLayerExtrusionViewController"]
+        
         sceneView.scene = scene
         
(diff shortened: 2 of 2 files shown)


arcgis-ios-sdk-samples/Extensions/Foundation/URL.swift
```diff
@@ -26,6 +26,10 @@ extension URL {
     
     // Map Services
     
+    /// The url of a map service containing sample census data of the United States.
```
pgruenler: I'd prefer these 4 lines to be back in the view controller to keep the sample more self contained.

philium: @pgruenler We are in the process of moving reused URLs to this file so as not to repeat ourselves. The sample apps reference many URLs that are used in multiple samples. A followup PR (similar to https://github.com/Esri/arcgis-runtime-samples-macos/pull/151) will update the other samples that use these URLs to use the static ones defined here.

pgruenler: I had a feeling that code reuse was the reason for the change.  I think that is going in the wrong direction.  Code reuse is a good thing and a best-practice that it is good for us to teach.  However, in this case I would rather see the service information in the view controller file.  If it's not it the file then a user can't copy and paste the code from the samples app into there own app and get it running.  To me that has always been one of the great things about the samples app.  Is there a way for a user to get the service information from the app without going to github?

philium: Not that I know of. But if a user is already working on their own app, won't they already have their own services?

philium: The flip side is that all sample service information is located in one place. Users could try out various different samples services just copying the `URL.swift` file.

pgruenler: If you @dg0yal are happy with the change I'll approve it as-is.  The code looks fine.

To you answer you question:
>won't they already have their own services?

I don't think that is a given.  I have always thought that these samples would be very useful to a person that has just created a new project and wants to get started learning about something our API can do.
If they do have their own services I think it's kind of nice to have an example of the URL for them to look at when they are changing the code to use their own service.  The URL is a bit complicated.
That is my two-cents.

dg0yal: I prefer to see the URLs inline with the code. It makes it much more readable and approachable to users. Many of us take it for granted but users don't know what a URL to a feature service layer or scene service or a map sevice looks like. Things should be easy to follow and in plain sight. We are trying to tell a story with the code - the story of how simple, logical, and intiutive our API is. If we need to update URLs, its just as simple to find+replace. I don't want to hold up the PR for this because I see many other samples have started to go down this path, but I want us to revisit it. 

Along the lines of keeping things simple and easy to follow, I much prefer the simpler `if` construct that was used previously to set expression instead of the new `Statistic`. It is yet another layer of code that needs to be understood to see what is really going on.

philium:
@saratchandrakarumuri Good catch! I'll fix that.

philium:
@pgruenler Could you please re-approve this PR?

PR #477, opened 2018-09-06
Improves the generate offline map sample

qsmmorgan:
Changes inspired by work on the macOS implementation of this sample

Changes runtime objects to be optionals rather than implicitly unwrapped optionals
Changes copyright date to 2018
Removes the defaultParameters function and incorporates the functionality into the generate button action
More uniformly handles errors
No longer shows an error if the user cancels authentication
Puts the offline geodatabase URL constructor into a separate function which also utilizes an ISO8601DateFormatter


philium:
Are there any more changes from the macOS implementation that need to be incorporated?

qsmmorgan:
@philium Yes, I'm working on it now that the macOS sample was accepted.

arcgis-ios-sdk-samples/Maps/Generate offline map/GenerateOfflineMapViewController.swift
```diff
@@ -62,98 +62,77 @@ class GenerateOfflineMapViewController: UIViewController, AGSAuthenticationManag
...
-                SVProgressHUD.showError(withStatus: error!.localizedDescription)
+            if let error = error{
+                if (error as NSError).code != NSUserCancelledError{
+                    //show error
+                    SVProgressHUD.showError(withStatus: error.localizedDescription)
+                }
+                return
+            }
+            
+            guard let strongSelf = self else{
```
philium: This should probably be the first thing done by the completion handler. If `self` is `nil`, there's no point in showing an error.

qsmmorgan:
I'm made all the improvements I can think of for the moment. This PR is ready for review.

arcgis-ios-sdk-samples/Maps/Generate offline map/GenerateOfflineMapViewController.swift
```diff
@@ -190,91 +169,135 @@ class GenerateOfflineMapViewController: UIViewController, AGSAuthenticationManag
...
-        let maxPoint = self.mapView.screen(toLocation: CGPoint(x: frame.origin.x+frame.width, y: frame.origin.y+frame.height))
-        let extent = AGSEnvelope(min: minPoint, max: maxPoint)
-        return extent
+        return documentDirectoryURL.appendingPathComponent("\(formattedDate).geodatabase")
     }
 
     deinit {
         
-        guard let progress = self.generateOfflineMapJob?.progress else {
+        guard let generateOfflineMapJob = generateOfflineMapJob else {
```
philium: Actually, I think the way it was before was better. The job is only used to get at the progress, so it make sense to keep a reference to the progress rather than the job.

arcgis-ios-sdk-samples/Maps/Generate offline map/GenerateOfflineMapViewController.swift
```diff
@@ -190,91 +169,135 @@ class GenerateOfflineMapViewController: UIViewController, AGSAuthenticationManag
...
         }
+        let progress = generateOfflineMapJob.progress
         
         let isCompleted = (progress.totalUnitCount == progress.completedUnitCount)
         let isCancelled = progress.isCancelled
         
         if !isCancelled && !isCompleted {
             //remove observer
-            self.generateOfflineMapJob?.progress.removeObserver(self, forKeyPath: #keyPath(Progress.fractionCompleted))
+            generateOfflineMapJob.progress.removeObserver(self, forKeyPath: #keyPath(Progress.fractionCompleted))
```
philium: You already have a reference to the progress, so use that instead of going through the job.

arcgis-ios-sdk-samples/Maps/Generate offline map/GenerateOfflineMapViewController.swift
qsmmorgan: Right, that is how it works in the macOS sample. Missed it here.

arcgis-ios-sdk-samples/Maps/Generate offline map/GenerateOfflineMapViewController.swift
```diff
@@ -62,98 +62,77 @@ class GenerateOfflineMapViewController: UIViewController, AGSAuthenticationManag
...
         
         //disable the bar button item until the map loads
-        self.mapView.map?.load { [weak self] (error) in
+        mapView.map?.load { [weak self] (error) in
             
-            guard error == nil else {
-                
-                //show error
-                SVProgressHUD.showError(withStatus: error!.localizedDescription)
+            guard let strongSelf = self else{
```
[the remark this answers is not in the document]
qsmmorgan: Thanks for the tip, I'll start using this.

arcgis-ios-sdk-samples/Maps/Generate offline map/GenerateOfflineMapViewController.swift
```diff
@@ -248,15 +248,15 @@ class GenerateOfflineMapViewController: UIViewController, AGSAuthenticationManag
     private func showLoginQueryAlert() {
         
         let alertController = UIAlertController(title: nil, message: "This sample requires you to login in order to take the map's basemap offline. Would you like to continue?", preferredStyle: .alert)
-        let yesAction = UIAlertAction(title: "OK", style: .default) { [weak self] (action) in
+        let loginAction = UIAlertAction(title: "Login", style: .default) { [weak self] (action) in
```
philium: FYI: Since you're presenting the alert controller right away, you don't need a `weak` reference to `self` in the handler. `self` cannot go away while the alert controller is presented. And when it is dismissed, the alert controller is released, breaking any retain cycle.

arcgis-ios-sdk-samples/Maps/Generate offline map/GenerateOfflineMapViewController.swift
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2017 Esri.
+// Copyright 2018 Esri.
```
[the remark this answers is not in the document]
qsmmorgan: Thanks for the clarification, I'll account for that in the future.

PR #480, opened 2018-09-10
Replaces the custom search UI with UISearchController

qsmmorgan:
Also:

- Adds searching of sample names and descriptions, instead of just readmes
- Adds a computed `readmeURL` property to `Sample` for use by both the search engine and the readme display
- Allows search beyond one word
- Adds search ranking by index and location (name, description, or readme)
- Renames `SearchEngine` to `SampleSearchEngine`, for clarity


qsmmorgan:
Updated with related search improvements:

- Searching that accounts for sample names and descriptions
- Relevancy rankings in search
- A `readmeURL` property on `Node` that stores the URL of the ReadMe file for re-use

esreli:
Is there an issue that this PR addresses?

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -26,33 +26,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
...
+ 
     func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey : Any] = [:]) -> Bool {
         if url.absoluteString.range(of: "auth", options: [], range: nil, locale: nil) != nil {
             AGSApplicationDelegate.shared().application(app, open: url, options: options)
         }
         return true
     }
 
-    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey : Any]? = nil) -> Bool {
+    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
```
philium: Though it doesn't seem to make a whole log of sense, the default value for launchOptions is part of the API and is part of the autocompleted function signature. Please let it the way it was.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -14,40 +14,74 @@
...
-        if containsSearchResults {
-            self.tableView.tableHeaderView?.removeFromSuperview()
-            self.tableView.tableHeaderView = nil
-        }
-        else {
-            self.headerView = self.tableView.tableHeaderView! as? CustomSearchHeaderView
-            self.headerView.delegate = self
-            self.headerView.hideSuggestionsTable()
-        }
+        searchEngine = SearchEngine(samples: samples)
```
philium: `samples` may or may not be set at this point. It most likely is, but that could change in the future if the caller does something to trigger the view to load. A better way might be to use a property observer of `samples`:
```swift
var samples = [Sample]() {
   didSet {
      searchEngine = SearchEngine(samples: samples)
   }
}
```

arcgis-ios-sdk-samples/Content Display Logic/Controllers/SampleInfoViewController.swift
```diff
@@ -32,24 +32,26 @@ class SampleInfoViewController: UIViewController {
...
-        if self.folderName != nil {
-            self.fetchFileContent(for: self.folderName)
+        if let readmeURL = readmeURL,
+            let html = markdownTextFromFile(at: readmeURL) {
+            displayHTML(html)
         }
     }
     
-    func fetchFileContent(for folderName:String) {
+    func markdownTextFromFile(at url:URL)->String? {
```
philium: Please, PLEASE put some more spaces in there! The colon (`:`) and the arrow (`->`) get lost without spaces. The rule we try to follow is there should always be a space (` `) after every colon (anywhere, not just in a function signature) and there should be a space before and after the arrow.

qsmmorgan:
This PR underwent several significant changes but is now ready for review.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/SampleInfoViewController.swift
```diff
@@ -32,24 +32,26 @@ class SampleInfoViewController: UIViewController {
...
-                    let newContent = regex.stringByReplacingMatches(in: content, options: NSRegularExpression.MatchingOptions(), range: NSMakeRange(0, content.count), withTemplate: "")
-                    self.displayHTML(newContent)
-                }
+        //read the content of the file
+        if let content = try? String(contentsOf: url, encoding: .utf8) {
+            //remove the images
+            let pattern = "!\\[.*\\]\\(.*\\)"
+            if let regex = try? NSRegularExpression(pattern: pattern, options: NSRegularExpression.Options.caseInsensitive) {
+                let newContent = regex.stringByReplacingMatches(in: content,
+                                                                options: NSRegularExpression.MatchingOptions(),
```
philium: Having this much leading whitespace is fairly commonplace in Objective-C, but it feels out of place in Swift. You might consider putting each parameter (including the first one) on a separate line so they aren't indented as much:
```swift
let newContent = regex.stringByReplacingMatches(
    in: content,
    options: NSRegularExpression.MatchingOptions(),
    range: NSMakeRange(0, content.count), withTemplate: ""
)
```
That being said, `NSRegularExpression.MatchingOptions()` is equivalent to the default value of `[]`, so the whole thing could be simplified to:
```swift
let range = NSRange(location: 0, length: content.count)
let newContent = regex.stringByReplacingMatches(in: content, range: range, withTemplate: "")
```

arcgis-ios-sdk-samples/Content Display Logic/Controllers/SampleInfoViewController.swift
```diff
@@ -19,7 +19,7 @@ class SampleInfoViewController: UIViewController {
     /// The web view that displays the readme.
     @IBOutlet private weak var webView: WKWebView!
     
-    var folderName:String!
+    var readmeURL:URL?
```
philium: Thank you for making this a `URL`. That makes much more sense.

arcgis-ios-sdk-samples/Content Display Logic/Model/Sample.swift
```diff
@@ -14,11 +14,19 @@
...
 //
 
-struct Sample: Equatable {
+struct Sample: Hashable {
+    
     var name: String
     var description: String
     var storyboardName: String
     var dependencies: [String]
+    var readmeURL:URL?{
```
philium: Please move this to an extension. That keeps the interface nice and sleek.

arcgis-ios-sdk-samples/Content Display Logic/Model/Sample.swift
```diff
@@ -14,11 +14,19 @@
...
+        return Bundle.main.url(forResource: "README", withExtension: "md", subdirectory: name)
+    }
+    
+    var hashValue: Int{
```
philium: Was the compiler-provided implementation not working for you?

arcgis-ios-sdk-samples/Content Display Logic/Model/SearchEngine.swift
```diff
@@ -14,165 +14,180 @@
...
-    private var isLoading = false
+    private var displayNamesByReadmeWords:[String: [String]] = [:]
+    private var isLoadingReadmeIndex = false
     
-    override init() {
-        super.init()
-        
-        self.commonInit()
-    }
+    let samples:[Sample]
```
philium: I'm just spitballing here, but it might make more sense to store the samples in a dictionary keyed off a unique value, say a UUID. The initializer could take any generic sequence of samples and then use that sequence to construct a dictionary. That dictionary could then be used for efficient lookup, rather than storing names and using those for lookup.

arcgis-ios-sdk-samples/Content Display Logic/Model/SearchEngine.swift
```diff
@@ -14,165 +14,180 @@
...
+                sample.name == name
+            })
         }
-        return nil
     }
     
     //MARK: - Public methods
     
-    func searchForString(_ string:String) -> [String]? {
+    func sortedSamples(matching query:String) -> [Sample] {
```
philium: I would suggest returning a `Set<Sample>` and let the caller decide how they want to sort the results, if at all.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -14,40 +14,74 @@
...
+        let searchController = UISearchController(searchResultsController:nil)
+        searchController.obscuresBackgroundDuringPresentation = false
+        searchController.hidesNavigationBarDuringPresentation = false
+        // send search query updates to the results controller
+        searchController.searchResultsUpdater = self
+        // retain a strong reference for iOS 10
+        self.filterSearchController = searchController
+        
+        let searchBar = searchController.searchBar
+        searchBar.placeholder = "Filter"
```
philium: I think Search would still be an appropriate placeholder. You are just searching through a subset of all samples, namely the samples in the selected category.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -14,40 +14,74 @@
...
+        searchBar.autocapitalizationType = .none
+        // set the color of "Cancel" text
+        searchBar.tintColor = .white
+        
+        if #available(iOS 11.0, *) {
+            // embed the search bar under the title in the navigation bar
+            navigationItem.searchController = searchController
+            
+            // find the text field to customize its appearance
+            if let textfield = searchBar.value(forKey: "searchField") as? UITextField {
```
philium: This is a HUGE hack. Do we really need to modify the search bar's text field? I'd rather we didn't.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -177,22 +211,32 @@ class ContentTableViewController: UITableViewController, CustomSearchHeaderViewD
...
+        let navController = UINavigationController(rootViewController: controller)
         
-        //create the info button and
-        //assign the readme url
+        //add the button on the left on the detail view controller
+        controller.navigationItem.leftBarButtonItem = splitViewController.displayModeButtonItem
+        controller.navigationItem.leftItemsSupplementBackButton = true
+        
+        //present the sample view controller
+        splitViewController.showDetailViewController(navController, sender: true)
```
philium: Please use `showDetailViewController(navController, sender: self)`. That should prevent you from needing to get a reference to the split view controller.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -217,51 +261,31 @@ class ContentTableViewController: UITableViewController, CustomSearchHeaderViewD
...
+        if searchController.isActive,
+            let query = searchController.searchBar.text?.trimmingCharacters(in: .whitespacesAndNewlines),
+            !query.isEmpty{
+            samples = searchEngine!.sortedSamples(matching: query)
         }
-        
-        SVProgressHUD.showError(withStatus: "No match found")
-        
+        else{
+            samples = searchEngine!.samples
```
philium: I would strongly suggestion using a separate property for the found samples rather than blowing away the master list. Yes, you passed it onto `SearchEngine`, but I would argue that that property should be private.

philium:
@qsmmorgan Could I make a suggestion? Try to limit the scope of PRs to one thing at a time. I understand you were trying to work on multiple issues at the same time, but it made this PR more complicated to review. It may seem like it would take longer waiting for one issue to be reviewed before starting on the next one, but combining issues often makes reviews take longer. Does that make sense?

qsmmorgan:
@philium Thanks for the feedback. I've been trying to follow that advice overall. In this case the involved issues were so connected that I felt I would be repeating work otherwise. I'm not sure if that ended up being true, but in any case I'll try harder to separate PRs in the future.

arcgis-ios-sdk-samples/Content Display Logic/Model/Sample.swift
qsmmorgan: It wasn't at the time for some reason but it is now.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
qsmmorgan: Despite both fields searching, I used "Filter" to indicate that this search bar is functionally different than the one on the collection view. Otherwise they look identical. I didn't know there was a difference when I first used the app (with the old search bars that were identical).

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
qsmmorgan: I spent awhile online trying to find the best way to change the background color of a text field embedded in the navigation bar. It's far more of a hack than we would hope, but this was the least bad solution I could find. Otherwise it looks like this on iOS 11+

screen shot 2018-09-13 at 3 34 32 pm

philium: That is weird. I wonder where that blue color is coming from?

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -23,70 +23,84 @@ class CustomFlowLayout:UICollectionViewFlowLayout {
...
+                if let backgroundview = textfield.subviews.first {
+                    backgroundview.backgroundColor = UIColor.white
+                    backgroundview.layer.cornerRadius = 12
+                    backgroundview.clipsToBounds = true
+                }
+            }
+            
+        } else {
+            // embed the search bar in the title area of the navigation bar
+            navigationItem.titleView = searchController.searchBar
```
philium: It is unfortunate to see that this replaces the view's title. Is it possible to add the search bas as a section header? I've done something similar using a search bar a header view of a table view.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
qsmmorgan: It appears to be a computed color derived from the blue navigation bar. See [this example](https://stackoverflow.com/questions/46007260/ios-11-customise-search-bar-in-navigation-bar) with a very similar circumstance and a solution that uses the hack.

arcgis-ios-sdk-samples/Content Display Logic/Model/Sample.swift
esreli: With the introduction of Swift 4.2 `Hashable` conformance is now done automatically.

https://github.com/apple/swift-evolution/blob/master/proposals/0206-hashable-enhancements.md

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
qsmmorgan: The reference is still needed for `controller.navigationItem.leftBarButtonItem = splitViewController.displayModeButtonItem` but I'll change this to `presentingController.showDetailViewController(navController, sender: self)`. Using`self` directly isn't an option because of samples opening from `UISearchController` results.

arcgis-ios-sdk-samples/Content Display Logic/Model/SearchEngine.swift
qsmmorgan: I think the result sort order is just as important as the results themselves, and should be handled by the search engine. This way the caller won't have to worry about the details, the results are ready to display.

philium: But then again should the search engine know how the results are going to be displayed? I don't think so. The caller may decide they want to sort them in descending order or sort by category and then by name. The search engine shouldn't care. It would be one thing if the search engine was ordering the results by ranking, but it is just sorting by name, correct?

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
qsmmorgan: I agree that this isn't ideal, but I didn't want to spend too much time writing version-specific code this since fewer and fewer users will be using versions of iOS prior to 11 going forward. I can revisit it if you really think it's worthwhile.

arcgis-ios-sdk-samples/Content Display Logic/Model/SearchEngine.swift
qsmmorgan: That could be a better solution. How would you suggest constructing UUIDs for the samples?

arcgis-ios-sdk-samples/AppDelegate.swift
qsmmorgan: This was accidental when merging in v.next.

arcgis-ios-sdk-samples/Content Display Logic/Model/SearchEngine.swift
philium: Something like this should do the trick:
```swift
self.samples = Dictionary(uniqueKeysWithValues: samples.lazy.map { (UUID(), $0) })
```

arcgis-ios-sdk-samples/Content Display Logic/Model/SearchEngine.swift
qsmmorgan: Nope, I added a non-trivial ranking system internal to the search engine.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/SampleInfoViewController.swift
qsmmorgan: Thanks, I'll try it out tomorrow.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
philium: I suppose it is fine the way it is.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
philium: It might be better to just remove the navigation bar's custom `barTintColor`. Particularly if we git rid of the blue title bar on macOS. But I think we should get feedback from the rest of the team before getting rid of the blue bar on either platform.

qsmmorgan: I actually don't mind the bar color. In any case, we shouldn't restrict ourselves on the bar color based on the search field.

philium: I don't mind the bar color either. What I do mind is hacking Apple provided components. I would much prefer to remove the custom color over relying on implementation details of `UISearchBar`.

nixta:
Agreed, this is much better. Thanks for doing this.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
nixta: Actually, I prefer the Filter text within a category since it describes what's happening. It might be more intuitive to replace the magnifying glass with a filter icon to match the placeholder text, something like this (but obviously with a better icon):

nixta: Or, as with the HIG Stock Search sample, provide some text to indicate you're only looking for samples in the current category:

PR #481, opened 2018-09-11
Renames the "Open an existing map" sample to "Open map (URL)"

qsmmorgan:
Also cleans up the sample somewhat


qsmmorgan:
See also #482

PR #482, opened 2018-09-11
Allows the user to re-select the first sample in the open map (URL) sample

qsmmorgan:
Also defaults the table selection to the default map in the same sample


arcgis-ios-sdk-samples/Maps/Open map (URL)/OpenMapURLViewController.swift
```diff
@@ -57,10 +57,17 @@ class OpenMapURLViewController: UIViewController, UITableViewDataSource, UITable
         tableView.estimatedRowHeight = 50
         tableView.rowHeight = UITableViewAutomaticDimension
         
+        tableView.selectRow(at: IndexPath(row: defaultMapIndex, section: 0), animated: false, scrollPosition: .none)
```
philium: Thank you! Not having the initial map selected was a minor annoyance.

PR #486, opened 2018-09-14
Refactors ContentCollectionViewController to inherit from UICollectionViewController

qsmmorgan:
Also updates the category collection item size method to account for the safe area so that the iPhone X notch is respected in landscape

### DIFF
--- a/arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
+++ b/arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
@@ -14,39 +14,21 @@
 
 import UIKit
 
-class CustomFlowLayout:UICollectionViewFlowLayout {
-    
-    override func shouldInvalidateLayout(forBoundsChange newBounds: CGRect) -> Bool {
-        return true
-    }
-}
+class ContentCollectionViewController: UICollectionViewController, UICollectionViewDelegateFlowLayout {
 
-private let reuseIdentifier = "CategoryCell"
-
-class ContentCollectionViewController: UIViewController, UICollectionViewDataSource, UICollectionViewDelegate, UICollectionViewDelegateFlowLayout {
-
-    @IBOutlet private var collectionView:UICollectionView!
     @IBOutlet private var collectionViewFlowLayout:UICollectionViewFlowLayout!
     
     /// The categories to display in the collection view.
-    var categories:[Category] = []
-
-    private var transitionSize:CGSize!
+    var categories:[Category] = []{
+        didSet{
+            // add search only after setting categories to ensure that the samples are available
+            addSearchController()
+        }
+    }
     
     // strong reference needed for iOS 10
     var searchController:UISearchController?
     
-    override func viewDidLoad() {
-        super.viewDidLoad()
-        
-        if #available(iOS 11.0, *) {
-            collectionView.contentInsetAdjustmentBehavior = .never
-        } else {
-            automaticallyAdjustsScrollViewInsets = false
-        }
-        
-        addSearchController()
-    }
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         // ensure that the search results appear beneath the navigation bar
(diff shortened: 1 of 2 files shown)


arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -175,12 +156,21 @@ class ContentCollectionViewController: UIViewController, UICollectionViewDataSou
...
         }
         return CGSize(width: width, height: width)
     }
     
 }
+
+class CustomFlowLayout:UICollectionViewFlowLayout {
+    
+    override func shouldInvalidateLayout(forBoundsChange newBounds: CGRect) -> Bool {
+        return true
```
philium: This isn't needed anymore. Could you please remove it, along with our implementation of `ContentCollectionViewController.viewWillTransition(to:with:)`? Unless you can find a case where it is still needed.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -141,24 +119,27 @@ class ContentCollectionViewController: UIViewController, UICollectionViewDataSou
...
     
     func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize {
-        
-        let size = self.itemSizeForCollectionViewSize(self.collectionView.bounds.size)
-        
-        return size
+        var collectionViewSize = collectionView.bounds.size
+        if #available(iOS 11.0, *) {
+            //account for the safe area when determining the item size
+            collectionViewSize.height -= collectionView.safeAreaInsets.bottom + collectionView.safeAreaInsets.top
```
philium: Please use `CGRect.inset(by:)`:
```swift
let collectionViewSize: CGSize
if #available(iOS 11.0, *) {
    //account for the safe area when determining the item size
    collectionViewSize = collectionView.bounds.inset(by: collectionView.safeAreaInsets).size
} else {
    collectionViewSize = collectionView.bounds.size
}
```

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -141,24 +119,27 @@ class ContentCollectionViewController: UIViewController, UICollectionViewDataSou
...
+            collectionViewSize.width -= collectionView.safeAreaInsets.left + collectionView.safeAreaInsets.right
+        }
+        return itemSizeForCollectionViewSize(collectionViewSize)
```
philium: Since `itemSizeForCollectionViewSize(_:)` is only called from here, and because the method is doing what this delegate callback is supposed to be doing, consider just inlining it here. I'm not sure there is any benefit to having it as a separate method.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -14,39 +14,21 @@
...
-            collectionView.contentInsetAdjustmentBehavior = .never
-        } else {
-            automaticallyAdjustsScrollViewInsets = false
-        }
-        
-        addSearchController()
-    }
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         // ensure that the search results appear beneath the navigation bar
```
philium: Is modifying `definesPresentationContext` in `viewWillAppear(_:)` and `viewWillDisappear(_:)` still needed after making these changes?

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
qsmmorgan: iOS 10 seems to still need these in order to resize the items when rotating the device. This is only a minor visual nicety and such an edge case that I'm going to take these methods out anyway.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
qsmmorgan: Unfortunately yes, or else the search field cannot be selected on iOS 10. I'll add `#available(iOS 11.0, *)` checks to make it a bit clearer.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -29,19 +29,29 @@ class ContentCollectionViewController: UICollectionViewController, UICollectionV
...
-    override func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
-        // ensure that the search results appear beneath the navigation bar
-        definesPresentationContext = true
-    }
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
-        // required in iOS 10 for the filter field to be interactable in the samples table
-        definesPresentationContext = false
+        guard #available(iOS 11.0, *) else {
```
philium: Wow. This is a fascinating use of `guard`. It most certainly is not an early return, but I understand why you did it this way. There's no `if !#available(…) { … }`, so using an `if` statement would mean an empty `if` clause. However, I still think an if statement is best. This really is not how `guard` is intended to be used. What would you think of this instead?
```swift
if #available(iOS 11.0, *) {} else {
    // required in iOS 10 for the filter field to be interactable in the samples table
    definesPresentationContext = false
}
```

qsmmorgan: I suppose this `guard` statement is somewhat uncommon but I prefer it to an empty `if`. If you're set against it I'll change it. I do wish there was a cleaner way to positively specify old versions, an "earlier" rather than a "later" check.

qsmmorgan: On second thought why don't I change it to `if` but leave a comment in the empty statement. That way it won't have to be empty!

philium: That would be fine as well. I've done something similar in the past.

PR #488, opened 2018-09-17
Adds missing import statement

philium:
The file uses `URL`, but it wasn't being imported.


philium:
> Why is this necessary?

@nixta Good question. It isn't necessary per se, but it is helpful. The import currently isn't needed because UIKit (and therefore Foundation) is implicitly imported by importing `SVProgressHUD.h` in the bridging header. 

Technically we don't need to import `UIKit` anywhere in the project, but we do because it is the right thing to do. `SVProgressHUD` could be swiftified some day or go away entirely. Source files may be moved around or copied to other projects. We guard against those changes potentially breaking the build by explicitly importing dependencies in our source files.

Does that make sense?

PR #492, opened 2018-09-22
remove /en/, http

jf990:
- remove /en/ url path (no longer used and redirected)
- use https


philium:
@jf990 Generally we don't merge into `master` unless we're doing a release. Do you feel your changes need to go into `master`? If so, please create a second pull request to merge these changes into `v.next`. If not, lets update the target branch to `v.next` and they'll get pulled into `master` for the next release.

jf990:
@philium none of these changes are critical, so I'll put it on `v.next`. thanks.

PR #494, opened 2018-09-24
Keeps collection view cells inside the safe area

philium:
Before | After
------ | -----
screen shot 2018-09-21 at 2 57 34 pm | screen shot 2018-09-24 at 2 49 05 pm


philium:
@qsmmorgan I thought there was too, but I couldn't find it. And even if there was, we couldn't use it as long as our deployment target is below iOS 11.0.

qsmmorgan:
Turns out it's in the Size inspector (ruler icon), not the Attributes inspector.
screen shot 2018-09-24 at 3 15 10 pm

And IB seems to deal with the availability on its own for properties like this, so I think you can use it.

philium:
Thanks. I'll do that.

PR #495, opened 2018-09-27
Adds the Generate Offline Map (Overrides) sample

qsmmorgan:


arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/GenerateOfflineMapOverridesViewController.swift
```diff
@@ -0,0 +1,340 @@
...
+            SVProgressHUD.dismiss()
+            
+            //will need the parameters for creating the job later
+            self.parameters = parameters
+            
+            //build the parameter overrides object to be configured by the user
+            offlineMapTask.generateOfflineMapParameterOverrides(with: parameters, completion: {[weak self] (parameterOverrides, error) in
+                
+                if let error = error {
+                    SVProgressHUD.showError(withStatus: error.localizedDescription)
```
philium: We are trying to get away from using `SVProgressHUD` for showing errors. Could you please use a `UIAlertController` instead?

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
```diff
@@ -0,0 +1,408 @@
...
+    }
+}
+private class BoolObject {
+    var bool: Bool
+    init(_ bool: Bool){
+        self.bool = bool
+    }
+}
+
+class OfflineMapParameterOverridesViewController: UITableViewController {
```
philium: I'm curious: why didn't you use static cells? It seems like that would have been easier than designing the table view in code.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/GenerateOfflineMapOverridesViewController.swift
```diff
@@ -0,0 +1,340 @@
...
+//
+
+import UIKit
+import ArcGIS
+
+class GenerateOfflineMapOverridesViewController: UIViewController, AGSAuthenticationManagerDelegate {
+
+    @IBOutlet weak var mapView: AGSMapView!
+    @IBOutlet weak var extentView: UIView!
+    @IBOutlet weak var barButtonItem: UIBarButtonItem!
```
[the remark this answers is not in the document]
qsmmorgan: That name was a holdover from the basic generate offline map sample. I'll change it!

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/GenerateOfflineMapOverridesViewController.swift
```diff
@@ -0,0 +1,340 @@
...
+        }
+        
+        let downloadDirectory = getNewOfflineGeodatabaseURL()
+        
+        let generateOfflineMapJob = offlineMapTask.generateOfflineMapJob(with: parameters,
+                                                                         parameterOverrides: parameterOverrides,
+                                                                         downloadDirectory: downloadDirectory)
+        self.generateOfflineMapJob = generateOfflineMapJob
+        
+        progressObservation = generateOfflineMapJob.progress.observe(\.fractionCompleted, options: .new) { (_, _) in
```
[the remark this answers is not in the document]
qsmmorgan: Yes! I did!

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/GenerateOfflineMapOverridesViewController.swift
```diff
@@ -0,0 +1,340 @@
...
+                //update progress label
+                self.progressLabel.text = progress.localizedDescription
+                
+                //update progress view
+                self.progressView.progress = Float(progress.fractionCompleted)
+            }
+        }
+        
+        //unhide the progress parent view
+        progressParentView.isHidden = false
```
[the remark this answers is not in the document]
qsmmorgan: This was also a holdover from the other sample but I'm not sure what sort of view you're thinking of.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/GenerateOfflineMapOverridesViewController.swift
qsmmorgan: With pleasure.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
qsmmorgan: …because I didn't know that was an option haha. Can you reuse static cells? It's nice to only have two models.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
```diff
@@ -0,0 +1,263 @@
...
+                       basemapExtentBufferSlider,
+                       minHydrantFlowRateSlider] {
+            updateTextField(for: slider!)
+        }
+    }
+    
+    //MARK: - Cancelling
+    
+    /// The completion handler to run if the user clicks cancel
+    var cancelHandler: ((OfflineMapParameterOverridesViewController)->())?
```
philium: It used to be ambiguous whether to use `()` or `Void` for a function that doesn't return anything, but Swift finally settled on using [void](https://developer.apple.com/documentation/swift/void).

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
```diff
@@ -0,0 +1,263 @@
...
+    }
+    
+    private func evaluateLayerVisiblity(){
+        
+        func excludeLayerFromDownload(named name: String) {
+            if let layer = operationalMapLayer(named: name),
+                let serviceLayerID = serviceLayerID(for: layer),
+                let parameters = getGenerateGeodatabaseParameters(forLayer: layer){
+                // Remove the options for this layer from the parameters
+                parameters.layerOptions.removeAll { return $0.layerID == serviceLayerID }
```
philium: You know that the `return` keyword isn't needed here, right? It isn't wrong; just unnecessary.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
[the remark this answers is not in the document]
qsmmorgan: It's just a matter of personal preference. I don't really like delegates. Closures are nice because the model is "here's some code to run when an event happens". Delegates involve declaring new types and writing conforming functions that could theoretically have other callers and it's all just more complicated to me.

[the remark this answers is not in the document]
qsmmorgan: Of course when there are lots of possible callbacks, like `UITableViewDelegate`, it makes a lot of sense. Anyway, I'll change it to a delegate if you prefer since that model is prevalent in the rest of the app.

qsmmorgan: Oh gotcha, I knew they were equivalent but I didn't know `Void` was recommend for closure types.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
qsmmorgan: Unintentional oversight.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
```diff
@@ -17,146 +17,78 @@
...
-            cell.setRange(min: 0, max: 1500)
-            cell.doubleObject = minHydrantFlowRate
-            return cell
-        case .cropLayer:
-            let cell = tableView.dequeueReusableCell(withIdentifier: "OfflineMapParamaterSwitchCell") as! OfflineMapParamaterSwitchCell
-            cell.label.text = "Water Pipes"
-            cell.boolObject = cropWaterPipesToExtent
-            return cell
+    private func updateTextField(for slider: UISlider){
+        let numberFormatter = NumberFormatter()
```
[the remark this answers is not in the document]
qsmmorgan: Good catch, thanks!

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
```diff
@@ -0,0 +1,263 @@
...
+        updateTextField(for: sender)
+    }
+    
+    //MARK: - Text field updating
+    
+    private func updateTextField(for slider: UISlider){
+        let numberFormatter = NumberFormatter()
+        if let text = numberFormatter.string(from: slider.value as NSNumber),
+            let superview = slider.superview,
+            let field = superview.subviews.first(where: { $0 is UITextField }) as? UITextField{
```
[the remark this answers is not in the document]
qsmmorgan: I designed it this way so all the sliders can all use the same action. Otherwise we'd need to know what slider action corresponds to what text field, so we'd either need two more actions or three slider outlets as well as three text field outlets. It also scales to adding more slider cells in the future, all without subclassing the table view cells.

I agree it's somewhat less efficient but not noticeably so.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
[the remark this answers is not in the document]
qsmmorgan: True! Delegates are certainly useful in many cases.

I think in this case I'm drawn to closures since the only callbacks are completion handlers. The primary view controller is sending off an asynchronous task and just wants to know when it's done. I think a delegate would be more useful if we wanted to be notified about more of the intervening events.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
```diff
@@ -0,0 +1,263 @@
...
+    //MARK: - Cancelling
+    
+    /// The completion handler to run if the user clicks cancel
+    var cancelHandler: ((OfflineMapParameterOverridesViewController)->())?
+    
+    @IBAction func cancelAction(_ sender: UIBarButtonItem) {
+        // Run the handler
+        cancelHandler?(self)
+        // close the view
+        navigationController?.dismiss(animated: true)
```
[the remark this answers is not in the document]
qsmmorgan: Sounds good.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/GenerateOfflineMapOverrides.storyboard
```diff
@@ -0,0 +1,422 @@
...
+                                                </label>
+                                                <slider opaque="NO" contentMode="scaleToFill" verticalCompressionResistancePriority="1000" contentHorizontalAlignment="center" contentVerticalAlignment="center" minValue="0.0" maxValue="23" translatesAutoresizingMaskIntoConstraints="NO" id="hJV-NB-7t7">
+                                                    <rect key="frame" x="14" y="38" width="347" height="48"/>
+                                                    <connections>
+                                                        <action selector="sliderChangeAction:" destination="324-87-VxW" eventType="valueChanged" id="fel-IF-keS"/>
+                                                    </connections>
+                                                </slider>
+                                                <textField opaque="NO" contentMode="scaleToFill" enabled="NO" contentHorizontalAlignment="left" contentVerticalAlignment="center" placeholder="0" textAlignment="right" minimumFontSize="17" translatesAutoresizingMaskIntoConstraints="NO" id="dQZ-Fa-xUC">
```
philium: Do you allow the user to enter their own value? If not, these should really be labels rather than textfields.

This is one of those `AppKit`/`UIKit` differences surprised me. I was blown away to find that there was no `NSLabel` class in `AppKit`. All labels on macOS are just textfields configured a certain way.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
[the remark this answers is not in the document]
philium: You would need more outlets and/or more action methods, but I think it would be worth it. I was confused the first time I saw this method. And when I did figure out what was going on, I was even more confused. The more I think about it, the more and more it looks like a hack. Adding outlets for the value labels and action methods for each slider takes a little more time, but it saves a subview lookup *every single time the slider value changes*. Is the lookup a big deal in this case? No. Is it a bad practice in general? Yes.

I'm not opposed to using hacks, but I am opposed to using them if there's another way. That being said, the sample does work and the code in question isn't demonstrating any of the `ArcGIS` API, so I'll approve it the way it is.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/GenerateOfflineMapOverrides.storyboard
qsmmorgan: Changed in the latest commit.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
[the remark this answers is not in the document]
qsmmorgan: I changed this to use outlets since I had to differentiate the labels anyway when I changed them to also use UILabel. I had forgotten that I already had slider outlets so it wasn't as much added code as I had imagined.

qsmmorgan:
Ready for re-review. I believe I addressed all the changes.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
```diff
@@ -0,0 +1,274 @@
...
+            }
+        }
+        else if sender == maxScaleLevelSlider {
+            // Disallow a max value less than the minimum
+            if sender.value < minScaleLevelSlider.value {
+                sender.value = minScaleLevelSlider.value
+            }
+        }
+        // Update this slider's text field to display the new value
+        updateTextField(for: sender)
```
philium: Since you already know which slider you've gat (tested in the `if…else` statement), you might consider just updating the corresponding label's text inside the `if` and `else if` clauses to prevent testing the slider again.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
```diff
@@ -0,0 +1,274 @@
...
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // Update each text field to display its slider's default value
+        for slider in [minScaleLevelSlider,
+                       maxScaleLevelSlider,
+                       basemapExtentBufferSlider,
+                       minHydrantFlowRateSlider] {
+            updateTextField(for: slider!)
```
philium: I'm torn over this. On one hand using a `for` loop is nice because you don't have to write the function call multiple times. On the other hand, it means doing an unnecessary comparison test for each slider. You already know which label corresponds to which slider, so if it were me, I would just update the label's value directly. I suppose you could have an `updateLabel(from slider: Slider)` function, but it would only be a single line.

qsmmorgan:
@saratchandrakarumuri This is ready for testing.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/OfflineMapParameterOverridesViewController.swift
esreli: What about something like this? This is easy to read and reduces doing unnecessary comparison for each slider.

```swift
override func viewDidLoad() {
    super.viewDidLoad()
    
    typealias Parameters = (slider: UISlider, label: UILabel)
    
    // Update each text field to display its slider's default value
    let parameters: [Parameters] = [(minScaleLevelSlider, minScaleLevelLabel),
                                    (maxScaleLevelSlider, maxScaleLevelLabel),
                                    (basemapExtentBufferSlider, extentBufferLabel),
                                    (minHydrantFlowRateSlider, minHydrantFlowRateLabel)]
    
    for parameter in parameters {
        parameter.label.text = numberFormatter.string(from: parameter.slider.value as NSNumber)
    }
}
```

esreli: You also could build a dictionary: `[UISlider: UILabel]` and reduce the overhead found in: `private func updateTextField(for slider: UISlider)`. Just some suggestions!

saratchandrakarumuri:
Couldn't verify this sample due to this [issue](https://devtopia.esri.com/runtime/common-samples/issues/933). Waiting for some update on what is going on wrong with these samples

qsmmorgan:
@saratchandrakarumuri I have been unable to reproduce the errors. I just successfully ran the sample on the latest build.

qsmmorgan:
@philium Could you approve the last commit?

PR #500, opened 2018-10-02
Adds the Display KML sample

qsmmorgan:


arcgis-ios-sdk-samples/Layers/Display KML/DisplayKMLViewController.swift
```diff
@@ -0,0 +1,169 @@
...
+// limitations under the License.
+
+import ArcGIS
+
+class DisplayKMLViewController: UIViewController {
+    
+    @IBOutlet weak var mapView: AGSMapView!
+    
+    /// A model for our three data sources.
+    enum KMLDataSource: String, CaseIterable {
```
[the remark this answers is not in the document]
esreli: If you change the raw type from `String` to `Int` then you have access to an index for the current `displaySource`. This would mean that you wouldn't have to iterate through each cell to find the one with a `reuseIdentifier` that matches `displayedSource`'s raw value.

This change would simplify the code's readability and remove the need for iterating through all cells.

arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
```diff
@@ -429,6 +429,14 @@
 					<string>SubdivisionsShapefile</string>
 				</array>
 			</dict>
+			<dict>
+				<key>displayName</key>
+				<string>Display KML</string>
+				<key>storyboardName</key>
+				<string>DisplayKML</string>
+				<key>descriptionText</key>
+				<string>Show a KML file from a URL, local file, or portal item</string>
```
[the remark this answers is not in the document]
qsmmorgan: Half the descriptions don't have periods and are fragments anyway. We have an issue to review and standardize the descriptions at some point. I'll add a period here.

arcgis-ios-sdk-samples/Info.plist
```diff
@@ -39,56 +39,63 @@
 	<dict>
 		<key>NSExceptionDomains</key>
 		<dict>
-			<key>buienradar.nl</key>
+			<key>arcgis.com</key>
```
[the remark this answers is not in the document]
qsmmorgan: The exceptions for `arcgis.com` and `arcgisonline.com` pre-existed my changes. I alphabetized the exceptions so that's why they appear changed. We can try removing them in a separate PR.

arcgis-ios-sdk-samples/Layers/Display KML/DisplayKMLViewController.swift
```diff
@@ -0,0 +1,169 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // Instantiate a map with a dark gray basemap centered on the United States
+        let map = AGSMap(basemapType: .darkGrayCanvasVector, latitude: 39, longitude: -98, levelOfDetail: 4)
+        // Display the map in the map view
+        mapView.map = map
+        
+        // Set the initial KML source
+        displayedSource = .url
```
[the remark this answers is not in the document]
qsmmorgan: 👍

qsmmorgan:
@esreli Thanks for the feedback! I was able to implement the changes and set the checkmark state without iterating.

qsmmorgan:
@philium Updated the sample for some of the changes we talked about.

esreli:
@qsmmorgan what did you and @philium talk about for [this change](https://github.com/Esri/arcgis-runtime-samples-ios/compare/a71a3ea6e64c89b899372585be68ff1e8e0efeb1..f75072e4488fa0603dd7c1edfaf0774d45cc848c)?

arcgis-ios-sdk-samples/Layers/Display KML/DisplayKMLViewController.swift
```diff
@@ -0,0 +1,180 @@
...
+    func adaptivePresentationStyle(for controller: UIPresentationController) -> UIModalPresentationStyle {
+        // prevent the popover from displaying fullscreen in narrow contexts
+        return .none
+    }
+}
+
+class DisplayKMLDataSourcesTableViewController: UITableViewController {
+    
+    fileprivate var initialSelectedIndex = 0
+    fileprivate var selectionHandler: ((KMLDataSource)->Void)?
```
esreli: I'd prefer to see this sample follow a delegate design pattern rather than a selection handler.

arcgis-ios-sdk-samples/Layers/Display KML/DisplayKMLViewController.swift
```diff
@@ -0,0 +1,180 @@
...
+            cell.accessoryType = .checkmark
+            tableView.selectRow(at: indexPath, animated: false, scrollPosition: .none)
+        }
+        return cell
+    }
+    
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        
+        // get the newly selected cell
+        let selectedCell = tableView.cellForRow(at: indexPath)!
```
esreli: I try to avoid implicitly unwrapping optionals. I know this will *never* fail but I try to avoid `!` whenever possible. Can you `guard let` this instead? This won't depend on my PR approval, just specifying my preference.

arcgis-ios-sdk-samples/Layers/Display KML/DisplayKMLViewController.swift
```diff
@@ -0,0 +1,180 @@
...
+        
+        // add a checkmark to the selected cell
+        selectedCell.accessoryType = .checkmark
+        
+        // change the displayed KML source based on the selected cell
+        let displayedSource = KMLDataSource(rawValue: indexPath.row)!
```
esreli: I try to avoid implicitly unwrapping optionals. I know this will *never* fail but I try to avoid `!` whenever possible. Can you `guard let` this instead? This won't depend on my PR approval, just specifying my preference.

qsmmorgan:
@esreli We talked about using a `UITableViewController` subclass rather than using the parent view controller as the table view delegate.

arcgis-ios-sdk-samples/Layers/Display KML/DisplayKMLViewController.swift
qsmmorgan: You're right, that is a better practice.

arcgis-ios-sdk-samples/Layers/Display KML/DisplayKMLViewController.swift
qsmmorgan: I prefer to use a closure when there's only one callback and it's essentially a completion handler. It's simpler than declaring another type, a function prototype, an implementation, and another object.

philium:
@esreli Sorry about that! We didn't intentionally leave you out of the conversation. I figured since we're in the same office, I'd just pop over and talk with @qsmmorgan in person.

arcgis-ios-sdk-samples/Layers/Display KML/DisplayKMLViewController.swift
philium: @qsmmorgan You know where I stand on using delegates. I will say that the more you use them, the easier they become to implement and the more natural they feel.

qsmmorgan: I could say the same about closures haha. If you want me to change it I will.

esreli: It is certainly easier. I'll suggest the delegate pattern for parent-child view controller relationships is pretty pervasive across cocoa and so it would feel more appropriate to implement rather than a callback. Alternatively, the callback pattern is often used with async completion and this is not that. Also, this places the memory burden on the child view controller, something that isn't needed.

arcgis-ios-sdk-samples/Layers/Display KML/DisplayKMLViewController.swift
```diff
@@ -0,0 +1,186 @@
...
+        kmlLayer.load {[weak self] (error) in
+            
+            // Close the progress indicator
+            SVProgressHUD.dismiss()
+
+            if let self = self,
+                let error = error {
+                // Display the error if one occurred
+                let alertController = UIAlertController(title: error.localizedDescription, message: nil, preferredStyle: .alert)
+                alertController.addAction(UIAlertAction(title: "OK", style: .cancel))
```
philium: This is a really minor issue, but I think `default` is the appropriate style for an OK button.

arcgis-ios-sdk-samples/Layers/Display KML/DisplayKMLViewController.swift
```diff
@@ -0,0 +1,186 @@
...
+        /// A KML layer created from an ArcGIS Online portal item
+        let kmlLayer = AGSKMLLayer(item: portalItem)
+        display(kmlLayer: kmlLayer)
+    }
+    
+}
+
+extension DisplayKMLViewController: UIPopoverPresentationControllerDelegate {
+    
+    func adaptivePresentationStyle(for controller: UIPresentationController) -> UIModalPresentationStyle {
```
philium: Per the documentation, please use `adaptivePresentationStyle(for:traitCollection:)` instead.

qsmmorgan: Thanks for the catch! I didn't realize there were two methods.

esreli:
@qsmmorgan what do you plan on doing re: the discussion surrounding the use of delegate vs. closure?

qsmmorgan:
@esreli Oh, I was going to leave it as a closure.

esreli:
@qsmmorgan what design pattern is used elsewhere throughout the app? (delegate vs closure)

qsmmorgan:
@esreli Mostly delegate, I think.

esreli:
@qsmmorgan understood. As a final note on the matter, it's my general preference to maintain a uniform design pattern across an app. And of course, your code is good so perhaps it doesn't matter. 

Good work!

qsmmorgan:
@saratchandrakarumuri @philium @esreli Thanks for reviewing!

PR #502, opened 2018-10-03
Fixes a visual bug in the search bar when opening a view controller

qsmmorgan:


arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -28,29 +28,51 @@ class ContentTableViewController: UITableViewController {
...
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
         
-        if !containsSearchResults{
-            addFilterSearchController()
+        if !containsSearchResults,
+            !hasSearchController {
+            // wait a little to avoid a crash
+            DispatchQueue.main.asyncAfter(deadline: .now()+0.1) { [weak self] () in
```
pgruenler: Is there an event that could be used here instead of just delaying?

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -28,29 +28,51 @@ class ContentTableViewController: UITableViewController {
...
-    var containsSearchResults = false
-    private var bundleResourceRequest:NSBundleResourceRequest!
-    private var downloadProgressView:DownloadProgressView!
+    private var expandedRowIndex: Int = -1
     
-    var searchEngine: SampleSearchEngine?
-    
-    // strong reference needed for iOS 10
-    var filterSearchController:UISearchController?
+    private var bundleResourceRequest: NSBundleResourceRequest!
```
pgruenler: I know this isn't related to the changes you made but it would be nice if these to variables were not implicitly unwrapped optionals.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
qsmmorgan: Not that I could find. Using `viewDidLoad` or `viewWillAppear` both cause the visual glitch. However, it turns out the waiting is unnecessary, though the async is still needed to avoid a crash.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
qsmmorgan: Fixed.

qsmmorgan:
@philium That's okay! It wasn't that much extra work since I was adding search anyway.

qsmmorgan:
@pgruenler This is ready for re-review when you get a chance.

PR #503, opened 2018-10-03
Removes the ArcGIS online domains from the ATS exceptions list

qsmmorgan:


nixta:
The thing to be conscious of is whether any of the sample data is either accessed via HTTP rather than HTTPS (I think not, but haven't checked) but more subtly whether any of the data itself has embedded HTTP references (e.g. a WebMap that references a layer via HTTP, or a KML file with HTTP rather than HTTPS network links). I don't myself have time to check this today, but if you want to investigate, that's what you should look for to verify.

qsmmorgan:
Removing review requests until I can verify all the samples continue to work.

qsmmorgan:
Closing this to start fresh.

PR #504, opened 2018-10-04
Adds the List KML Contents sample

qsmmorgan:


pgruenler:
Did I review this too soon?  I did not realize you were still working in it.

qsmmorgan:
@pgruenler Unfortunately, yes. @philium gave me some feedback in-person and I updated it based on that. It should be ready now.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+    private var kmlDataset: AGSKMLDataset?
+    
+    /// The full node tree flattened to an array of tuples that include each node's level.
+    /// Used as the model for the table view.
+    private var flattenedNodes: [(node: AGSKMLNode, level: Int)] = []
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        guard self.kmlDataset == nil else {
```
philium: I'm a bit confused by this. Is `kmlDataset` ever set elsewhere? If not, this check isn't needed. `viewDidLoad()` is only ever called once per instance, so the expression would never evaluate to `false`.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+        // load the dataset asynchronously so we can list its contents
+        kmlDataset.load {[weak self] (error) in
+            
+            guard let self = self else{
+                return
+            }
+            
+            guard error == nil else{
+                // display the error as an alert
+                let alertController = UIAlertController(title: error?.localizedDescription, message: nil, preferredStyle: .alert)
```
philium: Normally the error's description is part of the alert's message, not the title.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsSceneViewController.swift
```diff
@@ -0,0 +1,168 @@
...
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+class ListKMLContentsSceneViewController: UIViewController {
+    
+    @IBOutlet weak var sceneView: AGSSceneView!
+    
+    weak var kmlDataset: AGSKMLDataset?
```
philium: Neither of these need to be `weak` and they do not hold a strong reference to the view controller or anything with a strong reference to the view controller. In general, a view controller holds a strong reference to its data model.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+        cell.textLabel?.text = node.name
+        cell.detailTextLabel?.text = "\(type(of:node))"
+        // set the indentation based on the node hierarchy
+        cell.indentationLevel = flattenedNodeInfo.level
+        return cell
+    }
+    
+    //MARK: - UITableViewDelegate
+    
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
```
philium: If you wanted to, you could trigger the presentation via a segue configured in the storyboard. Then in `prepareForSegue()`, you would get the node corresponding to the table view's selected index path and assign that to the destination view controller. I think that method is more common, but what you have is just fine too.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+        guard indexPath.section == 0 else{
```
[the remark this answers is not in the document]
qsmmorgan: Thanks, this was leftover from the old implementation.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsSceneViewController.swift
```diff
@@ -0,0 +1,168 @@
...
+    private func sceneSurfaceElevation(for point: AGSPoint) -> Double? {
+        guard let surface = sceneView.scene?.baseSurface else {
+            return nil
+        }
+        
+        var surfaceElevation: Double? = nil
+        let group = DispatchGroup()
+        group.enter()
+        // we want to return the elevation synchronously, so run the task in the background and wait
+        DispatchQueue.global(qos: .userInteractive).async {
```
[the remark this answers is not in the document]
qsmmorgan: I use the group to wait for the inner closure to return, so using `sync` wouldn't work. The purpose of the outer `async` is to keep the thread from blocking since the inner closure is run on the same thread that deploys it.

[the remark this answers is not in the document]
philium: Oh! That makes perfect sense. Never mind.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContents.storyboard
```diff
@@ -0,0 +1,96 @@
...
+        </scene>
+        <!--ListKML Contents Scene View Controller-->
+        <scene sceneID="q0N-YS-N1m">
+            <objects>
+                <viewController storyboardIdentifier="ListKMLContentsSceneViewController" id="5tT-3W-19j" customClass="ListKMLContentsSceneViewController" customModule="arcgis_ios_sdk_samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <view key="view" contentMode="scaleToFill" id="v8D-7R-gRf">
+                        <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <subviews>
+                            <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="This node has no extent." textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="6Jx-DO-V0F">
```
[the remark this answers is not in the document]
qsmmorgan: All the logic that determines whether the scene has an extent or not is within the same view controller. This could potentially be changed but I don't think it's worth it so I'm going to leave it the way it is.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsViewController.swift
qsmmorgan: I don't even remember writing that line haha

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsSceneViewController.swift
qsmmorgan: They're already retained by the parent view controller but I guess there's no harm in making them strong references.

arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
```diff
@@ -429,6 +429,14 @@
 					<string>SubdivisionsShapefile</string>
 				</array>
 			</dict>
+			<dict>
+				<key>displayName</key>
+				<string>List KML contents</string>
+				<key>storyboardName</key>
+				<string>ListKMLContents</string>
+				<key>descriptionText</key>
+				<string>List the contents of a KML file.</string>
```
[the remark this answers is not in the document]
qsmmorgan: That may be better, but I can't find any other samples that use that tense so I'll just leave it for now until we rewrite all the descriptions.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsViewController.swift
qsmmorgan: It's already working this way so I don't feel compelled to change it.

qsmmorgan:
@saratchandrakarumuri This is ready for testing

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsSceneViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+                group.leave()
+            })
+        }
+        // wait, but not longer than three seconds
+        _ = group.wait(timeout: .now() + 3)
+        return surfaceElevation
+    }
+    
+    /// Returns the viewpoint showing the node, converting it from the node's AGSKMLViewPoint if possible.
+    private func viewpoint(for node: AGSKMLNode) -> AGSViewpoint? {
```
philium: What would you think of making this an asynchronous operation so as not to block the main thread while getting the elevation? Something like this;

```swift
/// Creates an `AGSKMLViewpoint` object based on an `AGSKMLViewpoint`
/// object. The completion handler may be called immediately.
///
/// - Parameters:
///   - kmlViewpoint: A KML viewpoint.
///   - completion: A closure called when the viewpoint has been created.
func makeViewpoint(from kmlViewpoint: AGSKMLViewpoint, completion: (AGSViewpoint?) -> Void) {
    …
}

makeViewpoint(from: kmlViewpoint) { [weak self] (viewpoint) in
    guard let self = self, node == self.node, let viewpoint = viewpoint else { return }
    self.sceneView.setViewpoint(viewpoint)
}
```

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsSceneViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+                center = AGSPoint(x: center.x, y: center.y, z: center.z + elevation, spatialReference: extent.spatialReference)
+                // Defaults based on Google Earth.
+                let camera = AGSCamera(lookAt: center, distance: 1000, heading: 0, pitch: 45, roll: 0)
+                // only the camera parameter is used by the scene
+                return AGSViewpoint(targetExtent: extent, camera: camera)
+                
+            }
+            else {
+                // expand the extent to give some margins when framing the node
+                let bufferRadius = [extent.width, extent.height].max()! / 20
```
philium: That is a fascinating way of getting the maximum of two values, but I would recommend you use the global `max(_:_:)` function:
```swift
max(extent.width, extent.height)
```
That way you avoid the force unwrap.

qsmmorgan: Oh right. Global functions exist.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsSceneViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+    
+    //MARK: - Viewpoint
+    
+    /// Sets the viewpoint of the scene based on the node, or hides the node
+    /// if a viewpoint cannot be determined.
+    private func setSceneViewpoint(for node:AGSKMLNode){
+        
+        if !isViewLoaded {
+            // load the view before setting a viewpoint
+            loadView()
```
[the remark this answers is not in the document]
qsmmorgan: Oops! Now I know.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsSceneViewController.swift
qsmmorgan: Good idea. I changed it to execute in the background, though not quite this way.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsSceneViewController.swift
```diff
@@ -0,0 +1,202 @@
...
+    }
+    
+    //MARK: - Viewpoint
+    
+    /// Sets the viewpoint of the scene based on the node, or hides the node
+    /// if a viewpoint cannot be determined.
+    private func setSceneViewpoint(for node:AGSKMLNode){
+        
+        // load the view before setting a viewpoint
+        loadViewIfNeeded()
```
philium: I should have noticed this sooner, but unless I'm missing something, the view doesn't need to be loaded here. Should this be moved to after dispatching back to the main queue, right before you unwrap `self.sceneView`?

Speaking of unwrapping the scene view, I'm curious: why did you make it an optional rather than an IUO? As you know, IUO is kind of the standard for `IBOutlet`s. In addition, it looks like you're rightly making sure that the view is loaded before accessing `self.sceneView`, so using an IUO would prevent you from having to explicitly unwrap it yourself.

arcgis-ios-sdk-samples/Layers/List KML contents/ListKMLContentsSceneViewController.swift
philium: I see what you did. You may not be blocking the main queue, but it looks like you're blocking the global `userInitiated` queue, which may not be the best idea. I assume that getting the elevation at a given point doesn't take an enormous amount of time, but because the queue being blocked is a global queue, who knows what else is being dispatched to it.

qsmmorgan: Are we not supposed to use the globe queues? I can change it to a custom queue, it doesn't matter to me.

philium: > Are we not supposed to use the globe queues?

No! Using the global queues is perfectly fine. It is blocking one of them while waiting on work from another queue that I'm concerned about. Tell me if I've got this right:

1. `setSceneViewpoint(for:)` is called on the main queue.
1. `setSceneViewpoint(for:)` dispatches work to the `userInitiated` global queue.
1. `viewpoint(for:)` is called from the `userInitiated` global queue.
1. `sceneSurfaceElevation(for:)` is called from the `userInitiated` global queue.
1. `sceneSurfaceElevation(for:)` waits (blocks) on work dispatched to the `userInteractive` global queue.
1. `AGSSurface.elevation(for:completion:)` is called from the `userInteractive` global queue. The method is asynchronous, so presumably it dispatches work to some arbitrary queue.
1. The `userInteractive` global queue is unblocked when `AGSSurface.elevation(for:completion:)`'s completion handler is called from some arbitrary queue.
1. When `viewpoint(for:)` returns, work is dispatched to the main queue.

Assuming that is correct, a couple observations:

- The dispatch to the `userInteractive` global queue is unnecessary. The only work done on it calls another asynchronous function on another queue.
- The blocking is unnecessary. `viewpoint(for:)` could be made asynchronous with a completion callback. When elevation is needed, it can be retrieved asynchronously and then the viewpoint could be reported to the completion handler on the main thread.

qsmmorgan: Oh, I misunderstood which dispatch you were referring to. I'm going to leave the blocking, for synchronous fetching of the elevation, but remove the extra call to userInteractive. All the viewpoint work will be done on a queue owned by the view controller.

qsmmorgan:
@saratchandrakarumuri This is ready for re-testing.

PR #506, opened 2018-10-04
Adds the OpenStreetMap layer sample

qsmmorgan:
This is a lower priority than the earlier sample PRs.


arcgis-ios-sdk-samples/Layers/OpenStreetMap layer/OpenStreetMapLayer.storyboard
```diff
@@ -0,0 +1,44 @@
...
+                                <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
+                                <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                            </view>
+                        </subviews>
+                        <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                        <constraints>
+                            <constraint firstItem="Q3F-3f-v07" firstAttribute="leading" secondItem="0Mi-J9-K9g" secondAttribute="leading" id="2L2-ir-zDf"/>
+                            <constraint firstItem="Q3F-3f-v07" firstAttribute="trailing" secondItem="0Mi-J9-K9g" secondAttribute="trailing" id="VDh-KF-tJ7"/>
+                            <constraint firstItem="Q3F-3f-v07" firstAttribute="bottom" secondItem="0Mi-J9-K9g" secondAttribute="bottom" id="ayY-9E-Cpr"/>
+                            <constraint firstItem="Q3F-3f-v07" firstAttribute="top" secondItem="0Mi-J9-K9g" secondAttribute="top" id="qad-fK-gwx"/>
```
philium: Please constrain the top of the map view to the top of the safe area rather than the superview. Constraining to the superview would be nice if we had a translucent navigation bar, but the problem is that the watermark can appear underneath the navigation bar, partially obscuring it.

arcgis-ios-sdk-samples/Layers/OpenStreetMap layer/OpenStreetMapLayerViewController.swift
```diff
@@ -0,0 +1,34 @@
+// Copyright 2016 Esri.
```
philium: The year should be `2018`.

arcgis-ios-sdk-samples/Layers/OpenStreetMap layer/OpenStreetMapLayer.storyboard
qsmmorgan: Good catch. I copied the storyboard from the display a map sample so that one should probably get updated at some point.

PR #516, opened 2018-10-11
Removes SVProgressHUD showError, showSuccess, showInfo, and showImage functions

qsmmorgan:


arcgis-ios-sdk-samples/Analysis/Analyze hotspots/AnalyzeHotspotsViewController.swift
```diff
@@ -99,6 +99,12 @@ class AnalyzeHotspotsViewController: UIViewController, HotspotSettingsVCDelegate
         }
     }
     
+    private func showAlert(error: Error) {
```
philium: I would recommend refactoring this into a convenience initializer of `UIAlertController`:
```swift
private extension UIAlertController {
    convenience init(error: Error) {
        self.init(errorMessage: error.localizedDescription)
    }
    
    convenience init(errorMessage message: String) {
        self.init(title: "Error", message: message, preferredStyle: .alert)
        let okayAction = UIAlertAction(title: "OK", style: .default)
        addAction(okayAction)
        preferredAction = okayAction
    }
}
```
You could then add a similar method to `UIViewController` via an extension:

```swift
private extension UIViewController {
    func showAlert(error: Error) {
        present(UIAlertController(error: error), animated: true)
    }
}
```

But I wouldn't recommend it for a couple reasons:

- The name `show` is similar to but semantically different from `UIViewController.show(_:sender:)`. "present" would be a better choice.
- The caller has no control over whether the presentation is animated.

Both of those issues could be resolved by renaming the method to something like `presentAlertController(for error: Error, animated: Bool)`, but at that point, it would be simpler for the view controller to just create and present the alert controller itself:

```
self.present(UIAlertController(error: error), animated: true)
```

Does that make sense?

arcgis-ios-sdk-samples/Shared resources/Libraries/SVProgressHUD/SVProgressHUD.h
```diff
@@ -118,18 +118,6 @@ typedef void (^SVProgressHUDDismissCompletion)(void);
```
philium: Since this is a third party source file, I wouldn't modify. Unless and until we no longer use it and it can be removed, we can (and should) regularly update it. We've had to do so in the past. Removing these unused methods would needlessly make updating more complicated. Does that make sense?

qsmmorgan: I didn't realize we still updated it. I removed the methods from there so no one would use them in the future, but I don't suppose we need to enforce that at the code level.

arcgis-ios-sdk-samples/Analysis/Analyze hotspots/AnalyzeHotspotsViewController.swift
```diff
@@ -99,6 +99,12 @@ class AnalyzeHotspotsViewController: UIViewController, HotspotSettingsVCDelegate
...
+        let alert = UIAlertController(title: nil, message: error.localizedDescription, preferredStyle: .alert)
+        alert.addAction(UIAlertAction(title: "OK", style: .default))
```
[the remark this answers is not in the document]
qsmmorgan: For some reason I thought that if there was only one action it was preferred by default, but I guess not.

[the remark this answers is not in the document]
philium: That would be nice, wouldn't it? Time for another trip to bugreport.apple.com…

arcgis-ios-sdk-samples/Analysis/Analyze hotspots/AnalyzeHotspotsViewController.swift
qsmmorgan: This makes sense but I'm not sure what the conclusion is. I renamed the `showAlert` functions to `presentAlert` to match the VC method; thanks for that suggestion. In any other app I would suggest making the extensions you mentioned, but in this case I think the simplest solution is to leave the code in private functions within each VC.

philium: My thought is that the code to create and present an alert controller doesn't add anything to the samples themselves. Putting that in a private extension at the bottom of the source file would keep it out of the way of the code that we actually want to show off.

nixta: I agree. If we have an easy way to keep code that doesn't contribute to learning about the Runtime API in the context of the sample out of the way, then we should take it. Either a private extension or at least in a `Helper Methods` section at the very end of the class (as [here](https://github.com/Esri/arcgis-runtime-samples-ios/blob/8e18a910a7704ab34a86152a12f2502690c216fa/arcgis-ios-sdk-samples/Analysis/Statistical%20query%20(group%20and%20sort)/StatisticalQueryGroupAndSortViewController.swift#L393)).

I think in this case, since we're defining two simple methods over and over again (well, one and a convenience version of it), I would even be OK with having a single extension to `UIViewController` shared by all samples, but that might ruffle some feathers.

nixta: Actually, having thought about this some more, I don't think it'll ruffle feathers. We'd be extending `UIViewController` and not one of the runtime classes. Sample code must show the Runtime APIs unmodified. Here we're showing an alert using code that we've written ourselves, so I would suggest a single implementation in a `UIViewController` extension that each sample can use.

qsmmorgan: Changed to an extension in the latest commit.

arcgis-ios-sdk-samples/Extensions/UIKit/UIViewController.swift
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+
+extension UIViewController {
+    
+    func presentAlert(title: String? = nil, message: String) {
```
philium: For what it's worth, based on my experience, it is better to keep view controller creation and presentation separate. One reason is testing. If the alert controller was created through an initializer, it would be easy to write an `testInitWithBlah()` function, create one, and assert that it was initialized properly. By conflating creation with presentation, that test becomes much more complicated. It has to create a another view controller to do the presenting. It has to wait for the animation to finish (assuming it knows the presentation will be animated--more on that in a bit). It has to get the presented view controller and cast it to an alert controller. In the end, it is testing view controller presentation as much as it is testing initialization of the alert controller.

I would encourage you to write tests for this code. And speaking of documentation, please add documentation comments for both of these methods. That will really help anybody else who comes across these methods being called and wonders what they do.

PR #517, opened 2018-10-11
Adds upside down orientation for supported devices

qsmmorgan:
Silences an Xcode warning when building against devices that support upside down orientation

### DIFF
--- a/arcgis-ios-sdk-samples/Info.plist
+++ b/arcgis-ios-sdk-samples/Info.plist
@@ -144,6 +144,7 @@
 		<string>UIInterfaceOrientationPortrait</string>
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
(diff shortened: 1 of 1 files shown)


philium:
I'm not in favor of this change. Most stock iOS apps don't support upside down orientation on iPhone and that is the default for new iOS projects.

If you haven't done so already, please file a bug at https://bugreport.apple.com for the intermittent Xcode warning.

qsmmorgan:
I agree we shouldn't enable upside orientation down for iPhones, but this key doesn't enable that. The system still blocks upside down on devices that shouldn't allow it. Adding the key does allow for upside down on iPads, which is desirable.

philium:
This change most definitely applies to iPhones, not iPads. The iPad key already contains `UIInterfaceOrientationPortraitUpsideDown` (line `152`) and upside down orientation works on iPads with or without this change.

Because we don't override `UIViewController. supportedInterfaceOrientations`'s default of `allButUpsideDown` for the iPhone idiom, this change essentially has no effect. Upside down orientation does not work on iPhones with or without this change.

PR #520, opened 2018-10-12
Updates the Display Grid sample to use a table view for the settings

qsmmorgan:
This is a prerequisite for replacing HorizontalPicker in this sample.


arcgis-ios-sdk-samples/Display information/Display grid/DisplayGridSettingsViewController.swift
```diff
@@ -265,4 +257,10 @@ class DisplayGridSettingsViewController: UIViewController, HorizontalPickerDeleg
             changeLabel(color: selectedColor)
         }
     }
+    
+    // MARK: UITableViewDelegate
+    
+    override func tableView(_ tableView: UITableView, shouldHighlightRowAt indexPath: IndexPath) -> Bool {
+        return false
```
philium: It might be easier to just disable selection on the table view for now:

screen shot 2018-10-12 at 4 08 51 pm

When the horizontal pickers go away, you could then re-enable selection and update the selection style to `None` for cells that should not be selectable.

screen shot 2018-10-12 at 4 10 53 pm

arcgis-ios-sdk-samples/Display information/Display grid/DisplayGridViewController.swift
```diff
@@ -39,26 +40,27 @@ class DisplayGridViewController: UIViewController, UIAdaptivePresentationControl
...
+        let settingsViewController: UIViewController
+        if let viewController = popoverViewController {
             settingsViewController = viewController
         } else {
-            settingsViewController = makeSettingsViewController()
-            self.settingsViewController = settingsViewController
+            settingsViewController = makeSettingsPopoverViewController()
+            self.popoverViewController = settingsViewController
         }
         settingsViewController.preferredContentSize = {
```
philium: It probably would have been easier to leave everything the way it was and then programmatically create the navigation controller here:

```swift
let navigationController = UINavigationController(rootViewController: settingsViewController)
navigationController.preferredContentSize = {
…
```

arcgis-ios-sdk-samples/Display information/Display grid/DisplayGrid.storyboard
```diff
@@ -1,12 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
philium: The console is displaying this warning when the popover is presented:

```[Warning] Warning once only: Detected a case where constraints ambiguously suggest a height of zero for a tableview cell's content view. We're considering the collapse unintentional and using standard height instead.```

That is because none of the cells' subviews have any constraints. Please add enough constraints to determine the location and size of all subviews.

For the cells with switches, you can actually use the `basic` cell type. Add the switch to the scene (NOT the view controller) and then hook it up as the accessory view of the cell:

screen shot 2018-10-12 at 4 23 00 pm

That approach has a few advantages:

- No custom constraints needed.
- The cell looks like a standard cell (size, label position, etc.) because it is a standard cell.
- It Just Works with VoiceOver. Don't believe me? Try it!

arcgis-ios-sdk-samples/Display information/Display grid/DisplayGridViewController.swift
qsmmorgan: You're probably right but I think either is fine.

arcgis-ios-sdk-samples/Display information/Display grid/DisplayGrid.storyboard
qsmmorgan: Thanks for the tip! I was able to convert all the cells to basic this way, not just the switches!

PR #522, opened 2018-10-15
Enhances the alert extension of UIViewController

qsmmorgan:
Follow-up on #516.

Adds UIAlertController setup to a convenience init for use with the UIViewController alert extension

### DIFF
--- a/arcgis-ios-sdk-samples/Extensions/UIKit/UIViewController.swift
+++ b/arcgis-ios-sdk-samples/Extensions/UIKit/UIViewController.swift
@@ -16,16 +16,30 @@ import UIKit
 
 extension UIViewController {
     
+    /// Shows an alert with the given title, message, and an OK button.
     func presentAlert(title: String? = nil, message: String) {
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Extensions/UIKit/UIViewController.swift
```diff
@@ -16,16 +16,30 @@ import UIKit
...
     
 }
+
+extension UIAlertController {
+    
+    /// Initializes the alert controller with the given parameters, adding the
+    /// actions successively and setting the first action as preferred.
+    fileprivate convenience init(title: String? = nil, message: String? = nil, preferredStyle: UIAlertController.Style = .alert, actions: [UIAlertAction] = []) {
+        self.init(title: title, message: message, preferredStyle: preferredStyle)
+        for action in actions {
```
philium: This would be a great place to use `forEach(_:)`:
```swift
actions.forEach(addAction)
```

PR #523, opened 2018-10-15
Updates the Viewshed (location) sample to use a table view for the settings

qsmmorgan:


arcgis-ios-sdk-samples/Analysis/Viewshed (location)/ViewshedLocation.storyboard
```diff
@@ -1,412 +1,424 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
philium: The switch and color picker cells do not have any constraints positioning their subviews. This results in the subviews not being placed properly:

screen shot 2018-10-17 at 11 25 41 am

Please either add constraints or use a non-custom table view cell style.

qsmmorgan: Thanks for finding that!

PR #524, opened 2018-10-16
Converts the settings UI of the Blend Renderer sample to a table view popover

qsmmorgan:


arcgis-ios-sdk-samples/Layers/Blend renderer/BlendRendererViewController.swift
```diff
@@ -19,91 +19,113 @@ import ArcGIS
...
+                if traitCollection.horizontalSizeClass == .regular,
+                    traitCollection.verticalSizeClass == .regular {
+                    height = 250
+                } else {
+                    height = 200
+                }
+                return CGSize(width: 375, height: height)
+            }()
+            controller.presentationController?.delegate = self
+            controller.loadViewIfNeeded()
```
philium: I see. The presenting view controller prevents the crash by loading the view first. I would argue that the settings view controller should Do The Right Thing™ and allow the caller to set these values before or after the view is loaded.

arcgis-ios-sdk-samples/Layers/Blend renderer/BlendRenderer.storyboard
```diff
@@ -29,223 +29,185 @@
...
             <point key="canvasLocation" x="305" y="372"/>
         </scene>
         <!--Blend Renderer SettingsVC-->
-        <scene sceneID="8lg-xS-cBk">
+        <scene sceneID="zuR-Aq-sif">
             <objects>
-                <viewController id="mbD-bO-mFc" customClass="BlendRendererSettingsVC" customModule="arcgis_ios_sdk_samples" customModuleProvider="target" sceneMemberID="viewController">
-                    <view key="view" contentMode="scaleToFill" id="PAD-ae-PsN">
-                        <rect key="frame" x="0.0" y="0.0" width="300" height="220"/>
+                <tableViewController id="VDk-RB-4PB" customClass="BlendRendererSettingsVC" customModule="arcgis_ios_sdk_samples" customModuleProvider="target" sceneMemberID="viewController">
```
philium: Please add constraints to position the subviews of the cells.

screen shot 2018-10-17 at 12 12 59 pm

arcgis-ios-sdk-samples/Layers/Blend renderer/BlendRenderer.storyboard
```diff
@@ -74,117 +74,167 @@
         <scene sceneID="zuR-Aq-sif">
```
philium: This scene is reporting the following errors:

screen shot 2018-10-19 at 3 53 23 pm

The problem is that the provided constraints specify a *minimum* height for the cell, not an *exact* height. The top and bottom constraints should be equal to (not greater than or equal to) and the centerY constraint should be removed.

qsmmorgan: Thanks for catching this! Not sure how I missed the error.

PR #525, opened 2018-10-16
Updates the app name and version

qsmmorgan:


arcgis-ios-sdk-samples/Info.plist
```diff
@@ -4,6 +4,8 @@
 <dict>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
+	<key>CFBundleDisplayName</key>
+	<string>Samples</string>
```
philium: The current bundle display name (on the App Store) is "iOS Samples". I'm fine with using just "Samples", although it  rather generic. @dg0yal Do you have a preference?

dg0yal: I am fine with Samples

PR #526, opened 2018-10-17
Updates the RGB Renderer and Stretch Renderer samples

qsmmorgan:
This PR updates both samples to use a popover to display their settings. Both samples are included in one PR since they shared a common `enum` which would cause merge conflicts if edited separately.


qsmmorgan:
I agree with your points. Everything except the vertical constraints was pre-existing, so I'm not going to worry about it now if that's okay.

PR #536, opened 2018-10-23
Makes OptionsTableViewController generic

philium:
@qsmmorgan Just some ideas I had. Take them or leave them.


philium:
You're welcome. I'm glad some of it was helpful.

> we shouldn't really be extending Runtime SDK objects anyway

I'm not sure that extending `ArcGIS` types, in and of itself, is a bad thing. There are several samples that do just that.

qsmmorgan:
> I'm not sure that extending ArcGIS types, in and of itself, is a bad thing. There are several samples that do just that.

@philium I was wondering if there were. I don't think it's _bad_ per se, but in the interest of clearly demonstrating the pure SDK, I'm trying to avoid it.

PR #540, opened 2018-10-24
Replaces use of HorizontalPicker in Blend Renderer

qsmmorgan:


arcgis-ios-sdk-samples/Layers/Blend renderer/BlendRendererSettingsVC.swift
```diff
@@ -60,104 +62,108 @@ class BlendRendererSettingsVC: UITableViewController {
...
-        switch slopeType {
-        case .none:
-            slopeTypePicker?.selectedIndex = 0
-        case .degree:
-            slopeTypePicker?.selectedIndex = 1
-        case .percentRise:
-            slopeTypePicker?.selectedIndex = 2
-        case .scaled:
-            slopeTypePicker?.selectedIndex = 3
+    private var slopeTypeID: Int {
```
philium: What would you think of just offsetting the slope type's raw value?
```swift
private var slopeTypeID: Int {
        return slopeType.rawValue + 1
    }
```
And where it was being set:
```swift
self.slopeType = AGSSlopeType(rawValue: newIndex - 1)!
```
Same with color ramp type.

PR #541, opened 2018-10-24
Replaces HorizontalPicker in RGB Renderer and Stretch Renderer

qsmmorgan:


qsmmorgan:
@philium Changed some things based on your feedback, ready for re-review.

arcgis-ios-sdk-samples/Layers/RGB renderer/RGB Renderer Cells.swift
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+import UIKit
+
+class RGBRendererInputCell: UITableViewCell {
+    
+    @IBOutlet weak var leadingLabel: UILabel!
+    @IBOutlet weak var textField: UITextField!
+    
+    let numberFormatter = NumberFormatter()
```
philium: There is no reason to give each cell its own number formatter. Formatters are expensive to create, so creating one per cell makes loading the table view necessarily expensive. In addition, it is the data source, not the cell itself, that should be responsible for configuring the cell's content. 

Please update the cell in the table view's data source instead. It is fine to have convenience methods like this on the data source to configure different kinds of cells. And by doing so, the data source can reuse a single number formatter for all cells.

qsmmorgan: I disagree that this is a problem but I'll change it.

PR #544, opened 2018-10-25
Makes PlaneStatsViewController more useful

philium:
Also uses measurement formatter for display of altitude, heading, pitch, and roll.

I realize you aren't responsible for `PlaneStatsViewController` being so simple, but it is nice to make it pull its own weight. And using measurement formatter for displaying value+symbol is a better pattern to model.

### DIFF
--- a/arcgis-ios-sdk-samples/Scenes/Animate 3D graphic/Animate3DGraphicViewController.swift
+++ b/arcgis-ios-sdk-samples/Scenes/Animate 3D graphic/Animate3DGraphicViewController.swift
@@ -37,8 +37,6 @@ class Animate3DGraphicViewController: UIViewController {
     private weak var planeStatsViewController: PlaneStatsViewController?
     private weak var missionSettingsViewController: MissionSettingsViewController?
     
-    private let numberFormatter = NumberFormatter()
-    
     private var isAnimating = false {
         didSet {
             playBBI?.title = isAnimating ? "Pause" : "Play"
@@ -186,6 +184,7 @@ class Animate3DGraphicViewController: UIViewController {
                 //create a frame object for each line
                 frames = lines.map { (line) -> Frame in
                     let details = line.components(separatedBy: ",")
+                    precondition(details.count == 6)
                     
                     let position = AGSPoint(x: Double(details[0])!,
                                             y: Double(details[1])!,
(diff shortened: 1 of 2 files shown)


arcgis-ios-sdk-samples/Scenes/Animate 3D graphic/Animate3DGraphicViewController.swift
```diff
@@ -388,16 +383,20 @@ extension Animate3DGraphicViewController: UIAdaptivePresentationControllerDelega
     }
 }
 
-private class Frame {
+struct Frame {
```
philium: Making this a struct eliminates the overhead of memory management for these object, which is a nice and free improvement. The type could probably do with a better name, but I had to stop experimenting somewhere.

arcgis-ios-sdk-samples/Scenes/Animate 3D graphic/Animate3DGraphicViewController.swift
```diff
@@ -186,6 +184,7 @@ class Animate3DGraphicViewController: UIViewController {
                 //create a frame object for each line
                 frames = lines.map { (line) -> Frame in
                     let details = line.components(separatedBy: ",")
+                    precondition(details.count == 6)
```
philium: Just a little sanity check. Ideally we'd convert the file to something more easily parsable, like json. Then we could model it with a decodable type.

arcgis-ios-sdk-samples/Scenes/Animate 3D graphic/PlaneStatsViewController.swift
```diff
@@ -16,10 +16,57 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        updateUI()
+    }
+    
+    private var tableViewContentSizeObservation: NSKeyValueObservation?
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        tableViewContentSizeObservation = tableView.observe(\.contentSize) { [unowned self] (tableView, _) in
```
philium: I like seeing popovers sized to fit the enclosing table view just right. This is the best and easiest way I know how to do that. I suppose the proper height could be hard coded, but that only works when not using dynamic type.

PR #547, opened 2018-10-29
Prevent `SVProgressHUD` deadlocking

qsmmorgan:
Adds or corrects placement of `dismiss()` calls to match `class show(withStatus:)` of `SVProgressHUD` and prevent deadlocking.


arcgis-ios-sdk-samples/Edit data/Offline edit and sync/OfflineEditingViewController.swift
```diff
@@ -408,14 +420,15 @@ class OfflineEditingViewController: UIViewController, AGSGeoViewTouchDelegate, A
             
             SVProgressHUD.show(withStatus: status.statusString())
             
-        }, completion: { (results: [AGSSyncLayerResult]?, error: Error?) -> Void in
+        }, completion: {[weak self] (results: [AGSSyncLayerResult]?, error: Error?) -> Void in
```
philium: Thank you for fixing the incorrect capture of `self`.

PR #549, opened 2018-10-29
Adds the Show Labels on Layers sample

qsmmorgan:
This sample is nearly identical to the macOS implementation, so I don't expect the review to be too taxing.


arcgis-ios-sdk-samples/Layers/Show labels on layers/ShowLabelsOnLayersViewController.swift
```diff
@@ -0,0 +1,99 @@
...
+        // https://developers.arcgis.com/web-map-specification/objects/labelingInfo/
+        
+        /// The styling for the label.
+        let textSymbol = AGSTextSymbol()
+        textSymbol.size = 12
+        textSymbol.haloColor = .white
+        textSymbol.haloWidth = 2
+        textSymbol.color = color
+        // the object must be converted to a JSON object
+        let textSymbolJSON = try! textSymbol.toJSON()
```
pgruenler: This will crash if `toJSON` fails (not that I expect it to); I think it would be good avoid crashing.  Maybe use a `try?` instead and then handle the optional.

arcgis-ios-sdk-samples/Layers/Show labels on layers/ShowLabelsOnLayersViewController.swift
```diff
@@ -0,0 +1,99 @@
...
+            "symbol": textSymbolJSON
+        ]
+        
+        do {
+            // create and return a label definition from the JSON object
+            return try AGSLabelDefinition.fromJSON(labelJSONObject) as? AGSLabelDefinition
+        }
+        catch {
+            presentAlert(error: error)
+            return nil
```
pgruenler: Returning `nil` here will cause a crash where the function is called and the returned value is force unwrapped.  Here again I think avoiding a crash would be good.

qsmmorgan:
@pgruenler I fixed the unsafe unwrapping.

PR #551, opened 2018-10-30
Replaces instances of HorizontalColorPicker and removes its definition

qsmmorgan:


arcgis-ios-sdk-samples/Analysis/Viewshed (location)/ViewshedLocation.storyboard
```diff
@@ -1,42 +1,58 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
philium: It seems like there would be a lot of value in making a reusable cell subclass for each type of cell:

- The cell would only need to be designed once.
- It would ensure a consistent look for each instance.
- The view controller would only need a single outlet for each cell.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/Color Picker/ColorPickerViewController.swift
```diff
@@ -69,6 +73,8 @@ class ColorPickerViewController: UITableViewController {
             saturation = value
         case brightnessSlider:
             brightness = value
+        case alphaSlider:
```
philium: The sliders are continue which means that these comparisons are being made many, many times. Could you make a separate action method for each slider to prevent doing a ton of unnecessary comparisons?

arcgis-ios-sdk-samples/Content Display Logic/Controllers/Color Picker/ColorPickerViewController.swift
```diff
@@ -54,10 +56,12 @@ class ColorPickerViewController: UITableViewController {
         hueSlider.value = Float(hue)
         saturationSlider.value = Float(saturation)
         brightnessSlider.value = Float(brightness)
+        alphaSlider.value = Float(alpha)
```
philium: It doesn't make a whole lot of sense to update every slider and label when a single component changes. Could you break this method up into a separate method per component?

arcgis-ios-sdk-samples/Analysis/Viewshed (location)/ViewshedLocationViewController.swift
```diff
@@ -21,8 +21,7 @@ class ViewshedLocationViewController: UIViewController {
     @IBOutlet weak var setObserverOnTapInstruction: UILabel!
     @IBOutlet weak var updateObserverOnDragInstruction: UILabel!
     
-    private var viewshed: AGSLocationViewshed!
-    private var analysisOverlay: AGSAnalysisOverlay!
+    private weak var viewshed: AGSLocationViewshed?
```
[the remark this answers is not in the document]
qsmmorgan: I think it's better to have it `weak` to show developers that the runtime retains a reference internally.

arcgis-ios-sdk-samples/Analysis/Viewshed (location)/ViewshedSettingsVC.swift
```diff
@@ -13,112 +13,174 @@
...
-class ViewshedSettingsVC: UITableViewController, HorizontalColorPickerDelegate {
+    weak var viewshed: AGSLocationViewshed? {
+        didSet {
+            updateUIForViewshed()
+        }
+    }
     
-    @IBOutlet weak var headingSlider: UISlider!
-    @IBOutlet weak var headingLabel: UILabel!
+    @IBOutlet weak var headingSlider: UISlider?
```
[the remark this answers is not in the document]
qsmmorgan: But it's safer to avoid IUO. That way the compiler enforces crash-avoidance so the developer doesn't have to remember to check `isViewLoaded`.

arcgis-ios-sdk-samples/Analysis/Viewshed (location)/ViewshedSettingsVC.swift
```diff
@@ -13,112 +13,174 @@
...
-    func viewshedSettingsVC(_ viewshedSettingsVC:ViewshedSettingsVC, didUpdateHorizontalAngle horizontalAngle:Double)
-    func viewshedSettingsVC(_ viewshedSettingsVC:ViewshedSettingsVC, didUpdateVerticalAngle verticalAngle:Double)
-    func viewshedSettingsVC(_ viewshedSettingsVC:ViewshedSettingsVC, didUpdateMinDistance minDistance:Double)
-    func viewshedSettingsVC(_ viewshedSettingsVC:ViewshedSettingsVC, didUpdateMaxDistance maxDistance:Double)
+class ViewshedSettingsVC: UITableViewController {
     
-}
-
-class ViewshedSettingsVC: UITableViewController, HorizontalColorPickerDelegate {
+    weak var viewshed: AGSLocationViewshed? {
```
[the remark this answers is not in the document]
qsmmorgan: I agree it doesn't need to be `weak`, but it also doesn't need to be strong. If all other references to `viewshed` are released, there's no point in keeping it around here nor updating it.

arcgis-ios-sdk-samples/Analysis/Viewshed (location)/ViewshedLocation.storyboard
qsmmorgan: In previous samples you recommended using static cells for settings. Maybe prototype cells make more sense here since there is lots of repetition, but shouldn't we be consistent? In any case, I'm not going to change it for this sample since it's already hooked up manually.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/Color Picker/ColorPickerViewController.swift
qsmmorgan: Sure, that's a good idea.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/Color Picker/ColorPickerViewController.swift
qsmmorgan: Yeah, you're right. As you can guess, this was faster to write 😆

arcgis-ios-sdk-samples/Display information/Display grid/DisplayGridSettingsViewController.swift
```diff
@@ -151,42 +172,66 @@ class DisplayGridSettingsViewController: UITableViewController {
         grid.labelVisibility = priorGrid.labelVisibility
         grid.isVisible = priorGrid.isVisible
         
-        if let selectedColor = gridColorPicker.selectedColor {
-            changeGridColor(selectedColor)
+        if let priorGridColor = gridColor(of: priorGrid) {
```
pgruenler: I'm confused by the use of `prior` in the variable name when that seems to be what is going to be the new color.  Or maybe I have that the wrong way around.  I think with the changes you've made here `prior` no longer applies.

arcgis-ios-sdk-samples/Analysis/Viewshed (location)/ViewshedSettingsVC.swift
```diff
@@ -13,112 +13,174 @@
...
+        updateUIForViewshed()
+    }
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        // set the colors redundantly to avoid a visual glitch when closing the color picker
+        updateColorUIForViewshed()
+    }
+    
+    private func updateUIForViewshed() {
+        guard let viewshed = viewshed else {
```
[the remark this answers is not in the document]
qsmmorgan: I put this as a check in `didSet` of `viewshed` so `updateUIForViewshed()` won't be called if the view isn't loaded.

arcgis-ios-sdk-samples/Analysis/Viewshed (location)/ViewshedLocation.storyboard
philium: You can use a `UITableViewCell` subclass with static cells.

qsmmorgan: Ohhhh now I see what you're saying. Interface builder is still the land of many secrets to me. I'll look into it.

philium: You should be able to make the cells designable, but designables seem to not compile successfully. I suspect it has to do with how we specify the path the `ArcGIS`. Now that Xcode 10.1 is out of beta, I'll submit a PR that updates the path to the framework and uses the new build system. That may fix the problem.

PR #553, opened 2018-10-31
Use new build system

philium:
Unfortunately it was necessary to revert to using `$HOME` in the framework's path. This means that the framework will once again show as missing (red) in Xcode, even though the project builds successfully.

Note: Once this PR is merged, Xcode 10.1 will be required to build the app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,8 @@ The ```master``` branch of this repository contains samples configured for the l
 ## Requirements
 
 * [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/ios/) 100.4.0 (or higher).
-* Xcode 10.0 (or higher)
-* iOS 12 SDK (or higher)
+* Xcode 10.1 (or higher)
+* iOS 12.1 SDK (or higher)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -23,8 +23,8 @@ The ```master``` branch of this repository contains samples configured for the l
 ## Requirements
 
 * [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/ios/) 100.4.0 (or higher).
-* Xcode 10.0 (or higher)
-* iOS 12 SDK (or higher)
+* Xcode 10.1 (or higher)
+* iOS 12.1 SDK (or higher)
```
qsmmorgan: This isn't related to this PR, but we should consider removing the minimum version of the iOS SDK needed to build the app and replacing it with the minimum version of iOS required to run the app. The min Xcode version implies the SDK version anyway.

qsmmorgan:
Oh, so Xcode added back a rather empty `WorkspaceSettings.xcsettings` file. Might want to push that. Not sure if it was when I opened the project or opened the Project Settings in Xcode.

philium:
I think Xcode adds it when opening the Project Settings.

PR #554, opened 2018-11-01
Converts the Manage Operational Layers settings to a popover

qsmmorgan:


arcgis-ios-sdk-samples/Maps/Manage operational layers/MMLLayersViewController.swift
```diff
@@ -15,77 +15,82 @@
...
+    private func dataSourceIndexForIndexPath(_ dataSource: NSMutableArray, indexPath: IndexPath) -> Int {
         return dataSource.count - indexPath.row - 1
     }
     
-    enum Section: Int {
+    private enum Section: Int {
         case added
         case removed
-        
-        init(_ rawValue: Int) {
```
philium: You prefer to force unwrap at each call site rather than in this initializer? This seemed like a pretty reasonable convenience to have. In all cases the table view is providing the section, providing the security that the passed in section will be a valid raw value.

qsmmorgan: I guess I don't care either way. I changed it because I'd never seen this pattern before and it seemed nominally less safe.

arcgis-ios-sdk-samples/Maps/Manage operational layers/MMLLayersViewController.swift
```diff
@@ -15,77 +15,82 @@
...
 class MMLLayersViewController: UITableViewController {
-    var layers: NSMutableArray!
-    var deletedLayers: [AGSLayer]!
     
-    weak var delegate: MMLLayersViewControllerDelegate?
+    /// Every layer on the map or that could be added to the map.
+    var allLayers: [AGSLayer] = []
+    
+    /// The layers attached to the map.
+    var operationalLayers: NSMutableArray?
```
pgruenler: Can you make this not optional?  It would get rid of a lot of forced unwrapping.

qsmmorgan: Made the changes in the latest commit.

qsmmorgan:
@philium I fixed the bug on this you sent in slack.

arcgis-ios-sdk-samples/Maps/Manage operational layers/MMLLayersViewController.swift
philium: Taking a look at all the places where you had to add `rawValue: ` and `!`, I don't think removing this was an improvement.

arcgis-ios-sdk-samples/Maps/Manage operational layers/MMLLayersViewController.swift
```diff
@@ -15,150 +15,149 @@
...
-    enum Section: Int {
-        case added
-        case removed
+    /// A convenience type for the table view sections.
+    private enum Section: Int {
+        case operational, removed
         
-        init(_ rawValue: Int) {
-            self.init(rawValue: rawValue)!
+        var label: String {
```
philium: This is only used in `titleForHeaderInSection`, so I think it makes more sense to leave the switch statement where it was.

arcgis-ios-sdk-samples/Maps/Manage operational layers/ManageMapLayersViewController.swift
```diff
@@ -15,51 +15,73 @@
...
+            let controller = navController.viewControllers.first as? MMLLayersViewController {
+            // convert and assign the map's operational layers as a Swift array
+            controller.operationalLayers = mapView.map?.operationalLayers as? [AGSLayer] ?? []
+            controller.removedLayers = removedLayers
+            controller.onChange = {[weak self] layers in
+                guard let map = self?.mapView.map else {
+                    return
+                }
+                // replace the map's operational layers with those from the view controller
+                map.operationalLayers.removeAllObjects()
```
philium: Dealing with an optional `NSMutableArray` is not great, but I'm not sure this is any better. Removing all layers and re-adding them on each change does not seem like a good idea. Wouldn't that cause all layers to be redrawn, not just the one that was added or removed?

arcgis-ios-sdk-samples/Maps/Manage operational layers/ManageMapLayersViewController.swift
```diff
@@ -15,51 +15,88 @@
...
-            controller.layers = self.map.operationalLayers
-            controller.deletedLayers = self.deletedLayers
-            controller.preferredContentSize = CGSize(width: 300, height: 300)
-            controller.delegate = self
+        if let navController = segue.destination as? UINavigationController,
+            let controller = navController.viewControllers.first as? MMLLayersViewController {
+            // convert and assign the map's operational layers as a Swift array
+            controller.operationalLayers = mapView.map?.operationalLayers as? [AGSLayer] ?? []
+            controller.removedLayers = removedLayers
+            controller.onChange = {[weak self] layers in
```
philium: Only one change can happen at a time, yes? Trying to handle all of them at the same time is an interesting idea, but it can be error prone and hard to understand. I had a hard time following what was going on. What would you think of passing the type of change, either as a parameter or by using separate callbacks per change type? I'm thinking something along the lines of `NSFetchedResultsControllerDelegate.controller(_:didChange:at:for:newIndexPath:)`. You could incorporate the index information into the change type itself:

```swift
enum ChangeType {
    case insert(index: Int)
    case delete(index: Int)
    case move(oldIndex: Int, newIndex: Int)
}
```

I'm really not trying to cause you extra work for no reason. The whole point of this sample is to show users "how to add, remove or reorder operational layers in a map". We want that to be as clear as possible. The way it is now, I don't think it is very clear.

arcgis-ios-sdk-samples/Maps/Manage operational layers/MMLLayersViewController.swift
philium: I understand experimenting. You may have thought that the title would be used from more than one place. But in the end, it wasn't. And it isn't likely to be used anywhere else. So leaving that change in your branch left me scratching my head. Why go through the extra work of moving the switch statement, wrapping it in a property, and then calling that property from the original location? It didn't make sense.

When you're proofreading your changes before submitting a PR, try to keep an eye out for unnecessary changes like this and removing the custom initializer. Taking the time to revert those unnecessary changes before review reduces the amount of work the reviewer has to do and avoids any potential confusion. Does that make sense?

arcgis-ios-sdk-samples/Maps/Manage operational layers/MMLLayersViewController.swift
```diff
@@ -15,150 +15,149 @@
...
 
-protocol MMLLayersViewControllerDelegate: AnyObject {
-    func layersViewControllerWantsToClose(_ layersViewController: MMLLayersViewController, withDeletedLayers layers: [AGSLayer])
-}
-
 class MMLLayersViewController: UITableViewController {
-    var layers: NSMutableArray!
-    var deletedLayers: [AGSLayer]!
     
-    weak var delegate: MMLLayersViewControllerDelegate?
```
philium: I can understand (a little bit) not wanting to take the time to implement a delegate. What I don't understand is removing a delegate that already exists and replacing it with a closure property.

qsmmorgan: It's never been about not wanting to take the time. I've told you before that I don't like delegates when there are only one or two callbacks. I changed it to a closure because I think it is the better solution.

arcgis-ios-sdk-samples/Maps/Manage operational layers/ManageMapLayersViewController.swift
qsmmorgan: This is why before I just removed all the layers every time.

qsmmorgan: That said, I don't love either of the implementations. I might revert some of them.

arcgis-ios-sdk-samples/Maps/Manage operational layers/MMLLayersViewController.swift
qsmmorgan: It seems like we've been repeating discussions over the same coding differences. I think a lot of our disagreements could be avoided if Esri (or even just Runtime) had comprehensive Swift guidelines. Both of our methods work, so lacking a standard I don't feel compelled to code a particular way.

qsmmorgan: In the case of this sample, I pushed new changes that removes the need for a callback at all.

arcgis-ios-sdk-samples/Maps/Manage operational layers/MMLLayersViewController.swift
```diff
@@ -15,150 +15,147 @@
...
+        return sourceIndexPath
+    }
+    
+    override func tableView(_ tableView: UITableView, moveRowAt sourceIndexPath: IndexPath, to destinationIndexPath: IndexPath) {
+        // update the order of layers in the array
         
-        layers.removeObject(at: sourceIndex)
-        layers.insert(layer, at:destinationIndex)
+        if destinationIndexPath != sourceIndexPath {
+            map?.operationalLayers.exchangeObject(at: sourceIndexPath.row, withObjectAt: destinationIndexPath.row)
```
philium: This line causes the following to be printed in the debugger:

> ArcGIS Runtime Error Occurred. Set a breakpoint on C++ exceptions to see the original callstack and context for this error:  Error Domain=com.esri.arcgis.runtime.error Code=24 "Object already owned." UserInfo={NSLocalizedFailureReason=Already owned., NSLocalizedDescription=Object already owned., Additional Message=Already owned.}

Have you looked into that? It doesn't look like you're doing anything wrong, but we might need to file a runtime issue for this.

qsmmorgan: Nice catch. I don't think my code does anything special to cause that. I think you're right about filing a runtime issue, seems like it's an error with `exchangeObject(at:, withObjectAt:)` internal to the SDK.

qsmmorgan:
@pgruenler Can you re-review when you get a chance?

arcgis-ios-sdk-samples/Maps/Manage operational layers/MMLLayersViewController.swift
pgruenler: Yes, please make an issue in cocoa-common-api.

PR #555, opened 2018-11-01
Converts Open Map URL settings to a popover

qsmmorgan:


philium:
I don't think I'm in favor of some of these changes. It is unusual for the settings view controller to initialize its own data model. And because it is private, the `OpenMapURLViewController` has to hardcode the initial URL. I think it would make more sense for `MapAtURL` to be a model shared between the two view controllers and therefore not a subtype. I also would like to see the sample view controller maintain the list of maps and pass it to the settings view controller. That is how we do this sort of thing in most of the other samples.

qsmmorgan:
@philium Okay, I'm fine with that. I was ambivalent about where to put things after essentially splitting one VC into two.

qsmmorgan:
@philium I made changes based on your feedback. Let me know if you had something more in mind.

philium:
Thank you!

PR #556, opened 2018-11-01
Updates Offline Map Edit and Sync to allow syncing without edits

qsmmorgan:
Also organizes and cleans up the sample's code somewhat, removing IUO for runtime objects.


qsmmorgan:
@philium Thanks for the note. Sometimes I get carried away haha but I can see why separating the tasks would be helpful when reviewing.

PR #558, opened 2018-11-02
Messages are now shown in a popover in Display KML Network Links

qsmmorgan:


arcgis-ios-sdk-samples/Layers/Display KML network links/DisplayKMLNetworkLinksViewController.swift
```diff
@@ -57,4 +62,34 @@ class DisplayKMLNetworkLinksViewController: UIViewController {
         scene.operationalLayers.add(kmlLayer)
     }
     
+    private func updatePopoverForMessage() {
```
philium: This should be the message view controller's responsibility.

arcgis-ios-sdk-samples/Layers/Display KML network links/DisplayKMLNetworkLinksViewController.swift
```diff
@@ -57,4 +62,34 @@ class DisplayKMLNetworkLinksViewController: UIViewController {
...
+
+extension DisplayKMLNetworkLinksViewController: UIAdaptivePresentationControllerDelegate {
+    
+    func adaptivePresentationStyle(for controller: UIPresentationController, traitCollection: UITraitCollection) -> UIModalPresentationStyle {
+        // show as this as a popover even on small displays
+        return .none
+    }
+}
+
+class KMLNetworkMessagesViewController: UIViewController {
```
philium: A view controller without any custom logic is just a view. A view controller mediates between a model and a view. What's the model for this view controller? It should be the message. The consumer of `KMLNetworkMessagesViewController` should be able to assign a message to the view controller, whether its view has loaded or not, and the view controller itself should update its own view appropriately.

PR #560, opened 2018-11-05
Converts the Display Device Location settings to a popover

qsmmorgan:
Supersedes issues where the sample's controls would overflow out of the safe area.

Updates the README to account for the changes.


arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
```diff
@@ -19,11 +19,13 @@ import UIKit
 class OptionsTableViewController: UITableViewController {
     
     private let labels: [String]
+    private let imageNames: [String]
     private var selectedIndex: Int
     private let onChange: (Int) -> Void
     
-    init(labels: [String], selectedIndex: Int, onChange: @escaping (Int) -> Void) {
+    init(labels: [String], imageNames: [String] = [], selectedIndex: Int, onChange: @escaping (Int) -> Void) {
```
philium: One of the issues I saw with `CustomContextSheet` was that it took a separate array for the titles, images, and highlight images. Instead, it should have taken an array of a type with a property for the title, the image, and the highlight image.

Could you please combine the title and image into a single type? That could either be a protocol or a concrete type:

```swift
struct Option {
    var title: String
    var image: UIImage?
}
```

Using `UIImage` for `image` makes sense because some options may already have `UIImage` objects that didn't come from named resources.

arcgis-ios-sdk-samples/Maps/Display device location/DisplayLocationSettingsViewController.swift
```diff
@@ -0,0 +1,128 @@
...
+    
+    //MARK: - Outlets
+    
+    @IBOutlet weak var showSwitch: UISwitch!
+    @IBOutlet weak var autoPanModeCell: UITableViewCell!
+    
+    //MARK: - Model
+    
+    /// The SDK object for displaying the device location. Attached to the map view.
+    weak var locationDisplay: AGSLocationDisplay?
```
philium: What happens if the location display is set after the view is loaded?

I realize that probably isn't happening now, but it could happen in the future if the sample were updated. Since a view controller has no control over when its own view will be loaded, it is a good idea (and a good habit to form) for any view controller to handle the model being set both before and after the view is loaded.

arcgis-ios-sdk-samples/Maps/Display device location/DisplayLocationSettingsViewController.swift
```diff
@@ -0,0 +1,128 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+class DisplayLocationSettingsViewController: UITableViewController {
```
philium: This settings view is so simple, it would have been nice to keep everything in one view. You could have used a picker for the auto pan mode and done away with the navigation controller.

I suppose it's not worth changing since you've already implemented it, but something to think about in the future.

qsmmorgan: I'm trying to be consistent with the controls.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
qsmmorgan: I mean this is theoretically better but it puts an unnecessary burden on samples to wrap the titles in an `Option` type when there isn't an image. If you're set on it then I'd make another initializer.

arcgis-ios-sdk-samples/Maps/Display device location/DisplayLocationSettingsViewController.swift
qsmmorgan: This pattern has been repeated in many samples now, and while I don't mind the safety, I'm less inclined to add code that will never be needed. It the sample is updated in the future, the developer can make the change then.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
philium: > it puts an unnecessary burden on samples to wrap the titles in an Option type

Mapping to another type isn't a burden. Consumers already have to map over their underlying type to get an array of titles, so updating the closure to return an `Option` rather than a `String` is rather trivial.

Another option would be to make `Option` a protocol:

```swift
protocol OptionProtocol {
    var title: String { get }
    var image: UIImage? { get }
}

extension OptionProtocol {
    // Convenience for options that do not have a corresponding image.
    var image: UIImage? { return nil }
}
```

And then make `OptionsTableViewController` generic. The advantage is that the underlying type can conform to `Option`, eliminating the need to do any wrapping/mapping. I see that as the best "option", pun intended.

qsmmorgan: Some of the consumers don't have an underlying type, they just pass in an array of titles directly. Even if they all did, using a protocol would mean extending the SDK and I'm against that. Both options make the sample unnecessarily more complicated.

As far as passing in images rather than image names, that would be a good option if there were use case that necessitated it. For now it's cleaner to let the options controller handle image initialization.

arcgis-ios-sdk-samples/Maps/Display device location/DisplayLocationSettingsViewController.swift
qsmmorgan: I made the change.

qsmmorgan:
@philium It'd be nice to get this one merged. Let me know if there's anything left you really think needs changing.

philium:
I still would like to see `OptionsTableViewController` not take separate arrays for the titles and the images. It reminds me of a C API, akin to [NSDictionary.init(objects:forKeys:count)](https://developer.apple.com/documentation/foundation/nsdictionary/1412631-init). I think making the view controller generic is probably the best way to do it, but there are other ways. You could make an `Option` type with `title` and `image` properties and have the initializer take a sequence of options.

qsmmorgan:
@philium Yeah you're right that would be nice. I'll whip something up.

qsmmorgan:
@philium Done.

qsmmorgan:
@philium Please re-approve due to merge conflicts.

PR #561, opened 2018-11-07
Updates the Change Basemap settings to use a popover

qsmmorgan:
Also adds more of Esri's default basemap types as options


arcgis-ios-sdk-samples/Maps/Change basemap/SwitchBasemapViewController.swift
```diff
@@ -15,93 +15,86 @@
...
     @IBOutlet private weak var mapView:AGSMapView!
-    @IBOutlet weak var blurView: UIVisualEffectView!
-    @IBOutlet weak var tableParentView: UIView!
-    @IBOutlet weak var tableView: UITableView!
     
-    private var titles = ["Streets (Raster)", "Streets (Vector)", "Streets - Night (Vector)", "Imagery (Raster)", "Imagery with Labels (Raster)", "Imagery with Labels (Vector)", "Dark Gray Canvas (Vector)", "Light Gray Canvas (Raster)", "Light Gray Canvas (Vector)", "Navigation (Vector)", "OpenStreetMap (Raster)"]
-    
-    var map:AGSMap!
+    /// The basemap options plus labels.
+    private let basemapInfoArray: [(basemap: AGSBasemap, label: String)] = [
```
philium: I just thought of something. The only disadvantage I can see to this approach is creating all of the base maps up front. I don't know what the ramifications are, particularly for memory usage. Have you done any research into that?

qsmmorgan: That's a good thought! In my quick testing, it looks like opening the sample from first launch raises memory usage to around 220 MB. Opening the Display A Map sample (which instantiates just one basemap) uses about 190 MB. Switching between the basemaps doesn't appear to leak memory either. Overall, it would be a small savings to create the basemaps ad hoc, but I think for the simplicity of the sample it's the right tradeoff to leave it as-is.

PR #562, opened 2018-11-07
Removes usage of NSSearchPathForDirectoriesInDomains

philium:
Its usage is discouraged in the documentation.

### DIFF
--- a/arcgis-ios-sdk-samples/Edit data/Offline edit and sync/OfflineEditingViewController.swift
+++ b/arcgis-ios-sdk-samples/Edit data/Offline edit and sync/OfflineEditingViewController.swift
@@ -187,14 +187,14 @@ class OfflineEditingViewController: UIViewController {
     
     private func deleteAllGeodatabases() {
         //Remove all files with .geodatabase, .geodatabase-shm and .geodatabase-wal file extensions
-        let path = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0]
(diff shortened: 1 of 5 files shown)


arcgis-ios-sdk-samples/Edit data/Offline edit and sync/OfflineEditingViewController.swift
```diff
@@ -187,14 +187,14 @@ class OfflineEditingViewController: UIViewController {
     
     private func deleteAllGeodatabases() {
         //Remove all files with .geodatabase, .geodatabase-shm and .geodatabase-wal file extensions
-        let path = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0]
+        let documentDirectoryURL = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
```
pgruenler: Is it safe to force-unwrap `first`?  I'm guessing the array will always contain at least one thing but I just wanted to double check.

pgruenler: @nixta, I agree a `guard` would be the safest but because it's the documents directory I think I'm okay with the force-unwrap.

nixta: @pgruenler I agree. It keeps nagging at me that a lot of the readers of the sample app code will either be relatively new to iOS and/or new to Swift so while _we_ know this is safe here, I personally tend to lean away from it. But yeah, it's good as is.

PR #576, opened 2018-11-13
Enables the empty_count SwiftLint rule

qsmmorgan:
Replaces most instances of `array[0]` with `.first`

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Create and save a map/CreateOptionsViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Create and save a map/CreateOptionsViewController.swift
@@ -143,8 +143,11 @@ class CreateOptionsViewController: UIViewController, UITableViewDataSource, UITa
         let basemap = self.basemaps[self.selectedBasemapIndex].copy() as! AGSBasemap
         
         //create an array of the selected operational layers
-        let selectedLayers: [AGSLayer]? = selectedLayersIndex.isEmpty ? nil : selectedLayersIndex.map { (index) -> AGSLayer in
-            return layers[index].copy() as! AGSLayer
+        let selectedLayers: [AGSLayer]?
+        if !selectedLayersIndex.isEmpty {
+            selectedLayers = selectedLayersIndex.map { layers[$0].copy() as! AGSLayer }
+        } else {
+            selectedLayers = nil
         }
         
         self.delegate?.createOptionsViewController(self, didSelectBasemap: basemap, layers: selectedLayers)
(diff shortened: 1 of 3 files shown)


arcgis-ios-sdk-samples/Maps/Create and save a map/CreateOptionsViewController.swift
```diff
@@ -143,13 +143,11 @@ class CreateOptionsViewController: UIViewController, UITableViewDataSource, UITa
         let basemap = self.basemaps[self.selectedBasemapIndex].copy() as! AGSBasemap
         
         //create an array of the selected operational layers
-        var layers = [AGSLayer]()
-        for index in self.selectedLayersIndex {
-            let layer = self.layers[index].copy() as! AGSLayer
-            layers.append(layer)
+        let selectedLayers: [AGSLayer]? = selectedLayersIndex.isEmpty ? nil : selectedLayersIndex.map { (index) -> AGSLayer in
```
philium: This, because of the ternary operator, is a little hard to follow. Maybe something like this would be more readable?

```swift
let selectedLayers: [AGSLayer]?
if !selectedLayersIndex.isEmpty {
    selectedLayers = selectedLayersIndex.map { layers[$0].copy() as! AGSLayer }
} else {
    selectedLayers = nil
}
```

arcgis-ios-sdk-samples/Search/Offline geocode/OfflineGeocodeViewController.swift
```diff
@@ -148,7 +153,8 @@ class GeocodeOfflineViewController: UIViewController, AGSGeoViewTouchDelegate, U
                 //if a result is found extract the required attributes
                 //assign the attributes to the graphic
                 //and show the callout
-                if let results = results, results.count > 0 {
+                if let results = results,
```
philium: Should this be `if let result = results.first { … }`?

arcgis-ios-sdk-samples/Search/Reverse geocode/ReverseGeocodeViewController.swift
```diff
@@ -79,7 +79,8 @@ class ReverseGeocodeViewController: UIViewController, AGSGeoViewTouchDelegate {
                 }
             }
             else {
-                if let results = results, results.count > 0 {
+                if let results = results,
```
philium: It seems like this should be `if let result = results.first { … }`.

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateOptionsViewController.swift
qsmmorgan: Looks good, thanks

PR #579, opened 2018-11-14
Enables the notification_center_detachment SwiftLint rule

qsmmorgan:


arcgis-ios-sdk-samples/Display information/Sketch on the map/SketchViewController.swift
```diff
@@ -48,6 +48,11 @@ class SketchViewController: UIViewController {
         self.map.initialViewpoint = AGSViewpoint(targetExtent: AGSEnvelope(xMin: -10049589.670344, yMin: 3480099.843772, xMax: -10010071.251113, yMax: 3512023.489701, spatialReference: .webMercator()))
     }
     
+    deinit {
+        //remove self as observer for notifications
+        NotificationCenter.default.removeObserver(self, name: .AGSSketchEditorGeometryDidChange, object: nil)
```
philium: Was this change due to a rule violation? If so, I would not be in favor of the rule. This (removing `self` as an observer in `deinit`) has not been necessary since like iOS 9.

arcgis-ios-sdk-samples/Edit data/Edit features (connected)/EditFeaturesOnlineViewController.swift
```diff
@@ -139,7 +146,6 @@ class EditFeaturesOnlineViewController: UIViewController, AGSGeoViewTouchDelegat
         //disable the done button until any geometry changes
         self.doneBBI.isEnabled = false
         
-        NotificationCenter.default.addObserver(self, selector: #selector(EditFeaturesOnlineViewController.sketchChanged(_:)), name: .AGSSketchEditorGeometryDidChange, object: nil)
```
philium: It isn't immediately clear why this was where it was and what the consequences of moving it would be. I think a better option would be to leave this where it was (since moving it wasn't needed to satisfy the rule) and then update the removal to specify the the notification name, namely `AGSSketchEditorGeometryDidChange`.

Same with `OfflineEditingViewController.swift`.

arcgis-ios-sdk-samples/Edit data/Edit features (connected)/EditFeaturesOnlineViewController.swift
```diff
@@ -50,6 +50,13 @@ class EditFeaturesOnlineViewController: UIViewController, AGSGeoViewTouchDelegat
                 
         //hide the sketchToolbar initially
         self.sketchToolbar.isHidden = true
+        
+        NotificationCenter.default.addObserver(self, selector: #selector(EditFeaturesOnlineViewController.sketchChanged(_:)), name: .AGSSketchEditorGeometryDidChange, object: nil)
```
philium: FYI: Most of the time notification actions in view controllers are updating the view. When that's the case, it make more sense to tie the lifetime of the notification to the lifetime of the visibility of the view. Register for it in `viewWillAppear` and unregister for the same one in `viewDidDisappear`.

arcgis-ios-sdk-samples/Display information/Sketch on the map/SketchViewController.swift
qsmmorgan: Oh, if the observation removal isn't necessary then yeah there isn't much point to the rule.

PR #580, opened 2018-11-14
Enables prefixed_toplevel_constant, unavailable_function, and untyped_error_in_catch

qsmmorgan:


.swiftlint.yml
```diff
@@ -51,13 +51,16 @@ opt_in_rules: # some rules are only opt-in
...
   - overridden_super_call
   - override_in_extension
+  - prefixed_toplevel_constant
   - prohibited_super_call
   - redundant_nil_coalescing
   - redundant_type_annotation
   - sorted_first_last
   - static_operator
   - toggle_bool
+  - unavailable_function
```
philium: While I understand the motivation for this rule, it seems like a solution without a problem. The most common violator is `init?(coder:)`. The compiler will generate a missing implementation that calls `fatalError()`, but enabling this rule also requires the developer to manually add the missing `@available` attribute. That isn't such a big deal, but to my knowledge, these functions have never posed a problem in the past. So enabling this rule causes extra work for little to no benefit.

qsmmorgan: The main benefit I considered would be the compiler flagging erroneous usage of these functions before building. That said, I'm fine disabling it too.

qsmmorgan:
@philium We talked about `prefixed_toplevel_constant` after this PR was merged. I'm fine disabling that in the future too.

PR #581, opened 2018-11-15
Enables the identifier_name SwiftLint rule

qsmmorgan:
@philium I started this before you commented on the macOS version. I think it's a decent rule. Even if you don't want it enabled, I think the app would benefit from the resultant changes.

### DIFF
--- a/arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
+++ b/arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
@@ -299,18 +299,10 @@ class StatisticalQueryGroupAndSortViewController: UIViewController, UITableViewD
                 selectedGroupByFieldNames.remove(at: indexPath.row)
                 
                 // Remove field from the order by fields
-                for (index, orderByField) in orderByFields.enumerated().reversed() {
-                    if orderByField.fieldName == selectedGroupByFieldName {
-                        orderByFields.remove(at: index)
-                    }
-                }
+                orderByFields.removeAll { $0.fieldName == selectedGroupByFieldName }
                 
                 // Remove field from the selected order by fields
-                for (index, selectedOrderByField) in selectedOrderByFields.enumerated().reversed() {
-                    if selectedOrderByField.fieldName == selectedGroupByFieldName {
-                        selectedOrderByFields.remove(at: index)
-                    }
-                }
+                selectedOrderByFields.removeAll { $0.fieldName == selectedGroupByFieldName }
             }
         default:
             if !selectedOrderByFields.isEmpty {
@@ -349,10 +341,8 @@ class StatisticalQueryGroupAndSortViewController: UIViewController, UITableViewD
         }
         
         // Remove selected order by field if it's not part of the new order by fields
-        for (index, orderByField) in selectedOrderByFields.enumerated().reversed() {
-            if !orderByFields.contains(where: { $0.fieldName == orderByField.fieldName }) {
-                selectedOrderByFields.remove(at: index)
-            }
+        selectedOrderByFields.removeAll { (orderBy) -> Bool in
+            !orderByFields.contains { $0.fieldName == orderBy.fieldName }
         }
         
         // Reload sections
--- a/arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
+++ b/arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
@@ -299,16 +299,16 @@ class StatisticalQueryGroupAndSortViewController: UIViewController, UITableViewD
                 selectedGroupByFieldNames.remove(at: indexPath.row)
                 
                 // Remove field from the order by fields
-                for (i, orderByField) in orderByFields.enumerated().reversed() {
+                for (index, orderByField) in orderByFields.enumerated().reversed() {
                     if orderByField.fieldName == selectedGroupByFieldName {
-                        orderByFields.remove(at: i)
+                        orderByFields.remove(at: index)
                     }
                 }
                 
                 // Remove field from the selected order by fields
-                for (i, selectedOrderByField) in selectedOrderByFields.enumerated().reversed() {
+                for (index, selectedOrderByField) in selectedOrderByFields.enumerated().reversed() {
                     if selectedOrderByField.fieldName == selectedGroupByFieldName {
-                        selectedOrderByFields.remove(at: i)
+                        selectedOrderByFields.remove(at: index)
                     }
                 }
             }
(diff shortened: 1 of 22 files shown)


qsmmorgan:
@philium I removed the length requirement so other developers don't have to worry about which short identifiers are allowed. We should still keep an eye out for instances of `i` or `id` in the future that should be more descriptive.

arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
```diff
@@ -299,16 +299,16 @@ class StatisticalQueryGroupAndSortViewController: UIViewController, UITableViewD
                 selectedGroupByFieldNames.remove(at: indexPath.row)
                 
                 // Remove field from the order by fields
-                for (i, orderByField) in orderByFields.enumerated().reversed() {
+                for (index, orderByField) in orderByFields.enumerated().reversed() {
```
philium: These `for` loops should really just be replaced by a call to `remove(where:)`, but that can be done as part of a separate PR.

arcgis-ios-sdk-samples/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
```diff
@@ -20,7 +20,7 @@ class FeatureLayerRenderingModeMapViewController: UIViewController {
     @IBOutlet weak var dynamicMapView: AGSMapView!
     @IBOutlet weak var staticMapView: AGSMapView!
     
-    var _zoomed = false
+    private var isZoomed = false
```
philium: I think this should just be `zoomed` since it is mutable. From the Swift API Design Guidelines:

> Uses of Boolean methods and properties should read as assertions about the receiver when the use is nonmutating

arcgis-ios-sdk-samples/Maps/Display device location/CustomContextSheet.swift
```diff
@@ -104,9 +104,9 @@ class CustomContextSheet: UIView {
         }
         
         //other buttons
-        for i in 0..<images.count {
+        for (index, image) in images.enumerated() {
```
philium: The following construct is preferred over `enumerated()`:

```swift
for index in collection.indices {
    let element = collection[index]
    …
}
```

Because the first element of the tuple is an *offset*, not an index. It always starts at `0`, which is not the case for the indices of every collection. The [documentation](https://developer.apple.com/documentation/swift/enumeratedsequence) uses the term `offset` throughout.

arcgis-ios-sdk-samples/Route & Directions/Find service area interactive/FindServiceAreaInteractiveVC.swift
```diff
@@ -178,9 +178,8 @@ class FindServiceAreaInteractiveVC: UIViewController, AGSGeoViewTouchDelegate, S
                 //are the same across facilities, we only need to draw the resultPolygons at
                 //facility index 0. It will contain either merged or multipart polygons
                 if let polygons = result?.resultPolygons(atFacilityIndex: 0) {
-                    for j in 0..<polygons.count {
-                        let polygon = polygons[j]
-                        let fillSymbol = self.serviceAreaSymbol(for: j)
+                    for (index, polygon) in polygons.enumerated() {
```
philium: Please use `for index in polygons.indices { … }` instead.

arcgis-ios-sdk-samples/Scenes/Extrude graphics/ExtrudeGraphicsViewController.swift
```diff
@@ -74,9 +74,9 @@ class ExtrudeGraphicsViewController: UIViewController {
         let y = self.cameraStartingPoint.y + 0.25
         
         //creating a grid of polygon graphics
-        for i in 0...6 {
-            for j in 0...4 {
-                let polygon = self.polygonForStartingPoint(AGSPoint(x: x + Double(i) * (squareSize + spacing), y: y + Double(j) * (squareSize + spacing), spatialReference: nil))
+        for column in 0...6 {
```
philium: Since `column` and `row` are not actually used as integers, it would be nice to update the for loops to use `stride` instead:

```swift
for column in stride(from: 0.0, through: 6.0, by: 1.0) {
    for row in stride(from: 0.0, through: 4.0, by: 1.0) {
        let polygon = self.polygonForStartingPoint(AGSPoint(x: x + column * (squareSize + spacing), y: y + row * (squareSize + spacing), spatialReference: nil))
        self.addGraphicForPolygon(polygon)
    }
}
```

arcgis-ios-sdk-samples/Scenes/Scene symbols/SceneSymbolsViewController.swift
```diff
@@ -66,8 +66,8 @@ class SceneSymbolsViewController: UIViewController {
         
         // Create graphics for each symbol.
         let symbols = [coneSymbol, cubeSymbol, cylinderSymbol, diamondSymbol, sphereSymbol, tetrahedronSymbol]
-        let graphics = symbols.enumerated().map { (i, symbol) -> AGSGraphic in
-            let point = AGSPoint(x: x + 0.01 * Double(i), y: y, z: z, spatialReference: .wgs84())
+        let graphics = symbols.enumerated().map { (index, symbol) -> AGSGraphic in
```
philium: `index` -> `offset`

arcgis-ios-sdk-samples/Shared resources/Libraries/ExpandableTableViewController/ExpandableTableViewController.swift
```diff
@@ -165,13 +165,13 @@ class ExpandableTableViewController: UITableViewController {
             UIView.animate(withDuration: 0.25, animations: {
                 imageView.image = UIImage(named: "Collapsed")
             })
-            var indexesPath = [IndexPath]()
-            for i in 0 ..< sectionData.count {
-                let index = IndexPath(row: i, section: section)
-                indexesPath.append(index)
+            var indexPaths = [IndexPath]()
```
philium: How about this?

```swift
let indexPaths = sectionData.indices.map { IndexPath(row: $0, section: section) }
```

Same with lines `189` through `193`.

arcgis-ios-sdk-samples/Maps/Display device location/CustomContextSheet.swift
qsmmorgan: Thanks for the tip! I didn't realize it didn't necessarily use the indexes.

arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
qsmmorgan: Good call, I made the change since I would have to edit it either way.

qsmmorgan:
@philium I made the changes so this is ready for re-review.

arcgis-ios-sdk-samples/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
```diff
@@ -68,14 +68,14 @@ class FeatureLayerRenderingModeMapViewController: UIViewController {
...
-        if _zoomed {
+        if zoomed {
             self.dynamicMapView.setViewpoint(zoomedOutViewpoint, duration: 5, completion: nil)
             self.staticMapView.setViewpoint(zoomedOutViewpoint, duration: 5, completion: nil)
         } else {
             self.dynamicMapView.setViewpoint(zoomedInViewpoint, duration: 5, completion: nil)
             self.staticMapView.setViewpoint(zoomedInViewpoint, duration: 5, completion: nil)
         }
-        _zoomed = !_zoomed
+        zoomed = !zoomed
```
philium:
```suggestion
        zoomed.toggle()
```

PR #584, opened 2018-11-16
Enables the function_default_parameter_at_end SwiftLint rule

qsmmorgan:
The only function that needed correcting was `func presentAlert(title: String? = nil, message: String)`, so I just changed `title` to required and `message` to default to `nil`. This has no user-facing effects since iOS displays the message as the title if the title is `nil` anyway.

### DIFF
--- a/arcgis-ios-sdk-samples/Analysis/Analyze hotspots/HotspotSettingsViewController.swift
+++ b/arcgis-ios-sdk-samples/Analysis/Analyze hotspots/HotspotSettingsViewController.swift
@@ -65,7 +65,7 @@ class HotspotSettingsViewController: UIViewController, UITextFieldDelegate {
         if !self.toTextField.text!.isEmpty && !self.fromTextField.text!.isEmpty {
             self.delegate?.hotspotSettingsViewController(self, didSelectDates: self.fromTextField.text!, toDate: self.toTextField.text!)
         } else {
-            presentAlert(message: "Both dates are required")
+            presentAlert(title: "Both dates are required")
         }
     }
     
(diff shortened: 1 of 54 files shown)


arcgis-ios-sdk-samples/Analysis/Analyze hotspots/HotspotSettingsViewController.swift
```diff
@@ -65,7 +65,7 @@ class HotspotSettingsViewController: UIViewController, UITextFieldDelegate {
         if !self.toTextField.text!.isEmpty && !self.fromTextField.text!.isEmpty {
             self.delegate?.hotspotSettingsViewController(self, didSelectDates: self.fromTextField.text!, toDate: self.toTextField.text!)
         } else {
-            presentAlert(message: "Both dates are required")
+            presentAlert(title: "Both dates are required")
```
philium: Both title and message are displayed the same way (like a title) when the corresponding one is `nil`? Fascinating. I did not know that. However, the message does appear like a message when the title is `""`. Did you consider doing that? Most of these alert messages really are more appropriate as messages than titles.

philium: Since most of these shouldn't be titles, I would not use `title` as the parameter label. I would suggest either not enabling the rule or modifying the function as follows to satisfy the rule:

```swift
func presentAlert(title: String = "", message: String? = nil) { … }
```

That way the rule is satisfied, there is no change needed to the existing call sites, and alert messages appear with message style rather than title style. It does allow the function to be called with no parameters, but that would be hard to do by mistake and would be no different that calling the function with "" and `nil` for the parameters, which is allowed either way.

qsmmorgan: Done.

PR #585, opened 2018-11-16
Replaces ExpandableTableViewController in Spatial Relationships

qsmmorgan:


qsmmorgan:
@philium I see your point, but I'm not sure it makes for a better sample. It's nice to keep the core SDK functionality in as few files as possible. Right now the table view controller is just a necessity for displaying the results and developers can safely ignore it.

philium:
> It's nice to keep the core SDK functionality in as few files as possible.

It may be nice, but it isn't always practical. Nor is it what customers would do in their own apps. If the issue is having two different files, then simply put both classes in the same file.

> the table view controller is just a necessity for displaying the results

Since that's the case, the class should just be a data source used by a stock `UITableViewController`:

```swift
class SpatialRelationshipsTableViewDataSource: NSObject, UITableViewDataSource {
    struct Section {
        let relationships: [String]
        let title: String
    }
    
    var sections: [Section] = []
    
    func numberOfSections(in tableView: UITableView) -> Int {
        return sections.count
    }
    
    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
        return sections[section].relationships.count
    }
    
    func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
        return sections[section].title
    }
    
    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
        let cell = tableView.dequeueReusableCell(withIdentifier: "SpatialRelationshipCell", for: indexPath)
        cell.textLabel?.text = sections[indexPath.section].relationships[indexPath.row]
        return cell
    }
}
```

qsmmorgan:
I appreciate the sentiment but it doesn't seem productive to rewrite it now.

philium:
> it doesn't seem productive to rewrite it now.

I disagree. You're introducing a new class with new functionality, not just moving old code around. Taking the time to make new code as good as reasonably possible before introducing it to the working branch is productive in my opinion. Whether that's by limiting the scope of the class to make it solely a table view data source or by expanding the class to be more of a controller is up to you.

qsmmorgan:
@philium I moved the data source to be an extension of the root view controller for the sample. Let me know if that works or if you want me to change it.

philium:
> I moved the data source to be an extension of the root view controller

I appreciate the time and effort you put into making that change, but I don't think making the sample view controller a data source of another view controller's table view is a good solution. It keeps everything in one class, but that isn't a good development pattern and probably not something we want to model to other developers.

I just reread the the documentation for `UITableViewController` and Apple is pretty insistent on subclassing it:

> Subclass `UITableViewController` when your interface consists of a table view and little or no other content.

> You create a custom subclass of `UITableViewController` for each table view that you want to manage.

Please revert the last commit and I'll approve the PR. Subclassing `UITableViewController` just to implement the data source probably isn't the way I would do it, but it is Apple approved. So be it.

PR #586, opened 2018-11-16
Enables the cyclomatic_complexity SwiftLint rule

qsmmorgan:


arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
```diff
@@ -28,9 +28,23 @@ class StatisticalQueryGroupAndSortViewController: UIViewController, UITableViewD
     private var orderByFields = [AGSOrderBy]()
     private var selectedOrderByFields = [AGSOrderBy]()
     private var statisticDefinitions = [AGSStatisticDefinition]()
-    private var sectionHeaderTitles = ["1. Add Statistic Definitions", "2. Select Group By Fields", "3. Select Order By Fields"]
     private var statisticTypes = ["Average", "Count", "Maximum", "Minimum", "StandardDeviation", "Sum", "Variance"]
     
+    private enum Section: Int, CaseIterable {
```
philium: Since you're conforming to `CaseIterable`, you could eliminate the raw value and avoid the optional initializer by using `allCases` to do the lookup:

```swift
switch Section.allCases[section] { … }
```

Please do that instead.

arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
```diff
@@ -28,9 +28,23 @@ class StatisticalQueryGroupAndSortViewController: UIViewController, UITableViewD
...
+        case definitions, groupByFields, orderByFields
+        
+        var title: String {
```
philium: I know I've suggested this before, but since this is only useful in `titleForHeaderInSection`, I would suggest moving it there. It eliminates a level of indirection and keeps the interface of `Section` nice and concise.

PS: In general, it is helpful to keep the definition of value types as concise as possible, particular when they are subtypes. One way of doing that is to move computed properties into an extension.

arcgis-ios-sdk-samples/Shared resources/Libraries/DemoTouchView/DemoTouchView.swift
```diff
@@ -108,58 +108,27 @@ extension UIApplication {
         
         if let touches = event.allTouches {
             
-            var began: Set<UITouch>?
-            var moved: Set<UITouch>?
-            var cancelled: Set<UITouch>?
-            var ended: Set<UITouch>?
-            
-            for touch in touches {
```
philium: The `for` loop was actually more efficient than four separate calls to `filter`. Not a big deal because the size of `allTouches` is going to be fairly small, but something to keep in mind. One way of simplifying this method while maintaining efficiency is by extending `UIEvent`:

```swift
extension UIEvent {
    var touchesByPhase: [UITouch.Phase: Set<UITouch>] {
        return allTouches?.reduce(into: [UITouch.Phase: Set<UITouch>](), { (result, touch) in
            result[touch.phase, default: []].insert(touch)
        }) ?? [:]
    }
}
```

arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
qsmmorgan: I see your point but this seems less standard to me than using the initializer and we don't use this pattern anywhere else in the sample. Plus it'll still crash if the index is out of bounds so it's not any safer than force-unwrapping the init. I'll still change it if you really want.

arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
qsmmorgan: That's a fine point.

arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
philium: > this seems less standard to me than using the initializer and we don't use this pattern anywhere else in the sample

`CaseIterable` is a fairly new protocol. Still, we wouldn't be the first ones to [use that approach](https://oleb.net/blog/2018/06/enumerating-enum-cases/).

> Plus it'll still crash if the index is out of bounds so it's not any safer than force-unwrapping the init

It isn't any safer, but it isn't any less safe. The advantage is 1) not needing a raw representation for the enum cases, and 2) not needing to deal with optionals. 

> I'll still change it if you really want.

I would. We're already taking advantage of the `CaseIterable` conformance in `numberOfSections(in:)`. It would be nice to take advantage of it in the other data source methods and get rid of the raw values.

arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
```diff
@@ -28,9 +28,12 @@ class StatisticalQueryGroupAndSortViewController: UIViewController, UITableViewD
     private var orderByFields = [AGSOrderBy]()
     private var selectedOrderByFields = [AGSOrderBy]()
     private var statisticDefinitions = [AGSStatisticDefinition]()
-    private var sectionHeaderTitles = ["1. Add Statistic Definitions", "2. Select Group By Fields", "3. Select Order By Fields"]
     private var statisticTypes = ["Average", "Count", "Maximum", "Minimum", "StandardDeviation", "Sum", "Variance"]
     
+    private enum Section: Int, CaseIterable {
```
philium: Please remove the raw representation. It is no longer needed.
```suggestion
    private enum Section: CaseIterable {
```

PR #589, opened 2018-11-19
Enables the attributes SwiftLint rule

qsmmorgan:


rolson:
For swift lint PRs please just work with @philium . Please don't include anybody else from the api team on the swift lint PRs.

cc @dg0yal @pgruenler

qsmmorgan:
By dropping the `Timer` selector we can drop the attributes which removes the SwiftLint warning. I don't see the usefulness of splitting them into another PR at this point.

philium:
> By dropping the Timer selector we can drop the attributes which removes the SwiftLint warning

I'm al for switching to a block-based API for creating timers, but that could have been done before (or after) enabling the linting rule. If done before, the function wouldn't get flagged by the linter. If done after, the attribute could simply be removed.

> I don't see the usefulness of splitting them into another PR at this point.

The title of the PR is "Enables the attributes SwiftLint rule". Is changing to a different `Timer` creation function necessary to enable the rule? No. And in one case, you're changing the run loop mode in which a timer is being added from `common` to `default`. That may or may not change the app's behavior, but it definitely is unrelated to enabling the linting rule. The safest (and simplest) change to satisfy the rule would have been to add the `@objc` attribute.

qsmmorgan:
I get that this could have been separate PRs, but these are precious small changes that are still related to the issue and are documented in the PR comment. If I pushed them separately I would have had to wait for one to be merged first and then deal with merge conflicts.

qsmmorgan:
@philium I reverted the `Timer` changes. Should be ready for re-review.

PR #590, opened 2018-11-19
Enables the conditional_returns_on_newline SwiftLint rule

qsmmorgan:


philium:
I'll give it some more thought, but I don't think I'm in favor of enabling this rule. The most frequent (only?) violator is `guard` statements. The `else` clause of a `guard` statement is unique and different from `if` statements clauses in that it must transfer control out of the current scope. That is most frequently done with a return statement. Because of that, some guard statements are quite short. Here are some examples:

```swift
guard isViewLoaded else { return }
guard let table = sublayer.table else { continue }
```

All of those are easily read and understood, even though the whole `guard` statement is on a single line. Sometimes we even know what the `else` clause does before reading it, just based on the context. Breaking them up into multiple lines may seem like it makes the statement easier to read, but I think it actually makes everything else harder to read. It unnecessarily makes the enclosing scope(s) longer and more difficult to scan.

The more I think about it, the more I don't like the rule. The decision to put the `else` clause on the same line or a separate line is a subjective thing and shouldn't be a linting rule. My general guideline is that if the `guard` condition is "too long" (again, subjective) or if the condition itself is multiple lines, then the body of the `else` clause goes on a separate line.

qsmmorgan:
Closing this since it doesn't seem like we'll be enabling this rule.

PR #594, opened 2018-11-20
Removes unneeded ATS exception domains

qsmmorgan:


qsmmorgan:
@philium Please re-review.

PR #597, opened 2018-11-20
Improves Edit Feature Attachments

qsmmorgan:
- Renames files to match the sample
- Corrects the order of the files in the project
- Adds the second view controller file to the source code viewer
- Makes the attachments view controller a UITableViewController subclass
- Uses a navigation controller rather than a navigation bar in the attachments view
- Makes the attachment changes save only when tapping Done
- Adds a status HUD for downloading attachment image data


arcgis-ios-sdk-samples/Edit data/Edit feature attachments/AttachmentsTableViewController.swift
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+    }
+    
+    // MARK: - Actions
+    
+    @IBAction func doneAction() {
+        
+        guard let table = feature?.featureTable as? AGSServiceFeatureTable else {
+            dismiss(animated: true, completion: nil)
```
philium: I would strongly suggest not doing any work inside the `else` clause of a `guard` statement, other than perhaps logging. The whole purpose of a `guard` statement is to transfer control out of the current scope--bail early, bail often. Doing additional work before balling changes that and makes the whole function harder to reason about.

I would suggestion using an `if` statement:

```swift
if let table = feature?.featureTable as? AGSServiceFeatureTable {
    // Apply edits to the table…
} else {
    dismiss(animated: true)
}
```

Or you could break it up into two different functions:

```swift
@IBAction func doneAction() {
    //show progress hud
    SVProgressHUD.show(withStatus: "Applying edits")
    
    applyEdits { [weak self] (error) in
        
        //dismiss progress hud
        SVProgressHUD.dismiss()
        
        guard let self = self else {
            return
        }
        
        if let error = error {
        }
        self.dismiss(animated: true)
    }
    
}

func applyEdits(completion: @escaping (_ error: Error?) -> Void) {
    if let table = feature?.featureTable as? AGSServiceFeatureTable {
        table.applyEdits { completion($1) }
    } else {
        completion(nil)
    }
}
```

arcgis-ios-sdk-samples/Edit data/Edit feature attachments/EditFeatureAttachmentsViewController.swift
```diff
@@ -0,0 +1,121 @@
...
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        
+        if let lastQuery = lastQuery {
+            lastQuery.cancel()
+        }
+        
+        //hide the callout
+        mapView.callout.dismiss()
+        
+        guard let featureLayer = featureLayer else {
```
philium: This seems like such a silly check to do every time because we know there will always be a feature layer. It would be nice to either make `featureLayer` an implicitly unwrapped optional or make it non-optional by creating it when the view controller is initialized.

arcgis-ios-sdk-samples/Edit data/Edit feature attachments/AttachmentsTableViewController.swift
```diff
@@ -0,0 +1,174 @@
...
+            
+            guard let self = self else {
+                return
+            }
+            
+            if let error = error {
+                self.presentAlert(error: error)
+            }
+            
+            self.attachments = attachments ?? []
```
philium: This might make a little more sense as an `else if`, since we know that either `attachments` or `error` will be non-`nil`. I think that's how we do it in most other places:

```swift
} else if let attachments = attachments {
    self.attachments = attachments
    self.tableView.reloadData()
}
```

arcgis-ios-sdk-samples/Edit data/Edit feature attachments/AttachmentsTableViewController.swift
```diff
@@ -0,0 +1,174 @@
...
+
+class AttachmentsTableViewController: UITableViewController {
+    
+    weak var feature: AGSArcGISFeature?
+    private var attachments: [AGSAttachment] = []
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        loadAttachments()
```
philium: There really isn't a good reason to tie the loading of attachments to the loading of the view controller's view. I would suggest calling this in a `didSet` observer of `feature`.

arcgis-ios-sdk-samples/Edit data/Edit feature attachments/AttachmentsTableViewController.swift
```diff
@@ -0,0 +1,174 @@
...
+            self.tableView.reloadData()
+        }
+    }
+    
+    private func deleteAttachment(_ attachment: AGSAttachment) {
+        feature?.delete(attachment) { [weak self] (error: Error?) in
+            if let error = error {
+                print(error)
+            } else {
+                self?.loadAttachments()
```
philium: Do you know why `loadAttachments()` is called here instead of removing the attachment from `attachments` and removing the appropriate row from the table? It seems like that would be much more efficient. Same question in `addAction()`. It is rarely a good idea to simply call `reloadData()` to add or remove a single row.

arcgis-ios-sdk-samples/Edit data/Edit feature attachments/AttachmentsTableViewController.swift
```diff
@@ -0,0 +1,174 @@
...
+    }
+    
+    override func tableView(_ tableView: UITableView, commit editingStyle: UITableViewCell.EditingStyle, forRowAt indexPath: IndexPath) {
+        if editingStyle == .delete {
+            let attachment = attachments[indexPath.row]
+            deleteAttachment(attachment)
+        }
+    }
+    
+    private func downloadAndSetAttachmentImageForCell(_ cell: UITableViewCell, at indexPath: IndexPath) {
```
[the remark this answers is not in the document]
qsmmorgan: I tried something like this but ran into an issue where `cellForRow` would return `nil` when this function is called during `cellForRowAtIndexPath`. I can try to find a workaround.

arcgis-ios-sdk-samples/Edit data/Edit feature attachments/AttachmentsTableViewController.swift
qsmmorgan: In this case we want the table to be cleared if there was an error so as not to show out-of-date data. I'll still change it to make that clearer.

qsmmorgan:
@philium I think I addressed all your comments. This is ready for re-review.

arcgis-ios-sdk-samples/Edit data/Edit feature attachments/AttachmentsTableViewController.swift
philium: That makes sense. I wrote that comment before I realized that this is being called for additions and deletions.

arcgis-ios-sdk-samples/Edit data/Edit feature attachments/AttachmentsTableViewController.swift
```diff
@@ -41,53 +40,90 @@ class AttachmentsTableViewController: UITableViewController {
...
+            cell.imageView?.image = UIImage(named: "CloudDownload", in: AGSBundle(), compatibleWith: nil)
         }
         
         return cell
     }
     
+    // MARK: - UITableViewDelegate
+    
+    override func tableView(_ tableView: UITableView, shouldHighlightRowAt indexPath: IndexPath) -> Bool {
+        return !attachments[indexPath.row].hasFetchedData
```
philium: 👍 Nice! It is still a bit odd that the download button isn't actually a button (usually on the right side of the cell), but that isn't your fault. This is definitely better than it way.

PR #598, opened 2018-11-21
Improves error handling in offline routing

qsmmorgan:


gdhakal:
@qsmmorgan just curious why the error is printed in console. I think it would be nice to display error message in the app using SVProgressHUD instead.

qsmmorgan:
@gdhakal That may indeed be a better solution. For this PR I just left the error reporting as it had been.

gdhakal:
@qsmmorgan sure, that sounds good.

PR #599, opened 2018-11-21
Adds a custom SwiftLint rule enforcing a single space between { and [ in a closure

qsmmorgan:


.swiftlint.yml
```diff
@@ -7,6 +7,10 @@ custom_rules:
     name: WGS84 Spatial Reference
     regex: AGSSpatialReference\(wkid:\s*(4326)\)
     message: "Use AGSSpatialReference.wgs84() instead."
+  closure_opening_space:
+    name: Closure Opening Space
+    regex: \{\[|\{\s{2,}\[
```
philium: I'm not an expert on regular expressions by any means, but might this be a little easier to understand? I think it is equivalent.

```suggestion
    regex: \{(?! )\[
```

qsmmorgan: Mine flags `{[` as well as instances with more than one whitespace between them, such as `{   [`. In essence the rule enforces one and only one space. Not sure if there's a simpler way to write that but feel free to propose one.

philium: > Not sure if there's a simpler way to write that but feel free to propose one.

I did.

screen shot 2018-11-21 at 12 56 30 pm

qsmmorgan: Huh. I tried it in RegexPal and it didn't match multiple spaces.

philium: Apparently what I came up with isn't the same, although I can't explain why. I did say I wasn't an expert!

PR #600, opened 2018-11-21
Adds a custom SwiftLint rule flagging redundant "-> Void" in closure definitions

qsmmorgan:


arcgis-ios-sdk-samples/Content Display Logic/Model/SampleSearchEngine.swift
```diff
@@ -25,7 +25,7 @@ class SampleSearchEngine {
         self.samples = samples
         
         isLoadingReadmeIndex = true
-        DispatchQueue.global(qos: .background).async { [weak self] () -> Void in
+        DispatchQueue.global(qos: .background).async { [weak self] () in
```
philium: It isn't just the `Void` that is unnecessary; the `()` isn't needed either. What would you think of something like this?

```yml
  infer_closure_form:
     name: Infer Closure Form
     regex: \(\)( -> Void)? in
     message: Closure form can be inferred.
```

The documentation states that "you can still make the types explicit if you wish, and doing so is encouraged if it avoids ambiguity for readers of your code." But I think I agree with you that being explicit in this case doesn't add any value.

qsmmorgan: I wouldn't mind that, but it should be a different rule since `-> Void` can apply to closures with parameters.

philium: So we would need three rules? One for `() -> Void in`, one for `() in`, and one for `-> Void`? I think those could be combined into a single rule. If not, perhaps this doesn't need to be a rule. Neither of the three is needed, but neither of the three is wrong. There might be more important things to work on.

qsmmorgan: Nope, just for `() in` and `-> Void in`, since fixing the latter will create and flag the former, if applicable. It probably could be reduced to one though.

qsmmorgan: @philium I changed the rule to account for empty arguments as well.

PR #601, opened 2018-11-21
Updates the design of the sample table view cells

qsmmorgan:
This PR makes for a less drastic change than the previous redesign attempt. The basic interaction is the same, but there are a number of improvements:

- The cells are no longer contained in a grey box
- The table now uses the standard background and separators
- There is now a disclosure indicator indicating that tapping the cell opens a new view
- The system cell accessory is now used as the info detail button
- Multiple cells can be expanded at the same time
- Cells no longer expand upon tapping to open a sample

A larger overhaul could be considered in the future.

simulator screen shot - iphone xr - 2018-11-21 at 15 30 52


arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -49,37 +49,24 @@ class ContentTableViewController: UITableViewController {
...
-        } else {
-            cell.detailLabel.text = nil
-        }
-        
-        cell.infoButton.addTarget(self, action: #selector(ContentTableViewController.expandCell(_:)), for: .touchUpInside)
-        cell.infoButton.tag = indexPath.row
-
-        cell.backgroundColor = .clear
         
+        let cell = tableView.dequeueReusableCell(withIdentifier: "ContentTableCell") as! ContentTableCell
```
philium: Please use `UITableView.dequeueReusableCell(withIdentifier:for:)`.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -49,37 +49,24 @@ class ContentTableViewController: UITableViewController {
...
+        cell.titleLabel.text = sample.name
+        cell.detailLabel.text = sample.description
+        cell.isExpanded = expandedRowIndices.contains(indexPath.row)
```
philium: This is an interesting approach, but I think there might be a better one. One would be to use a subtitle cell and set the detail text label's text based one whether the cell should be expanded. That would be closer to what you have now, but without the need for a custom cell class. Another option would be to dequeue a basic cell or a subtitle cell based on whether the cell should be expanded. The advantage to that approach is that subviews are not being rearranged (potentially) each time a cell is loaded.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -205,25 +195,19 @@ class ContentTableViewController: UITableViewController {
 
     }
     
-    @objc
-    func expandCell(_ sender: UIButton) {
-        updateExpandedRow(IndexPath(row: sender.tag, section: 0), collapseIfSelected: true)
-    }
-    
     private func updateExpandedRow(_ indexPath: IndexPath, collapseIfSelected: Bool) {
```
philium: A better name for this function might be `toggleExpansion(at:)`. It looks like `true` is always passed for `collapseIfSelected`, so that parameter probably isn't needed.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -33,7 +33,7 @@ class ContentTableViewController: UITableViewController {
 
     var searchEngine: SampleSearchEngine?
     
-    private var expandedRowIndex: Int = -1
+    private var expandedRowIndices: Set<Int> = []
```
philium: You might consider just storing the index paths themselves: `Set<IndexPath>`

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -49,37 +49,24 @@ class ContentTableViewController: UITableViewController {
         self.downloadProgressView = downloadProgressView
     }
 
-    // MARK: - Table view data source
-
-    override func numberOfSections(in tableView: UITableView) -> Int {
```
philium: Even though `1` is the default value returned by the class when this method isn't implemented, I have rarely seen this method NOT implemented. An implementation for `numberOfSections(in:)` is almost always provided. I would suggest leaving it.

qsmmorgan: Are you sure? I never include it for single-section tables since it's unnecessary and I don't see how it otherwise adds value.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
qsmmorgan: In general I'd like that approach, but in this case I'd rather leave it as a custom subclass. I tried to use the detail cell in the prior redesign and received pushback for not having the control over vertical spacing that we wanted. Also, we'll need the subclass in order to achieve more advanced functionality in the future.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
qsmmorgan: Good point!

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -249,6 +228,9 @@ extension ContentTableViewController: UISearchResultsUpdating {
             return
         }
         
+        // do not preserve cell expansion when loading new results
+        expandedRowIndexPaths = []
```
philium: I'm sorry for not noticing this the first time around.

```suggestion
        expandedRowIndexPaths.removeAll()
```

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
philium: > Are you sure?

Pretty sure. After five years of cocoa development I didn't even know that `UITableViewController` provided an implementation for that method, let alone that the default is `1`. I saw the default implementation used for the first time a month or two ago. Every other time (with the exception of static cells) I've seen the data source provide its own implementation, even just to return `1`. Part of that might be because Xcode's template for a `UITableViewController` subclass provides an implementation that returns `0`. It is easiest to just change `0` to `1`. Anecdotally, the `MasterViewController` class of the Master-Detail app template provides an implementation that returns `1`.

That being said, `UITableViewController` does provide a default implementation that returns `1` when it could throw. So it isn't wrong to use the default implementation here. That's why I made it a suggestion and not an imperative.

qsmmorgan: Interesting. In this case I'd rather just leave it off.

arcgis-ios-sdk-samples/Content Display Logic/Cells/ContentTableCell.swift
```diff
@@ -16,46 +16,21 @@ import UIKit
...
-            self.parentView.backgroundColor = .secondaryBlue
-            self.titleLabel.textColor = .white
-            self.detailLabel.textColor = .white
-            self.infoButton.tintColor = .white
-        } else {
-            self.parentView.backgroundColor = .white
-            self.titleLabel.textColor = .primaryTextColor
-            self.detailLabel.textColor = .secondaryTextColor
-            self.infoButton.tintColor = .secondaryBlue
+    var isExpanded: Bool = false {
```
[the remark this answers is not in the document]
qsmmorgan: Oh you're right. Interesting, typically when you hide a view it still participates in the layout.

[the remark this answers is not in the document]
qsmmorgan: We're currently getting the desired animation behavior so I don't think we need a way to manually flag animations right now.

qsmmorgan:
This should be ready for re-review.

arcgis-ios-sdk-samples/Content Display Logic/Cells/ContentTableCell.swift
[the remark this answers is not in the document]
philium: > We're currently getting the desired animation behavior

Ah, I see. You're animating the expansion by reloading the cell. That works, but it has a couple drawbacks:

1. It causes the cell to be reloaded. That doesn't really change anything other than updating the `isExpanded` state of the cell. It would be more efficient to set that directly.
2. The animation is a bit strange. The detail button disappears and reappears in a different spot rather than moving to the new location.

I played around with animating the expansion rather than the reloading of the cell. Forget what I said about having the cell perform that animation; that doesn't work. There's no way that I know of for the cell to trigger a resize of itself inside the table view. This almost works:

```swift
if cell.isExpanded {
    UIView.animate(withDuration: 0.2, animations: {
        cell.detailLabel.alpha = 0
    }, completion: { (finished) in
        cell.isExpanded.toggle()
        self.tableView.performBatchUpdates(nil)
    })
} else {
    cell.detailLabel.alpha = 0
    cell.isExpanded.toggle()
    tableView.performBatchUpdates(nil) { finished in
        guard finished else { return }
        UIView.animate(withDuration: 0.2, animations: {
            cell.detailLabel.alpha = 1
        })
    }
}
```

The only remaining problem is that on collapse, the `titleLabel`'s frame is immediately expanded to fill the whole stack view. And since a label's text is vertically centered, the title jumps and then animates back to where it should have been all along. That can be fixed by subclassing `UILabel` to vertically align the text to the top, but that's probably an exercise for another day. Reloading the cell works well enough for now.

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/Main.storyboard
```diff
@@ -203,121 +203,47 @@
...
-                                <rect key="frame" x="0.0" y="66" width="375" height="80"/>
-                                <autoresizingMask key="autoresizingMask"/>
                                 <tableViewCellContentView key="contentView" opaque="NO" clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="center" tableViewCell="UHI-3v-SDe" id="hjB-p9-kF0">
-                                    <rect key="frame" x="0.0" y="0.0" width="375" height="80"/>
+                                    <rect key="frame" x="0.0" y="0.0" width="307" height="43.666666666666664"/>
                                     <autoresizingMask key="autoresizingMask"/>
                                     <subviews>
-                                        <view contentMode="scaleToFill" placeholderIntrinsicWidth="580" placeholderIntrinsicHeight="62" translatesAutoresizingMaskIntoConstraints="NO" id="jE0-B2-BRg">
-                                            <rect key="frame" x="18" y="12" width="339" height="56"/>
+                                        <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" spacing="6" translatesAutoresizingMaskIntoConstraints="NO" id="pVi-Nv-PRP">
```
philium: One more thing I noticed. The alignment should be Leading (it shows up as `top` in the storyboard). The way it is now, the labels are sized to fit the entire width of the stack view:

screen shot 2018-11-30 at 10 58 56 am

By using leading alignment, the labels are sized to fit their content:

screen shot 2018-11-30 at 10 56 59 am

arcgis-ios-sdk-samples/Content Display Logic/Cells/ContentTableCell.swift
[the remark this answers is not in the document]
qsmmorgan: Thanks for looking into this. It also struck me as awkward to reload the cells in order to expand them (that was the existing behavior, I think). You're also right about the odd detail button animation. I didn't look into it further since it's still better than it was and we can change it later. It's also not in actual sample code so it's unlikely developers will be studying it.

PR #604, opened 2018-11-26
Updates the Analyze Hotspot settings to use date pickers

qsmmorgan:


arcgis-ios-sdk-samples/Analysis/Analyze hotspots/AnalyzeHotspots.storyboard
```diff
@@ -31,156 +31,142 @@
...
+                            <outlet property="delegate" destination="8cL-Af-ycY" id="XVw-aS-jnK"/>
+                        </connections>
+                    </tableView>
+                    <navigationItem key="navigationItem" title="Analysis Date Range" id="1ds-Jt-ObJ">
+                        <barButtonItem key="leftBarButtonItem" systemItem="cancel" id="mkw-Fc-h6Q">
+                            <connections>
+                                <action selector="cancelAction:" destination="8cL-Af-ycY" id="Tbs-4U-YSa"/>
+                            </connections>
+                        </barButtonItem>
+                        <barButtonItem key="rightBarButtonItem" systemItem="done" id="gFz-M8-OCs">
```
philium: The style of this item should be `done`:

```suggestion
                        <barButtonItem key="rightBarButtonItem" style="done" systemItem="done" id="gFz-M8-OCs">
```

arcgis-ios-sdk-samples/Analysis/Analyze hotspots/HotspotSettingsViewController.swift
```diff
@@ -16,78 +16,28 @@ import UIKit
...
 }
 
-class HotspotSettingsViewController: UIViewController, UITextFieldDelegate {
-
-    @IBOutlet var fromTextField: UITextField!
-    @IBOutlet var toTextField: UITextField!
-    
-    private var datePicker: UIDatePicker!
-    private var dateFormatter: DateFormatter!
+class HotspotSettingsViewController: UITableViewController {
```
philium: Using a different section for each picker is an interesting idea, but it doesn't seem to fit with the human interface guidelines for [pickers](https://developer.apple.com/design/human-interface-guidelines/ios/controls/pickers/). It says that "Pickers can also appear inline, such as while editing a date in a Calendar event." The calendar event view is actually a really good example because it provides a UI for picking a date range, just as this view does. It would be nice to have a view more similar to that. A grouped table view with one section with four rows, two for displaying the selected dates and two for the pickers themselves. The data source maintains whether the pickers should be shown a adds and removes the corresponding rows as appropriate. I've designed similar interfaces before, so I'd be happy to help you with the data source implementation.

With a view like that, it might make more sense to display the settings view controller as a popover. It might be useful to see the date range and the map view at the same time.

qsmmorgan: The HIG doesn't discourage using pickers like I did. I do like your solution better, but at this point I don't feel much need to change it. 

The calculations take quite awhile, so it's not feasible to update the view every time a date picker changes. While we could make use of a popover nonetheless, it doesn't seem particularly more useful to me. Other "workflow" samples use fullscreen modals for setting options before starting a job, like generate offline map w/ overrides. Another option would be to allow the default adaptive presentation style, so it'd be modal on small screens and a popover on large ones.

qsmmorgan: @philium Let me know how you want to proceed on this. I'll change it if you want.

philium:
Though not ideal, I suppose the picker setup is fine the way it is. I would like to see the table view use the `grouped` style and the view controller's modal presentation style be `formSheet`.

qsmmorgan:
@philium Done. Should be good to go.

arcgis-ios-sdk-samples/Analysis/Analyze hotspots/HotspotSettingsViewController.swift
```diff
@@ -28,14 +28,28 @@ class HotspotSettingsViewController: UITableViewController {
...
         super.viewDidLoad()
-
+        updateDatePickerLimits()
+    }
+    
+    /// Set the date picker limits to maintain a valid state.
+    private func updateDatePickerLimits() {
+        let oneDayInSeconds: TimeInterval = 24 * 60 * 60
+        // keep the start date less than the end date
+        startDatePicker.maximumDate = endDatePicker.date - oneDayInSeconds
```
philium: Presumably the operators just call into `Date.addingTimeInterval(_:)`. The documentation for that function states "This only adjusts an absolute value. If you wish to add calendrical concepts like hours, days, months then you must use a Calendar." Could you look into that?

qsmmorgan: Huh, I didn't know about those APIs. Done.

philium: Neither did I!

PR #605, opened 2018-11-26
Standardizes the settings UI of Create and save a map

qsmmorgan:


arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMap.storyboard
```diff
@@ -31,414 +31,279 @@
...
-                                <include reference="Axc-K8-MdF"/>
                             </mask>
                         </variation>
                     </view>
                     <connections>
-                        <outlet property="createOptionsBlurView" destination="KBt-X3-X8f" id="r6G-kL-BPe"/>
                         <outlet property="mapView" destination="ULb-Bf-8qz" id="xZ0-Ip-0MS"/>
-                        <outlet property="saveAsBlurView" destination="ov1-ls-mLv" id="4D4-XW-lGB"/>
-                        <outlet property="savingToolbar" destination="KTS-fi-WLf" id="hif-pS-iyI"/>
+                        <segue destination="lP5-Fn-JsW" kind="presentation" identifier="SaveAsSegue" id="IUH-0f-3gl"/>
```
philium: The presentation style should be `formSheet`:

```suggestion
                        <segue destination="lP5-Fn-JsW" kind="presentation" identifier="SaveAsSegue" modalPresentationStyle="formSheet" id="IUH-0f-3gl"/>
```

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMap.storyboard
```diff
@@ -31,414 +31,279 @@
...
-        <!--Create Options View Controller-->
-        <scene sceneID="jZU-7l-7Je">
+        <!--Create A Map-->
+        <scene sceneID="6Fu-2h-pxB">
             <objects>
-                <viewController id="jve-jS-MPk" customClass="CreateOptionsViewController" customModule="arcgis_ios_sdk_samples" customModuleProvider="target" sceneMemberID="viewController">
-                    <view key="view" contentMode="scaleToFill" id="gik-bE-6tD">
-                        <rect key="frame" x="0.0" y="0.0" width="300" height="350"/>
+                <tableViewController id="Mrh-2f-MIb" customClass="CreateOptionsViewController" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <tableView key="view" clipsSubviews="YES" contentMode="scaleToFill" alwaysBounceVertical="YES" dataMode="prototypes" style="plain" separatorStyle="default" rowHeight="-1" estimatedRowHeight="-1" sectionHeaderHeight="28" sectionFooterHeight="28" id="ZmO-0N-3qB">
```
philium: I would expect this to be a grouped table view.

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMap.storyboard
```diff
@@ -31,414 +31,279 @@
                                     <barButtonItem title="New" id="nTf-dP-Gn9">
                                         <connections>
                                             <action selector="newAction" destination="oSL-y7-4lj" id="nG7-Sz-O2W"/>
+                                            <segue destination="iro-gz-9sV" kind="presentation" id="16f-LT-pfg"/>
```
philium: The presentation style should be `formSheet`:

```suggestion
                                            <segue destination="iro-gz-9sV" kind="presentation" modalPresentationStyle="formSheet" id="16f-LT-pfg"/>
```

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMap.storyboard
```diff
@@ -31,414 +31,279 @@
...
+                                            <fontDescription key="fontDescription" type="system" pointSize="16"/>
+                                            <color key="textColor" red="0.0" green="0.0" blue="0.0" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                            <nil key="highlightedColor"/>
+                                        </label>
+                                    </subviews>
+                                    <color key="backgroundColor" red="0.0" green="0.0" blue="0.0" alpha="0.0" colorSpace="custom" customColorSpace="sRGB"/>
+                                </tableViewCellContentView>
+                                <color key="backgroundColor" red="0.0" green="0.0" blue="0.0" alpha="0.0" colorSpace="custom" customColorSpace="sRGB"/>
+                            </tableViewCell>
+                            <tableViewCell contentMode="scaleToFill" selectionStyle="default" indentationWidth="10" reuseIdentifier="CreateBasemapCell" textLabel="WvO-Vx-8ha" style="IBUITableViewCellStyleDefault" id="lEd-0Q-Sgr">
```
philium: The same prototype cell could be used in both sections, so I would just do that and get rid of one of them.

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -16,51 +16,37 @@ import UIKit
 
 protocol SaveAsVCDelegate: AnyObject {
     func saveAsViewController(_ saveAsViewController: SaveAsViewController, didInitiateSaveWithTitle title: String, tags: [String], itemDescription: String)
-    func saveAsViewControllerDidCancel(_ saveAsViewController: SaveAsViewController)
```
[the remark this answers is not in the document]
qsmmorgan: My belief is that samples should be a simple as possible, especially when it comes to boilerplate features. The callback adds an unnecessary layer of complexity since the sample doesn't require cancel customization.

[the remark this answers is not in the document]
philium: > My belief is that samples should be a simple as possible

Being simple can be good, but being overly simplistic is not so good. A new implementation? Sure, it would be simpler to not create a cancel delegate callback. But in this case the delegate callback was already in place, so it didn't add any value to get rid of it. I would argue that it took away value by removing usage of a best practice.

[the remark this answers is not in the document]
qsmmorgan: I disagree that this is a best practice in this scenario.

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -16,51 +16,37 @@ import UIKit
...
     }
     
     @IBAction private func saveAction() {
-        //Validations
-        guard let title = self.titleTextField.text, let tagsText = self.tagsTextField.text else {
+        
+        guard let title = titleTextField.text,
+            !title.isEmpty else {
             //show error message
-            presentAlert(message: "Title and tags are required fields")
```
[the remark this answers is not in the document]
qsmmorgan: According to the documentation, only the title is required: https://developers.arcgis.com/ios/latest/api-reference/group__licensing.html#ga0bd4d3ba3be09a170c9ca66598cf651a

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
```diff
@@ -35,169 +35,123 @@ extension UIImage {
...
+            //also to cut on the size
+            let croppedImage: UIImage? = image?.croppedImage(CGSize(width: 200, height: 200))
+            
+            self.mapView.map?.save(as: title, portal: self.portal!, tags: tags, folder: nil, itemDescription: itemDescription, thumbnail: croppedImage, forceSaveToSupportedVersion: true) { [weak self] (error) in
+                //dismiss progress hud
+                SVProgressHUD.dismiss()
+                if let error = error {
+                    self?.presentAlert(error: error)
+                } else {
+                    self?.showSuccess()
```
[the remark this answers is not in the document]
qsmmorgan: I implemented this but with `saveAsViewController.dismiss` instead of `self?.dismiss`

qsmmorgan:
@philium I changed it back so that the create UI appears when you open the sample. This should be ready for re-review.

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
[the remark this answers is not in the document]
qsmmorgan: Anyway, if you really want me to change it back I will. Marking this as resolved for now though.

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
```diff
@@ -35,169 +35,124 @@ extension UIImage {
...
+            //crop the image from the center
+            //also to cut on the size
+            let croppedImage: UIImage? = image?.croppedImage(CGSize(width: 200, height: 200))
+            
+            self.mapView.map?.save(as: title, portal: self.portal!, tags: tags, folder: nil, itemDescription: itemDescription, thumbnail: croppedImage, forceSaveToSupportedVersion: true) { [weak self] (error) in
+                //dismiss progress hud
+                SVProgressHUD.dismiss()
+                if let error = error {
+                    saveAsViewController.presentAlert(error: error)
+                    saveAsViewController.dismiss(animated: true)
```
philium: I'm confused by this. Why is the view controller being dismissed if it just presented an alert?

qsmmorgan: Good point. I updated it so it's dismissed first and then the parent view controller presents the alert.

philium: > I updated it so it's dismissed first

I kind of like how the Save As view stayed up when there was an error. It made it easier to try again. That being said, it isn't a big deal.

Since both branches now dismiss the Save As view controller, you could combine the two calls to `dismiss` by moving the `if` statement into the completion handler:

```swift
saveAsViewController.dismiss(animated: true) {
    if let error = error {
        self?.presentAlert(error: error)
    } else {
        self?.showSuccess()
    }
}
```

qsmmorgan: > I kind of like how the Save As view stayed up when there was an error. It made it easier to try again.

Ohh I didn't realize this was desired behavior. Your code from before makes more sense now.

qsmmorgan: Okay, this should be good to go.

qsmmorgan:
@philium Thanks for reviewing!

PR #606, opened 2018-11-27
Converts the Route Around Barriers settings to a table view

qsmmorgan:


arcgis-ios-sdk-samples/Route & Directions/Route around barriers/RouteParametersViewController.swift
```diff
@@ -16,44 +16,53 @@
...
-            self.routeParameters.findBestSequence = sender.isOn
-            self.enableSubSwitches(sender.isOn)
-        } else if sender == self.preservceFirstStopSwitch {
-            self.routeParameters.preserveFirstStop = self.preservceFirstStopSwitch.isOn
-        } else {
-            self.routeParameters.preserveLastStop = self.preservceLastStopSwitch.isOn
+        switch sender {
+        case findBestSequenceSwitch:
+            routeParameters?.findBestSequence = sender.isOn
+            setupUI()
```
philium: Something seems to be off. When both `preserveFirstStop` and `preserveLastStop` are `false`, toggling `findBestSequenceSwitch` toggles both of the other switches. When only one is false, that one stays false when `findBestSequenceSwitch` is toggled. That seems to be a behavior change. Was that intentional?

qsmmorgan:
@philium I converted it to hide/show the secondary controls in a second section like the demo you sent. Tell me what you think.

arcgis-ios-sdk-samples/Route & Directions/Route around barriers/RouteParametersViewController.swift
```diff
@@ -16,44 +16,53 @@
...
+        case preserveFirstStopSwitch:
+            routeParameters?.preserveFirstStop = preserveFirstStopSwitch?.isOn == true
```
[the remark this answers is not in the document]
philium: Oops. This is an out-of-date from before I realized that you could just use the sender to get the `isOn` state.

arcgis-ios-sdk-samples/Route & Directions/Route around barriers/RouteParametersViewController.swift
```diff
@@ -56,13 +49,26 @@ class RouteParametersViewController: UITableViewController {
...
         default:
             break
         }
     }
+    
+    override func numberOfSections(in tableView: UITableView) -> Int {
+        if routeParameters?.findBestSequence == false {
+            return 1
+        }
+        return super.numberOfSections(in: tableView)
```
philium: I would suggest always getting `super`'s value and then subtracting off the section you don't want. Something like this:

```swift
override func numberOfSections(in tableView: UITableView) -> Int {
    let superCount = super.numberOfSections(in: tableView)
    if routeParameters?.findBestSequence ?? false {
        return superCount - 1
    } else {
        return superCount
    }
}
```

arcgis-ios-sdk-samples/Route & Directions/Route around barriers/RouteParametersViewController.swift
```diff
@@ -56,13 +49,26 @@ class RouteParametersViewController: UITableViewController {
         switch sender {
         case findBestSequenceSwitch:
             routeParameters?.findBestSequence = sender.isOn
+            if sender.isOn {
```
philium: This looks good. One thing I would suggest is to adjust the size of the popover to show the new sections. Here's one way of doing that:

```swift
let sections: IndexSet = [1]
tableView.performBatchUpdates({
    if sender.isOn {
        tableView?.insertSections(sections, with: .fade)
    } else {
        tableView?.deleteSections(sections, with: .fade)
    }
}, completion: { (finished) in
    guard finished else { return }
    self.preferredContentSize.height = self.tableView.contentSize.height
})
```

And it looks something like this:

screen recording 2018-11-30 at 9 59 41 am 2018-11-30 10_00_29

philium:
Did you want to update the initial height of the popover? It currently has a different (larger) initial height then the height it animates to when the second section is removed. It also doesn't have the right height when Find Best Sequence is On. One way to fix that is to reload the table view and update the preferred content size before the popover appears:

```swift
    tableView.reloadData()
    preferredContentSize.height = tableView.contentSize.height
}
```

PR #607, opened 2018-11-27
Converts the text field area of Format Coordinates to a table

qsmmorgan:


arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
```diff
@@ -178,16 +47,182 @@
...
                 <placeholder placeholderIdentifier="IBFirstResponder" id="Wlm-1W-Txj" userLabel="First Responder" sceneMemberID="firstResponder"/>
             </objects>
-            <point key="canvasLocation" x="713" y="199"/>
+            <point key="canvasLocation" x="712.79999999999995" y="198.35082458770617"/>
+        </scene>
+        <!--Format Coordinates Table View Controller-->
+        <scene sceneID="eIn-QB-OfK">
+            <objects>
+                <tableViewController id="geL-f0-eoD" customClass="FormatCoordinatesTableViewController" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <tableView key="view" clipsSubviews="YES" contentMode="scaleToFill" alwaysBounceVertical="YES" keyboardDismissMode="onDrag" dataMode="static" style="plain" separatorStyle="default" allowsSelection="NO" rowHeight="-1" estimatedRowHeight="-1" sectionHeaderHeight="28" sectionFooterHeight="28" id="5vY-pq-BWx">
```
philium: Consider using a grouped table view. Plain table views with static content (ie. fixed number of rows) generally look odd because of the extra blank rows at the bottom.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
```diff
@@ -1,168 +1,37 @@
...
-                                            <constraint firstItem="f0e-3N-1ay" firstAttribute="width" secondItem="13U-3h-0zt" secondAttribute="width" id="EYC-tF-SQB"/>
-                                            <constraint firstAttribute="trailing" secondItem="f0e-3N-1ay" secondAttribute="trailing" id="OaD-rE-WNQ"/>
-                                            <constraint firstAttribute="bottom" secondItem="f0e-3N-1ay" secondAttribute="bottom" id="Uuc-kp-abY"/>
-                                            <constraint firstAttribute="height" constant="151" id="Z4H-mT-Fo4"/>
-                                            <constraint firstAttribute="height" constant="233" id="er3-ni-T2e"/>
-                                            <constraint firstAttribute="width" constant="250" id="sm8-7f-Fjg"/>
-                                            <constraint firstItem="f0e-3N-1ay" firstAttribute="top" secondItem="13U-3h-0zt" secondAttribute="top" id="ysb-Np-CVo"/>
+                                            <constraint firstAttribute="height" constant="300" id="NX2-ML-r3F"/>
```
philium: This constraint should be removed. It generates an unsatisfiable constraints warning in compact height environments and wouldn't be needed if the distribution is changed.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
```diff
@@ -1,168 +1,37 @@
...
         <!--Format Coordinates View Controller-->
         <scene sceneID="Tpr-Zl-3jf">
             <objects>
-                <viewController automaticallyAdjustsScrollViewInsets="NO" id="IIZ-g5-rn0" customClass="FormatCoordinatesViewController" customModule="arcgis_ios_sdk_samples" customModuleProvider="target" sceneMemberID="viewController">
+                <viewController automaticallyAdjustsScrollViewInsets="NO" id="IIZ-g5-rn0" customClass="FormatCoordinatesViewController" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target" sceneMemberID="viewController">
                     <view key="view" contentMode="scaleToFill" id="nRX-gd-Kdt">
                         <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
                             <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" translatesAutoresizingMaskIntoConstraints="NO" id="GzU-Me-xVd">
```
philium: I would suggest inverting the order of the subviews. It would make the whitespace at the bottom of the table view in regular height environments less noticeable.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
```diff
@@ -1,168 +1,37 @@
...
+                                        <connections>
+                                            <segue destination="geL-f0-eoD" kind="embed" id="fDz-l6-wCh"/>
+                                        </connections>
+                                    </containerView>
                                     <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="vLE-Mm-bqf" customClass="AGSMapView">
-                                        <rect key="frame" x="0.0" y="233" width="375" height="414"/>
+                                        <rect key="frame" x="0.0" y="300" width="375" height="347"/>
                                     </view>
                                 </subviews>
                                 <variation key="heightClass=compact" axis="horizontal"/>
```
philium: I would suggest using regular width, compact height for the variant. That way the stack view uses a horizontal distribution on iPad, which I think is a bit more natural.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
philium: The stack view's leading and top constraints should be updated to use the superview rather than the safe area.

screen shot 2018-12-03 at 10 41 10 am

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
```diff
@@ -178,16 +47,182 @@
...
+                                                    <nil key="highlightedColor"/>
+                                                </label>
+                                            </subviews>
+                                            <constraints>
+                                                <constraint firstItem="5aq-8r-2lB" firstAttribute="leading" secondItem="297-yI-a6o" secondAttribute="leadingMargin" id="NRZ-fA-Vrf"/>
+                                                <constraint firstItem="Flo-zm-KFZ" firstAttribute="top" secondItem="5aq-8r-2lB" secondAttribute="bottom" constant="4" id="ija-gt-iDX"/>
+                                                <constraint firstItem="Flo-zm-KFZ" firstAttribute="trailing" secondItem="5aq-8r-2lB" secondAttribute="trailing" id="kVf-8F-EHK"/>
+                                                <constraint firstItem="Flo-zm-KFZ" firstAttribute="leading" secondItem="5aq-8r-2lB" secondAttribute="leading" id="qRW-it-UHG"/>
+                                                <constraint firstAttribute="trailingMargin" secondItem="5aq-8r-2lB" secondAttribute="trailing" id="vBM-bl-Cat"/>
```
philium: The trailing constraint of each label should be `greaterThanOrEqual` so that they are only as wide as they need to be.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
```diff
@@ -178,16 +47,182 @@
...
+                                        <rect key="frame" x="0.0" y="0.0" width="375" height="44"/>
+                                        <autoresizingMask key="autoresizingMask"/>
+                                        <tableViewCellContentView key="contentView" opaque="NO" clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="center" insetsLayoutMarginsFromSafeArea="NO" tableViewCell="uS2-Xt-y5G" id="297-yI-a6o">
+                                            <rect key="frame" x="0.0" y="0.0" width="375" height="43.5"/>
+                                            <autoresizingMask key="autoresizingMask"/>
+                                            <subviews>
+                                                <textField opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="left" contentVerticalAlignment="center" placeholder="45.0000N 120.0000E" textAlignment="natural" adjustsFontForContentSizeCategory="YES" minimumFontSize="17" translatesAutoresizingMaskIntoConstraints="NO" id="Flo-zm-KFZ">
```
philium: It would be nice to put in some sample values so that the text fields are actually viewable on the canvas.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
```diff
@@ -178,16 +47,182 @@
...
+                                                    <fontDescription key="fontDescription" style="UICTFontTextStyleSubhead"/>
+                                                    <color key="textColor" white="0.33333333333333331" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                                                    <nil key="highlightedColor"/>
+                                                </label>
+                                            </subviews>
+                                            <constraints>
+                                                <constraint firstItem="5aq-8r-2lB" firstAttribute="leading" secondItem="297-yI-a6o" secondAttribute="leadingMargin" id="NRZ-fA-Vrf"/>
+                                                <constraint firstItem="Flo-zm-KFZ" firstAttribute="top" secondItem="5aq-8r-2lB" secondAttribute="bottom" constant="4" id="ija-gt-iDX"/>
+                                                <constraint firstItem="Flo-zm-KFZ" firstAttribute="trailing" secondItem="5aq-8r-2lB" secondAttribute="trailing" id="kVf-8F-EHK"/>
```
philium: The trailing constraints of each text field should be equal to the trailing margin of the super superview.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
qsmmorgan: I don't know what the variant even is.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
qsmmorgan: This doesn't affect the UI but okay.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
qsmmorgan: Hmm it'd be nice if they displayed. It doesn't look like adding text values helps. It's all about the compression resistance priority I think.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
qsmmorgan: The issue with switching the order is that the text field area appears squished when the software keyboard is onscreen.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
philium: It doesn't affect the UI, but it does affect performance. Enable Color Blended Layers in the simulator to see how.

Before: 
screen shot 2018-12-03 at 3 40 48 pm

After: 
screen shot 2018-12-03 at 3 41 50 pm

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
philium: A variation allows setting a different value for a property in a given size class. This is what it looks like:

screen shot 2018-12-03 at 3 54 41 pm

And that's the combination I would recommend. In my experience, it is best to use side-by-side everywhere except compact width, regular height environments.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
philium: Huh. I don't really know why it has a height of zero. It already has a placeholder, which should give it an intrinsic height.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
philium: I see what you mean. That makes sense.

arcgis-ios-sdk-samples/Geometry/Format coordinates/FormatCoordinates.storyboard
qsmmorgan: Thanks for that explanation. I updated it to your recommendation. This PR should be good to go.

dg0yal:
@qsmmorgan @philium With this change did we lose the ability to enter some coordinate text and see the corresponding location on the map?

qsmmorgan:
@dg0yal Nope, you can still type coordinates by tapping on the text boxes, they just look a little less like text boxes.

PR #609, opened 2018-11-30
Replaces ExpandableTableViewController in statistical query group and sort

qsmmorgan:


philium:
Adding a row to a non-topmost view when tapping a plus button on the topmost view seems like a highly unusual workflow. First, you have to tap two difference plus buttons. Second, usually the modal view that configures the item to be added is dismissed when done with configuration. What would you think of turning the Statistics Definition plus button into a done button?

Also, since the number of rows in the Statistics Definition view is fixed, it would be best to use a grouped table view.

qsmmorgan:
@philium Yeah, the UI for adding a definition has been wonky since the start and could use some work. I'm pretty sure it's outside the scope of the changes in this PR though. These changes only affect the display of the query results, not the inputs. Correct me if I'm wrong about that.

philium:
> These changes only affect the display of the query results

There definitely is a behavior difference between what was in 100.4 and this version. Before, tapping the plus button would add new rows to the Static Definitions view and tapping the Done button would dismiss it. Now, The Done button has been replaced by the plus button and tapping it adds new rows to the Statistical Query (Group & Sort) view.

qsmmorgan:
Oh it's definitely different than 100.4, but the change was made in an intervening PR, not this one. Here is a screenshot from the current `v.next`. 
simulator screen shot - iphone xr - 2018-12-06 at 14 59 03

qsmmorgan:
I've been removing the Done buttons from the popovers since they're redundant to just tapping outside the popover.

qsmmorgan:
Here's the PR that made the change: https://github.com/Esri/arcgis-runtime-samples-ios/pull/542

philium:
> they're redundant to just tapping outside the popover

That's only true if all the done button does is dismiss the popover. In the case of this view, the done button did more than that. Oddly enough, tapping outside the popover on 100.4 doesn't dismiss the popover. But if it did, that would be a different behavior, more like cancel. Regardless, the PR you linked to didn't mention anything about changing the behavior of the plus button or replacing the done button with it. Behavior changes like that should be called out in the PR description.

qsmmorgan:
@philium Okay, well I'll keep that in mind for the future!

arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
```diff
@@ -372,3 +362,74 @@ class StatisticalQueryGroupAndSortViewController: UIViewController, UITableViewD
...
+        didSet {
+            statisticRecords = statisticsQueryResult?.statisticRecordEnumerator().allObjects ?? []
+        }
+    }
+    
+    var selectedGroupByFieldNames: [String] = []
+    
+    var statisticRecords: [AGSStatisticRecord] = [] {
+        didSet {
+            tableView?.reloadData()
```
philium: Querying the `tableView` property will cause the view to load. I would suggest guarding `isViewLoaded` before this call. Same with line `419`.

qsmmorgan:
@philium Done.

PR #611, opened 2018-11-30
Limits the statistic definitions to numeric fields in statistical query group and sort

qsmmorgan:


arcgis-ios-sdk-samples/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
```diff
@@ -64,12 +65,21 @@ class StatisticalQueryGroupAndSortViewController: UIViewController, UITableViewD
...
-            self.fieldNames = serviceFeatureTable.fields.compactMap({ (field) -> String? in
+            self.fieldNames = serviceFeatureTable.fields.compactMap { (field) -> String? in
                 if field.type != .OID && field.type != .globalID {
                     return field.name
                 }
                 return nil
-            })
+            }
+            self.numericFieldNames = serviceFeatureTable.fields.compactMap { (field) -> String? in
+                if field.type == .double ||
```
philium: Consider combining these checks into a single operation. One way of doing that is using a helper set:

```swift
let numericTypes: Set<AGSFieldType> = [.double, .float, .int32, .int16]
self.numericFieldNames = serviceFeatureTable.fields.compactMap { numericTypes.contains($0.type) ? $0.name : nil }
```

If it were me, I would take it a step further:

```swift
extension AGSField {
    var isNumeric: Bool {
        let numericTypes: Set<AGSFieldType> = [.double, .float, .int32, .int16]
        return numericTypes.contains(type)
    }
}

self.numericFieldNames = serviceFeatureTable.fields.compactMap { $0.isNumeric ? $0.name : nil }
```

PR #614, opened 2018-12-06
Added the feature layer rendering mode (scene) sample

qsmmorgan:


qsmmorgan:
I was basing this on the [Java implementation](https://github.com/Esri/arcgis-runtime-samples-java/tree/dev/src/main/java/com/esri/samples/scene/feature_layer_rendering_mode_scene), which appears to animate automatically. I'll add a button.

qsmmorgan:
@philium I updated it to use a button.

PR #617, opened 2018-12-10
Converts the Manage sublayers settings to a popover

qsmmorgan:


philium:
I'm not sure if the view makes sense as a popover. Not all of the rows are visible at a time, so most of the time you can't see where an added or removed layer went. And using the table editing controls inside a popover feels a little clunky. I just reread the HIG regarding popovers and a few things stood out:

> Avoid making a popover too big.

The popover does feel big. It takes up half of the screen and the navigation bar at the top feels like waisted space.

> Avoid displaying popovers on iPhones.

That one surprised me a bit. When Apple started allowing popovers on iPhone, I just started using them because they were there and stopped thinking about whether they are a good solution. I do like how popovers keep the user in the current context, but the smaller the better.

I would suggest keeping the view as a modal view, moving the done button to the right, and adding a cancel button on the left. I would also suggest updating the removed rows to use the standard add control.

qsmmorgan:
I see your points. My main thought is that it's valuable to see the layers changing live as you manage them, particularly since that is the focus of the sample. In my mind that takes precedence over the size considerations. If you are sure you want to keep it a modal, that's fine too. Let me know.

As far as avoiding popovers on iPhones, I feel like that part of the HIG must not have been updated in the past few years. It makes sense for SE sizes and smaller, but with Plus and X sizes you wouldn't want to make everything a modal.

qsmmorgan:
@philium Have you decided about this?

philium:
I've given it some more thought and I still think it makes more sense as a modal view rather than a popover.

qsmmorgan:
@philium I made the modal change so this should be good to go. I took the liberty of converting the settings to a table view controller while I was at it. I didn't add a cancel button because that would be a heavy lift with managing state, and I don't think it's necessary since this isn't a workflow.

arcgis-ios-sdk-samples/Layers/Manage sublayers/ManageSublayers.storyboard
```diff
@@ -29,9 +29,9 @@
                                 </constraints>
                                 <items>
                                     <barButtonItem style="plain" systemItem="flexibleSpace" id="J9K-Vc-wO5"/>
-                                    <barButtonItem title="Sublayers" id="CmR-Ri-JIK">
+                                    <barButtonItem title="Manage Sublayers" id="CmR-Ri-JIK">
                                         <connections>
-                                            <segue destination="cg1-iZ-LA9" kind="presentation" identifier="MapImageSublayersSegue" modalPresentationStyle="formSheet" id="MXS-1S-AVQ"/>
+                                            <segue destination="cg1-iZ-LA9" kind="presentation" identifier="MapImageSublayersSegue" id="MXS-1S-AVQ"/>
```
philium: Please keep the modal presentation style as `formSheet`. That way the view is not full screen in a regular width environment (such as on an iPad).

arcgis-ios-sdk-samples/Layers/Manage sublayers/ManageSublayers.storyboard
```diff
@@ -60,72 +60,57 @@
...
         </scene>
         <!--Sublayers-->
-        <scene sceneID="CRF-ji-nlo">
+        <scene sceneID="hgS-yY-ZCr">
             <objects>
-                <viewController automaticallyAdjustsScrollViewInsets="NO" id="JeC-e0-SwS" customClass="MapImageSublayersVC" customModule="arcgis_ios_sdk_samples" customModuleProvider="target" sceneMemberID="viewController">
-                    <view key="view" contentMode="scaleToFill" id="ank-tO-olW">
-                        <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
+                <tableViewController id="Cix-X6-4BY" customClass="MapImageSublayersVC" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <tableView key="view" clipsSubviews="YES" contentMode="scaleToFill" alwaysBounceVertical="YES" dataMode="prototypes" style="plain" separatorStyle="default" allowsSelection="NO" rowHeight="-1" estimatedRowHeight="-1" sectionHeaderHeight="28" sectionFooterHeight="28" id="7uH-wp-jAp">
```
philium: Thank you for converting this to a table view controller. Please use the grouped table view style.

arcgis-ios-sdk-samples/Layers/Manage sublayers/ManageSublayers.storyboard
qsmmorgan: Ah I was wondering what the difference was.

qsmmorgan:
@philium Done.

PR #619, opened 2018-12-11
Fixes warning for SwiftLint v0.29.1

qsmmorgan:
https://github.com/realm/SwiftLint/releases/tag/0.29.1

The warnings didn't appear to come from new default rules, but rather rules we had already implemented that the old version of SwiftLint has missed.

I just went ahead and removed `closure_body_length` since there doesn't seem to be a great way to conform to that rule and it's not critical to enforcing a uniform style, which is our goal with linting.

### DIFF
--- a/arcgis-ios-sdk-samples/Scenes/Extrude graphics/ExtrudeGraphicsViewController.swift
+++ b/arcgis-ios-sdk-samples/Scenes/Extrude graphics/ExtrudeGraphicsViewController.swift
@@ -99,7 +99,7 @@ class ExtrudeGraphicsViewController: UIViewController {
     //add a graphic to the graphics overlay for the given polygon
     private func addGraphicForPolygon(_ polygon: AGSPolygon) {
         
-        let rand = arc4random_uniform(UInt32(self.maxHeight))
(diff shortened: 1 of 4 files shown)


arcgis-ios-sdk-samples/Scenes/Extrude graphics/ExtrudeGraphicsViewController.swift
```diff
@@ -99,7 +99,7 @@ class ExtrudeGraphicsViewController: UIViewController {
     //add a graphic to the graphics overlay for the given polygon
     private func addGraphicForPolygon(_ polygon: AGSPolygon) {
         
-        let rand = arc4random_uniform(UInt32(self.maxHeight))
+        let rand = Int.random(in: 0...maxHeight)
```
philium: It might be worth pointing out that `maxHeight` is now a possible random number whereas before it wasn't. I don't see a problem with that though.

qsmmorgan: Nice catch. I think it's an extremely minor behavior change so I'm not worried about it.

PR #620, opened 2018-12-11
Adds the Buffer sample

qsmmorgan:


arcgis-ios-sdk-samples/Geometry/Buffer/BufferViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+
+extension BufferViewController: AGSGeoViewTouchDelegate {
+    
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        
+        /// The buffer radius converted to meters.
+        let bufferRadiusInMeters = bufferDistance.converted(to: .meters).value
+       
+        // ensure that the buffer radius is a positive value
+        guard bufferRadiusInMeters > 0 else {
```
philium: Were you seeing a negative measurement? I'm not sure how that would be possible.

arcgis-ios-sdk-samples/Geometry/Buffer/README.md
```diff
@@ -0,0 +1,26 @@
...
+2. Graphics for a planar buffer (brown) and geodesic buffer (green) are added for the tap location using the distance specified in the text box.
+3. Continue tapping to create additional buffers. Notice that buffers closer to the equator appear similar in size. As you move north or south from the equator, however, the geodesic polygons become much larger. Geodesic polygons are in fact a better representation of the true shape and size of the buffer.   
+4. Tap `Clear All` to remove all the graphics and start again.
+
+## How it works
+The map point for a tap on the display is captured and the buffer distance entered (in miles) is converted to meters. The `AGSGeometryEngine` static method `bufferGeometry(_:byDistance:)` is called to create a planar buffer polygon from the map location and distance. Another static method, `geodeticBufferGeometry(_:distance:distanceUnit:maxDeviation:curveType:)` is called to create a geodesic buffer polygon using the same inputs. The polygon results (and tap location) are displayed in the map view with different symbols in order to highlight the difference between the buffer techniques due to the spatial reference used in the planar calculation.
```
philium: > the buffer distance entered (in miles) is converted to meters

I'm wondering if this should be reworded. Specified might be better than entered--entered makes me think of tapping keys on a keyboard. Also the specified distance is not always in miles.

arcgis-ios-sdk-samples/Geometry/Buffer/BufferViewController.swift
qsmmorgan: It shouldn't be, but I figure this is a fine sanity check to show developers the value should always be positive.

arcgis-ios-sdk-samples/Geometry/Buffer/BufferOptionsViewController.swift
```diff
@@ -0,0 +1,61 @@
...
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        updateUIForBufferRadius()
+    }
+    
+    private func updateUIForBufferRadius() {
+        // update the slider and label to match the buffer distance
+        distanceSlider?.value = Float(bufferDistance.converted(to: .miles).value)
+        distanceLabel?.text = measurementFormatter.string(from: bufferDistance)
```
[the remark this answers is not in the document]
qsmmorgan: Yeah I considered that but fixing it would make more of a mess than it's worth.

qsmmorgan:
@philium I made the changes. This is ready for re-review.

PR #628, opened 2018-12-19
Improves GeocodeOfflineViewController.graphicForPoint(_:attributes:)

philium:
- Renames function to `makeGraphic(point:attributes:)`
- Changes type of `attributes` parameter to match `AGSGraphic.init(geometry:symbol:attributes:)`
- Provides default value of `attributes parameter`
- Uses image literal


philium:
@gdhakal Thank you!

PR #632, opened 2018-12-19
Removes objc attribute from DownloadProgressViewDelegate

philium:
It wasn't needed, so it just cause the compiled code to be larger than needed.

Also changed the header to the standard one.


philium:
@gdhakal Thank you!

PR #634, opened 2018-12-19
Changes import from UIKit to Foundation

philium:
`SampleSearchEngine` doesn't depend on anything from `UIKit`.


philium:
Thank you!

PR #635, opened 2018-12-20
Remove SearchForWebmapByKeywordViewController.searchController

philium:
The property is no longer needed.


philium:
@sbaskaran Thank you!

PR #636, opened 2018-12-26
Route Around Barriers: Prevent animation on appearance

philium:
The Route Around Barriers view controller was animating the layout of its view while the view was appearing on screen. This ensures that the view is fully laid out before appearing.


arcgis-ios-sdk-samples/Route & Directions/Route around barriers/RouteAroundBarriersViewController.swift
```diff
@@ -67,8 +67,15 @@ class RouteAroundBarriersViewController: UIViewController, AGSGeoViewTouchDelega
         //get default parameters
         self.getDefaultParameters()
         
+//        //hide directions list
+//        self.setRouteDetailsVisibility(visible: false)
+    }
```
gdhakal: The code is commented. Should it be removed?

philium: Good catch! Yes, I'll remove it.

philium:
@gdhakal Thank you!

PR #637, opened 2018-12-27
Add Calculate 3D Distance sample

philium:


arcgis-ios-sdk-samples/Scenes/Calculate 3D distance/Calculate3DDistanceViewController.swift
```diff
@@ -0,0 +1,283 @@
...
+/// A point in a two-dimensional coordinate system.
+struct Point: Equatable {
+    /// The x-coordinate of the point.
+    var x: Double
+    /// The y-coordinate of the point.
+    var y: Double
+}
+
+/// A function of a single variable.
+protocol Function {
```
pgruenler: Instead of just `Function` what about `PositionFunction` to make it a bit clearer what the function does.

arcgis-ios-sdk-samples/Scenes/Calculate 3D distance/Calculate3DDistanceViewController.swift
```diff
@@ -0,0 +1,283 @@
...
+    /// Applies the function to a given value.
+    ///
+    /// - Parameter x: A value.
+    /// - Returns: The result of applying the function to `x`.
+    func apply(to x: Double) -> Double
```
pgruenler: Here I'd suggest `func position(for x: Double) -> Double` to make things a bit more specific as to what is going on when it's used.

arcgis-ios-sdk-samples/Scenes/Calculate 3D distance/Calculate3DDistanceViewController.swift
```diff
@@ -0,0 +1,283 @@
...
+}
+
+/// An absolute value function of the form `y=a|x-h|+k`. The vertex of the
+/// function is at `(h, k)`.
+struct AbsoluteValueFunction: Function {
```
pgruenler: I don't really have a suggestion here but I wouldn't mind if it was named more for what it does in terms of calculating position, rather than how it does it.  That being said, this name gets hidden when the assignments to `redPositionFunction` and `greenPositionFunction` are made so I'd be okay if it didn't change.

arcgis-ios-sdk-samples/Scenes/Calculate 3D distance/Calculate3DDistanceViewController.swift
philium: Interesting suggestion. While the absolute value function is being used to calculate a position in this sample, a `Function` type isn't sample specific. It is possible that other samples could use `AbsoluteValueFunction` or declare their own function type to calculate something other than a position.

It might actually make more sense to move the `Function` code out of the sample since it isn't ArcGIS related. That would simplify the sample view controller while emphasizing that these `Function` components are reusable. Then if `PositionFunction` would be clearer, it can be declared as a private typedef to `AbsoluteValueFunction` inside the sample view controller source file.

pgruenler: Moving things into other files and adding reusable components would go against some design ideas we've [discussed before](https://github.com/Esri/arcgis-runtime-samples-ios/pull/464).

As far as names, are you happy with calling `apply(to x: ...)` to get the next position?

philium: My understanding is that it is not desirable to make `ArcGIS`-related components reusable. We do have `ArcGIS`-unrelated reusable components in the sample app, such as [the methods for presenting an alert](https://github.com/Esri/arcgis-runtime-samples-ios/pull/516). Is my understanding incorrect? If so, please correct me.

I do like the name `apply(to:)` since it mirrors what a [mathematical function](https://en.wikipedia.org/wiki/Function_(mathematics)) actually does. We could rename it to `position(for:)`, but that seems a little redundant, assuming the function is named accordingly. My thought is that the names `redPositionFunction` and `greenPositionFunction` imply that the output is a position. What do you think?

pgruenler: @dg0yal, I'll defer to you.  Do the names look okay?

philium:
We have decided not to implement this sample at this time.

PR #640, opened 2018-12-31
Refactor Change Sublayer Renderer view controller

philium:
This started out as an effort to rename `weakSelf` (which was in fact strong) to `self` (optionally bound). One change led to another and I ended up refactoring parts of the sample. The biggest change is the addition of a property for the counties sublayer.

### DIFF
--- a/arcgis-ios-sdk-samples/Layers/Change sublayer renderer/ChangeSublayerRendererVC.swift
+++ b/arcgis-ios-sdk-samples/Layers/Change sublayer renderer/ChangeSublayerRendererVC.swift
@@ -22,7 +22,9 @@ class ChangeSublayerRendererVC: UIViewController {
     @IBOutlet private var applyRendererBarButtonItem: UIBarButtonItem!
     
     //map image layer
-    private var mapImageLayer = AGSArcGISMapImageLayer(url: URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer")!)
+    private let mapImageLayer = AGSArcGISMapImageLayer(url: URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer")!)
+    /// The counties sublayer of the map image layer.
+    private var countiesSublayer: AGSArcGISMapImageSublayer?
     private var originalRenderer: AGSRenderer?
     
     override func viewDidLoad() {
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Layers/Change sublayer renderer/ChangeSublayerRendererVC.swift
```diff
@@ -48,33 +50,49 @@ class ChangeSublayerRendererVC: UIViewController {
         
         //load map image layer to access sublayers
         self.mapImageLayer.load { [weak self] (error: Error?) in
-            
-            guard let weakSelf = self else {
-                return
+            guard let self = self else { return }
+            if let error = error {
+                print("Error loading map image layer: \(error)")
```
gdhakal: I think we should show the error to user using `SVProgressHUD`.

arcgis-ios-sdk-samples/Layers/Change sublayer renderer/ChangeSublayerRendererVC.swift
```diff
@@ -48,33 +50,49 @@ class ChangeSublayerRendererVC: UIViewController {
...
+        guard mapImageSublayers.count >= 3,
+            let sublayer = mapImageSublayers[2] as? AGSArcGISMapImageSublayer else {
+                return
+        }
+        countiesSublayer = sublayer
+        //load the sublayer to get the original renderer
+        sublayer.load { [weak self] (error) in
+            guard let self = self else { return }
+            if let error = error {
+                print("Error loading map image sublayer: \(error)")
```
gdhakal: Should show error to user?

philium: That's a good question. We don't use `SVProgressHUD` to show errors anymore. We do use `UIAlertController` to show some errors, but usually only errors that the user can do something about. In the event of a load error, the user can't really do anything about it (other than close and reopen the sample). And since a load error is unlikely to occur, it is simpler to just log those errors rather than try to show them to the user.

gdhakal: ok, thanks for clarification.

philium:
@gdhakal Thank you!

PR #641, opened 2018-12-31
Renames weakSelf to self in EditFeaturesOnlineViewController

philium:


philium:
@sbaskaran Thank you!

PR #643, opened 2019-01-03
Add Graphics Overlay (Dictionary Renderer) 3D sample

philium:


arcgis-ios-sdk-samples/Display information/Graphics overlay (dictionary renderer) 3D/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## How it works
+
+To apply an `AGSDictionaryRenderer` and display mil2525d graphics in 3D:
+
+1. Create an `AGSDictionarySymbolStyle` object with the "mil2525d" specification type.
+2. Create an `AGSDictionaryRenderer` object with the dictionary symbol style.
+3. Create an instance of `AGSGraphicsOverlay`.
+4. Set the dictionary renderer as the renderer of the graphics overlay.
+5. Parse an XML file conforming to the specification. It should have key-value pairs to use as attributes for each graphic:
```
rolson: I don't know much about this api - but it seems like we should have something in the api that can parse these and create graphics for the user. Are we sure that doesn't exist in our api?

arcgis-ios-sdk-samples/Display information/Graphics overlay (dictionary renderer) 3D/GraphicsOverlayDictionaryRenderer3DViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+    
+    private var controlPoints = [ControlPoint]()
+    private var wkid: Int?
+    private var attributes = [String: Any]()
+    private var contentsOfCurrentElement = ""
+    
+    private var parsedMessages = [Mil2525DMessage]()
+    
+    func didFinishParsingMessage() {
+        let spatialReference = AGSSpatialReference(wkid: wkid!)
```
rolson: dangerous?

arcgis-ios-sdk-samples/Display information/Graphics overlay (dictionary renderer) 3D/GraphicsOverlayDictionaryRenderer3DViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+        case wkid = "_wkid"
+    }
+    
+    func parser(_ parser: XMLParser, didEndElement elementName: String, namespaceURI: String?, qualifiedName qName: String?) {
+        if let element = Element(rawValue: elementName) {
+            switch element {
+            case .controlPoints:
+                controlPoints = contentsOfCurrentElement.split(separator: ";").map { (pair) in
+                    let coordinates = pair.split(separator: ",")
+                    return ControlPoint(x: Double(coordinates.first!)!, y: Double(coordinates.last!)!)
```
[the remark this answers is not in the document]
philium: It does at first glance, but I think it is safe for two reasons:
1. We have control over the input. The only file parsed is included in the app bundle, so we have complete control over it.
2. We have a unit test that exercises this code on that file. Any changes to the file that break this code would be caught when that test is run.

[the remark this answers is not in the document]
rolson: The problem is that even when that happens and the unit test breaks, you fix the test. You don't remember that you also need to fix the code. Personally, I think I would avoid forced unwrapping here and the other place.

arcgis-ios-sdk-samples/Display information/Graphics overlay (dictionary renderer) 3D/README.md
philium: I'm fairly certain, but I'll double check. My understanding is that the XML schema is dependent on the specification, so it wouldn't be possible to create a single parser that works for all specifications.

rolson:
Overall - the PR seems ok, just IMO we shouldn't be force unwrapping in the two pointed out locations.

Also - I have asked Ralf G to chime in because the xml parsing in this sample is confusing.

ralfgottschalk:
@rolson asked me to take a look at this.  The sample is fine but there are two parts that I think might be misleading to the customer.  

>The dictionary renderer creates these graphics using the mil2525d style file plus an XML file with key-value attributes for each graphic.

This is not actually true.  Dictionary Renderers are renderers that use a list of attributes to determine what symbols to fetch from an associated style file and assemble the symbols into a multilayer symbol.  XML is not necessary. 

You could just say "The dictionary renderer uses the attributes of the graphics or features to assemble symbols from the Mil2525D style file." Or something like that.

I would also probably change the class name of the `Mil2525DMessageParser`.  To my knowledge there is no spec in Mil2525D for xml.  Some agencies transfer messages around but there is no standard format for this.  This would just be an example of parsing a message.  Calling it Mil2525D is a little misleading because its just some data in xml.

philium:
@ralfgottschalk Thank you for your feedback. Is there a standard schema for military messages?

ralfgottschalk:
> @ralfgottschalk Thank you for your feedback. Is there a standard schema for military messages?

Not that I know of.

rolson:
Maybe part of the problem is the naming `Mil2525DMessageParser`. It implies an adherence to some mil2525d spec.

> If there is a better way to do this in iOS, I'm ok with that (perhaps using JSON or some other data format)

Well - that is up to you. JSON parsing would probably require much less code. So if the point is to show just any custom datasource that might be better. However, if we think clients that will want this functionality will be dealing with XML 80% of the time, then it's fine to keep it as as.

The name `Mil2525DMessageParser` definitely needs to change however, IMO.

ldanzinger:
> The name Mil2525DMessageParser definitely needs to change 

Seems like a good idea. Let's leave XML for now since that is how it is done already and that is how customers that are familiar with our "military messages" in the message processor with 10.2.x will likely  expect it

philium:
@rolson @ldanzinger I renamed `Mil2525DMessage` to `Message`, although the former seems to make sense based on the name of the XML file: `Mil2525DMessages.xml`.

rolson:
In my opinion we should move from the "Message" terminology. Also the xml file should be renamed. @ldanzinger can you help @philium come up with appropriate terminology. cc @ralfgottschalk

rolson:
"Message" is a loaded term because of our past apis of "Message Processor" etc.

ldanzinger:
@rolson @philium I find "Messages" to be an appropriate term. Even though there is no one standard for "messages", this is still a commonly used term to refer to pieces of information sent between individuals. This is one of the key capabilities in the DSA app (https://developers.arcgis.com/example-apps/dsa-app-qt/#real-time-feeds), and this sample can be thought of as a very stripped down version of the DSA app. We actually ended up naming all of our classes related to displaying military message feeds with the "Message" prefix - https://github.com/Esri/dynamic-situational-awareness-qt/tree/master/Shared/messages

If you are really keen on removing the term all together, we could just call it "MilitarySymbolData"/"MilitarySymbolDataParser" or something, but I think what we have here represents a more realistic scenario and it mirrors the DSA app, which was built based on the requirements Kerry has gathered over the years working with lots of customers.

rolson:
@ldanzinger ok - I will defer to your expertise on this. what is the appropriate name for the xml file?

PR #644, opened 2019-01-03
Change OptionsTableViewController.Option initializer to take an image…

philium:
… rather than an image name

It really should be the callers responsibility to lookup images by name.


philium:
@pgruenler Thank you!

PR #646, opened 2019-01-03
Remove nil completion parameter in calls to UIViewController.present(…

philium:
…_:animated:completion:)

The completion parameter has a default value of `nil`.


philium:
@gdhakal Thank you!

PR #653, opened 2019-02-12
Remove check of minScale == maxScale

dg0yal:
100.5 resolves the issue that prevented downloading a single scale level.


dg0yal:
@philium please review

philium:
@dg0yal Thanks for making these changes. I noticed that you branched from `master`, but there have been more recent changes to `ExportTilesViewController.swift` on the `v.next` branch. Could you please branch from `v.next` and make your changes there? That way we don't have to resolve any conflicts and we don't pull in any additional commits from `master`.

dg0yal:
Ah, my mistake. I tried to use the web editor to create this patch, I based it off the wrong branch.

Looks like v.next doesn't need this change. The guard to fail when minscale==maxscale has been removed and replaced with just a check that minscale isn't smaller than maxscale.

https://github.com/Esri/arcgis-runtime-samples-ios/blob/v.next/arcgis-ios-sdk-samples/Layers/Export%20tiles/ExportTilesViewController.swift#L86-L91

PR #654, opened 2019-02-19
Search For Webmap by Keyword: Use milliseconds for uploaded date

philium:


arcgis-ios-sdk-samples/Cloud & Portal/Search for webmap by keyword/SearchForWebmapByKeywordViewController.swift
```diff
@@ -84,9 +84,11 @@ class SearchForWebmapByKeywordViewController: UICollectionViewController {
         // if the last search hasn't returned yet, cancel it
         lastQueryCancelable?.cancel()
         
-        // webmaps authored prior to July 2nd, 2014 are not supported - so search only from that date to the current time
-        let date1 = dateFormatter.date(from: "July 2, 2014")!
-        let dateString = "uploaded:[000000\(Int(date1.timeIntervalSince1970)) TO 000000\(Int(Date().timeIntervalSince1970))]"
+        // webmaps authored prior to July 2nd, 2014 are not supported - so
+        // search only from that date to the current time
+        let dateRange = dateFormatter.date(from: "July 2, 2014")!...Date()
```
mhdostal: Do you want to force unwrap this?  I know it was that way in the original code, but should it be updated since it's not really a "best practice"?

philium: Good question. I think force unwrapping here is fine since the input (and therefore the output) is fixed. I don't think its any different than force unwrapping the creation of a URL from a string literal, which we do all over the place.

mhdostal: Ok, makes sense to me.

PR #659, opened 2019-02-27
Update project file as recommended by Xcode

philium:
Xcode recommends these changes whenever a change is made to the project file.


gdhakal:
@philium It looks like the files are just removed and added back. Do you know why Xcode recommends this change?

philium:
@gdhakal In my experience, Xcode recommends changes like this from time to time when recently added files are not ordered properly. That usually is the result of resolving a merge conflict in the project file. Whenever another change is made to the project file (adding a file, renaming a file, etc.), Xcode rebuilds the project file, causing ordering issues like this to show up. I've found it is helpful to commit these recommended changes via a separate PR so as not to pollute the next PR that makes a change to the project file.

gdhakal:
@philium Thanks for clarifying that.

PR #663, opened 2019-03-13
Add Integrated Mesh Layer sample

philium:


arcgis-ios-sdk-samples/Scenes/Add an integrated mesh layer/AddIntegratedMeshLayerViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+import UIKit
+import ArcGIS
+
+/// A view controller that manages the interface of the Add an Integrated Mesh
+/// Layer sample.
+class AddIntegratedMeshLayerViewController: UIViewController {
+    /// The scene view managed by the view controller.
+    @IBOutlet var sceneView: AGSSceneView! {
+        didSet {
+            sceneView.scene = makeScene()
```
sbaskaran: you could add `self` here for more clarity. What do you think?

philium: > you could add `self` here

`self.sceneView`, `self.makeScene()`, or both?

> for more clarity.

Could you please elaborate? I'm not seeing how adding `self` would make it more clear. Could you help me understand?

> What do you think?

Our practice on the samples team has been to follow the leading of Apple and the majority of the Swift community, which omit `self` when accessing instance members unless absolutely necessary. There are some older samples that still use `self` when not necessary, but we're trying to clean those up as we go.

sbaskaran: Well, I prefer to use `self` just to make it clear we are accessing a property (in this case, it's within the `didSet` method anyway) and in cases where there might be a local variable with a same name. It also didn't appear to be consistent with some of the other samples I looked at. 

I wasn't aware Apple's recommended way is to omit `self` unless it's absolutely necessary

philium: Thank you for the explanation! That was helpful.

It isn't Apple recommended so much as Apple practiced. Most (if not all) of their source code does what they describe in the Swift Programming Language [book](
https://docs.swift.org/swift-book/LanguageGuide/Methods.html#ID235):

> Every instance of a type has an implicit property called `self`, which is exactly equivalent to the instance itself. You use the `self` property to refer to the current instance within its own instance methods.
>
> …
>
> In practice, you don’t need to write `self` in your code very often. If you don’t explicitly write `self`, Swift assumes that you are referring to a property or method of the current instance whenever you use a known property or method name within a method. …
>
> The main exception to this rule occurs when a parameter name for an instance method has the same name as a property of that instance. In this situation, the parameter name takes precedence, and it becomes necessary to refer to the property in a more qualified way. You use the `self` property to distinguish between the parameter name and the property name.

sbaskaran: Ok, good to know

PR #667, opened 2019-03-19
Use middle truncation for source file names

philium:


philium:
@pgruenler Thank you!

PR #668, opened 2019-03-20
Add Map Reference Scale sample

philium:


arcgis-ios-sdk-samples/Maps/Map reference scale/MapReferenceScaleSettingsViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+    private var referenceScaleObserver: NSObjectProtocol?
+    /// The observer of the scale of the map.
+    private var scaleObserver: NSObjectProtocol?
+    
+    // MARK: UIViewController
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        
+        precondition(map != nil)
```
mhdostal: If the client does not set a map, will this `precondition` crash the app?

arcgis-ios-sdk-samples/Maps/Map reference scale/MapReferenceScaleSettingsViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import UIKit
+import ArcGIS
+
+/// The delegate of a `MapReferenceScaleSettingsViewController`.
+protocol MapReferenceScaleSettingsViewControllerDelegate: AnyObject {
+    /// Tells the delegate that the user changed the map scale.
```
mhdostal: Are these comment types (with three `///`) the recommended practice for Samples?

arcgis-ios-sdk-samples/Maps/Map reference scale/MapReferenceScaleSettingsViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+    }()
+    /// The scale values displayed in the reference scale picker view.
+    private let possibleReferenceScales = [500_000, 250_000, 100_000, 50_000]
+    
+    /// Returns a string containing the formatted value of the provided scale.
+    ///
+    /// - Parameter scale: A scale value.
+    /// - Returns: A string.
+    func string(fromScale scale: Double) -> String {
+        return String(format: "1:%@", scaleFormatter.string(from: scale as NSNumber)!)
```
mhdostal: can there ever be a case where `scaleFormatter.string` can return nil?  I know that sometimes a map `scale` will return `nan` if the map isn't loaded (or other circumstances), so would that cause problems with this?

I know in general we want to avoid force-unwrapping...

arcgis-ios-sdk-samples/Maps/Map reference scale/MapReferenceScaleSettingsViewController.swift
philium: Correct, just like if they system didn't set one of the IB outlets. Both the data model and the views are needed for this view controller to work properly. I'd love to be able to use an initializer (something like `MapReferenceScaleSettingsViewController.init(map:)`), but that isn't possible with a storyboard. It is possible with a XIB, but XIBs do not support static cells.

Since I made the map an IUO, a crash would happen anyway the first time it was accessed before it was set. I just wanted to make it explicitly clear that the view controller expects the map to be set before being presented. Could I have made the map just an optional? Yes. And there would be a lot more question marks.

arcgis-ios-sdk-samples/Maps/Map reference scale/MapReferenceScaleSettingsViewController.swift
philium: Documentation comments (`/// …` or `/** … */`) are the recommended practice when documenting any symbol. That way the comments appear in Xcode's Quick Help:

Screen Shot 2019-03-20 at 12 52 27 PM

I've followed that practice when writing comments above any symbol (with the exception of local variables) in all of the samples that I've implemented.

arcgis-ios-sdk-samples/Maps/Map reference scale/MapReferenceScaleSettingsViewController.swift
philium: > can there ever be a case where `scaleFormatter.string` can return nil?

Good question. I haven't found a case where it does:

```swift
numberFormatter.string(from: Double(0) as NSNumber) // "0"
numberFormatter.string(from: Double.nan as NSNumber) // "NaN"
numberFormatter.string(from: Double.infinity as NSNumber) // "+∞"
```

I suspect it returns an optional for the sake of Objective-C. I tested it and `-[NSNumberFormatter stringFromNumber:]` will return `nil` when passed `nil`. Obviously that isn't possible from Swift.

arcgis-ios-sdk-samples/Maps/Map reference scale/MapReferenceScaleSettingsViewController.swift
mhdostal: Is that documented anywhere for the samples?  If more people are writing samples, it would be good to be consistent.

philium: Good question. I don't think it is in any sample-specific documentation, but we do have a backlog item for documenting our process. We could add this when working on that issue.

PR #670, opened 2019-03-20
Add sample to display WFS Layer

sbaskaran:


arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+class DisplayWFSViewController: UIViewController {
+
```
philium: Please remove this empty line. It generates a warning for violating the [Vertical Whitespace after Opening Braces](
https://github.com/realm/SwiftLint/blob/master/Rules.md#vertical-whitespace-after-opening-braces) linting rule.

Screen Shot 2019-03-20 at 3 59 46 PM

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+    @IBOutlet private weak var mapView: AGSMapView!
+    private var wfsFeatureTable : AGSWFSFeatureTable!
```
philium:
```suggestion
    private var wfsFeatureTable: AGSWFSFeatureTable!
```

Please remove the space before the colon. It generates a warning for violating the [Colon](
https://github.com/realm/SwiftLint/blob/master/Rules.md#colon) linting rule.

Screen Shot 2019-03-20 at 4 04 05 PM

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+        
+        // Assign the map to the map view
+        mapView.map = map
+
+        // Add the source code button item to the right of navigation bar
+        (navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["DisplayWFSViewController"]
+    }
+    
+    private func populateFeaturesFromQuery() {
+
```
philium: Please remove this empty line as it generates a warning.

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+        let params = AGSQueryParameters()
+        params.geometry = self.mapView.visibleArea?.extent
+        params.spatialRelationship = .intersects
+        
+        // Show progress
+        SVProgressHUD.show(withStatus: "Querying")
+    
+        // Populate features based on query
+        self.wfsFeatureTable.populateFromService(with: params, clearCache: true, outFields: ["*"]) { [weak self] (result: AGSFeatureQueryResult?, error: Error?) in
+
```
philium: Please remove this empty line as it generates a warning.

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+        
+        // Initialize a WFS feature table with service URL and uniquely identifying table name
+        wfsFeatureTable = AGSWFSFeatureTable(url: wfsServiceURL, tableName: wfsTableName)
+        
+        // Set feature request mode to manual - only manual is supported at v100.5.
+        // In this mode, you must manually populate the table - panning and zooming won't request features automatically.
+        wfsFeatureTable.featureRequestMode = .manualCache
+    
+        // Initialize a feature layer from WFS feature table
+        let wfsFeatureLayer = AGSFeatureLayer(featureTable: self.wfsFeatureTable)
```
philium: FYI: Since you created `wfsFeatureTable` in this scope, and therefore know that it is non-`nil`, going through the IUO property creates more work for the compiler. It needs to check for `nil` each time the property is accessed so that it can trap when the value is `none`. In cases like this, it is better to store the needed value to a local variable and use that in the local scope where needed and to update the value of the IUO property. Does that make sense? It's not something you have to fix. Just keep it in mind for the future.

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+        // Populate features based on query
+        self.wfsFeatureTable.populateFromService(with: params, clearCache: true, outFields: ["*"]) { [weak self] (result: AGSFeatureQueryResult?, error: Error?) in
+            
+            // Hide Progress
+             SVProgressHUD.dismiss()
+            
+            // Check for error
+            if let error = error {
+                self?.presentAlert(error: error)
+            } else {
```
philium: I would suggest unwrapping `result` here so that you don't have to handle the `nil` case in the print statement:

```swift
} else if let result = result {
    …
    print("Populated \(result.featureEnumerator().allObjects.count) features.")
}
```

You also might consider making this the `if` clause and the error case the `else if` since (hopefully) `result` being non-`nil` is the most likely case.

arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
```diff
@@ -493,6 +493,14 @@
 				<key>descriptionText</key>
 				<string>Display a map with an OpenStreetMap basemap.</string>
 			</dict>
+			<dict>
+				<key>displayName</key>
+				<string>Display WFS</string>
```
saratchandrakarumuri: Display name is little different than what we have in issue and read me doc. We can either update sample name or readme doc to maintain uniformity.

arcgis-ios-sdk-samples/Layers/Display WFS/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## How to use the sample
+
+Pan and zoom to see features within the current map extent.
+
+## How it works
+1. Create an instance of `AGSWFSFeatureTable` with a URL. 
+2. Set the feature request mode to `manualCache`.
+3. Create an instance of `AGSFeatureLayer` from the feature table and add it to the map.
+4. Whenever map's view point is changed, call `populateFromService(with:clearCache:outFields:)` to load the table with data for the current visible extent
```
philium:
```suggestion
4. Whenever map's view point is changed, call `populateFromService(with:clearCache:outFields:completion:)` to load the table with data for the current visible extent.
```

arcgis-ios-sdk-samples/Layers/Display WFS/README.md
```diff
@@ -0,0 +1,33 @@
...
+## How it works
+1. Create an instance of `AGSWFSFeatureTable` with a URL. 
+2. Set the feature request mode to `manualCache`.
+3. Create an instance of `AGSFeatureLayer` from the feature table and add it to the map.
+4. Whenever map's view point is changed, call `populateFromService(with:clearCache:outFields:)` to load the table with data for the current visible extent 
+
+## Relevant API
+* `AGSFeatureLayer`
+* `AGSWFSFeatureTable`
+* `AGSWFSFeatureTable.PopulateFromService(with:clearCache:outFields:)`
```
philium:
```suggestion
* `AGSWFSFeatureTable.populateFromService(with:clearCache:outFields:completion:)`
```

arcgis-ios-sdk-samples/Layers/Display WFS/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+WFS is an open standard with functionality similar to ArcGIS feature services. Runtime support for WFS allows you to interoperate with open systems, which are often used in inter-agency efforts, like those for disaster relief.
+
+![](image1.png)
+
+## How to use the sample
+
+Pan and zoom to see features within the current map extent.
+
+## How it works
```
philium: Please ensure that each header has an empty line between it and its contents. I've seen a case where a list didn't render properly on iOS when there wasn't an empty line.

arcgis-ios-sdk-samples/Layers/Display WFS/README.md
```diff
@@ -0,0 +1,33 @@
...
+* `AGSGeoView.viewpointChangeHandler`
+
+## About the data
+This service shows building footprints for downtown Seattle. For additional information, see the underlying service [on ArcGIS Online](https://arcgisruntime.maps.arcgis.com/home/item.html?id=1b81d35c5b0942678140efc29bc25391).
+
+## Tags
+
+OGC, WFS, feature, web, service, layers, browse, catalog, interaction cache
+
+
```
philium: Not a big deal by any means, but we usually leave a single blank line at the end of a text file. We have a linting rule which checks for that in source files, but we don't lint readmes.

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+    
+    private func populateFeaturesFromQuery() {
+        
+        // Create query parameters
+        let params = AGSQueryParameters()
+        params.geometry = self.mapView.visibleArea?.extent
+        params.spatialRelationship = .intersects
+        
+        // Show progress
+        SVProgressHUD.show(withStatus: "Querying")
```
saratchandrakarumuri: This is raising a flag with Main thread checker with following warning

`Main Thread Checker: UI API called on a background thread: -[UIApplication applicationState]`

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+    
+        // Populate features based on query
+        self.wfsFeatureTable.populateFromService(with: params, clearCache: true, outFields: ["*"]) { [weak self] (result: AGSFeatureQueryResult?, error: Error?) in
```
philium: When panning/zooming, does an in-process query automatically get cancelled when a new query is submitted? If not, we might want to hang on to the returned cancelable and cancel it when a new query is submitted.

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
philium: The HUD seems to rapidly disappear and appear again while interacting, rather than staying visible. Not a big deal, but I'm wondering if there's anything we can do about it? I would hope that `SVProgressHUD` is smart enough to not hide and show when asked to show when already visible. If not, there is an `isVisible` class method we could check ourselves. If we were to cancel the previous request, perhaps we could not dismiss the HUD when the error is cancelled by user? Those are just a couple ideas. I didn't try any of them.

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+        // Apply renderer to feature layer
+        wfsFeatureLayer.renderer = renderer
+        
+        // Add feature layer to the map
+        map.operationalLayers.add(wfsFeatureLayer)
+        
+        // Query for features whenever map's viewpoint changes
+        self.mapView.viewpointChangedHandler = { [weak self] in
+            DispatchQueue.main.async {
+                self?.populateFeaturesFromQuery()
```
saratchandrakarumuri: This is causing calling the query function multiple times for a single pan and it is impacting the performance of the sample. It would be better to cancel all previous queries

wfsPan

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
sbaskaran: > it is better to store the needed value to a local variable and use that in the local scope where needed

Do this in addition to having the IUO property, correct?

philium: Correct.

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
sbaskaran: I think not dismissing the alert when it's a user canceled error makes more sense.

arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
sbaskaran: I followed the naming patterns found for WMS Layer. Would you prefer the name to be `Display a WFS layer`? @saratchandrakarumuri @philium

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
sbaskaran: Hmm, I'm not sure why that would be. There are other samples that use `SVProgressHUD`, is the warning present in only this one?

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
sbaskaran: Yes, good point

arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
philium: I would go with "Display a WFS layer" as that is the name of the issue. Our standard practice is to title the sample implementation issue with the name of the sample.

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+            // Check and get results
+            if let result = result {
+                // The resulting features should be displayed on the map
+                // Print the count of features
+                print("Populated \(result.featureEnumerator().allObjects.count) features.")
+            }
+            // Check for error. If it's a user canceled error, do nothing.
+            // Otherwise, display an alert.
+            else if let error = error {
+                if error.localizedDescription.contains("User canceled error") {
```
philium: We normally check for this by testing the code of the error like so:

```swift
(error as NSError).code != NSUserCancelledError
```

Did you try that? If so, did it not work for you?

sbaskaran: That's a better way for sure. Will change it

philium:
@sbaskaran Thanks for making those changes. Everything looks good except for a couple new linting warnings. The project should compile with zero warnings. If you don't have swiftlint installed, that should generate a warning. If it didn't please let me know. Could you please install SwiftLint (`brew install swiftlint`) and build again? The linter automatically runs as part of the build process.

arcgis-ios-sdk-samples/Layers/Display WFS/DisplayWFSViewController.swift
saratchandrakarumuri: Actually I am seeing this in some other samples where we trigger progress HUD this from a completion handler. Maybe we can handle all these in a different issue? what do you think @philium ?

philium: I am not able to reproduce the the warning with this sample. But if it is happening in other samples, I would definitely create an issue for it. As to whether we can create a one-size-fits-all solution, I don't know. Maybe? I'd have to think about it.

sbaskaran:
Nice catch @saratchandrakarumuri I've changed the group name to match. Please take another look.

philium:
@sbaskaran This is not a big deal, but the screenshot still shows the old title.

sbaskaran:
@philium Updated

sbaskaran:
@philium @saratchandrakarumuri resolved merge conflicts

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -7,12 +7,16 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		092A1C2722456D9500C5F230 /* DisplayWFS.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 092A1C2322456D9500C5F230 /* DisplayWFS.storyboard */; };
+		092A1C2A22456D9500C5F230 /* DisplayWFSViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 092A1C2622456D9500C5F230 /* DisplayWFSViewController.swift */; };
+		092A1C2B22456DBF00C5F230 /* DisplayWFSViewController.swift in Resources */ = {isa = PBXBuildFile; fileRef = 092A1C2622456D9500C5F230 /* DisplayWFSViewController.swift */; };
```
philium: `DisplayWFSViewController.swift` is being copied as part of both the Copy Bundle Resources and Copy Files phases. Since our practice is to copy source files as part of the Copy Files phase, please remove it from the Copy Bundle Resources phase.

Screen Shot 2019-03-22 at 12 38 38 PM

PR #676, opened 2019-03-25
Change development region from English to en

philium:
Prevents a warning when building with Xcode 10.2 since `English` has been deprecated in favor of `en`. This change is backwards compatible with Xcode 10.1.

### DIFF
--- a/arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
+++ b/arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
@@ -3499,7 +3499,7 @@
 			};
 			buildConfigurationList = 3E23A9DB1AFC28F6002E2214 /* Build configuration list for PBXProject "arcgis-ios-sdk-samples" */;
 			compatibilityVersion = "Xcode 3.2";
-			developmentRegion = English;
+			developmentRegion = en;
 			hasScannedForEncodings = 0;
 			knownRegions = (
 				en,


esreli:
I don't know enough about the problem you're trying to fix to offer good feedback to this PR. Can you elaborate?

philium:
@esreli The following is the warning generated when opening the project in Xcode 10.2:

Screen Shot 2019-03-25 at 3 42 41 PM

Tapping on it brings up the following sheet:

Screen Shot 2019-03-25 at 3 44 25 PM

Tapping Migrate results in the change submitted in this PR.

esreli:
Was the result of this migration what you submit in this PR?

philium:
@esreli I'm not sure what you mean. From the Xcode [release notes](https://developer.apple.com/documentation/xcode_release_notes/xcode_10_2_release_notes):

> Opening a project that uses any deprecated localization identifiers now produces a warning for each one used. Selecting one of these warnings presents an assistant for migrating files in the associated legacy “lproj” directories to “lproj” directories named for the equivalent modern identifier. If necessary, this process also updates the project’s development region to a modern identifier. Migrated projects are compatible with older versions of Xcode.

Since this project doesn't actually have any localized resources, only that second part ("this process also updates the project’s development region to a modern identifier") is relevant. That is what this PR does. It updates the development region to use a modern identifier.

esreli:
> That is what this PR does. It updates the development region to use a modern identifier.

Answers my question exactly, thanks.

philium:
@esreli Thank you.

PR #681, opened 2019-03-29
Make MILLegendTableViewController.orderArray non-optional

philium:


philium:
Good question. The corresponding issue is specific to IUO (Swift) arrays, particularly those that are backing stores for table view data sources. It would be nice to look into whether `operationalLayers` could be made non-optional (and also made into an `Array`), but that is outside the scope of that issue.

PR #684, opened 2019-03-29
add browse wfs layers sample

sbaskaran:
@philium @saratchandrakarumuri please review.


arcgis-ios-sdk-samples/Layers/Browse WFS layers/README.md
```diff
@@ -0,0 +1,42 @@
...
+Some WFS services return coordinates in X,Y order, while others return coordinates in lat/long (Y,X) order. If you don't see features rendered or you see features in the wrong location, use the swap switch to change the coordinate order and reload.
+
+## How it works
+
+1. Create an instance of `AGSWFSService` with a URL to a WFS feature service. 
+2. Obtain a list of `AGSWFSLayerInfo` objects from `AGSWFSService.serviceInfo`
+3. When a layer is selected, create an instance of `AGSWFSFeatureTable` from the `AGSWFSLayerInfo` object.
+    3.1 Set the axis order if necessary.
+4. Create a feature layer from the feature table. 
+5. Add the feature layer to the map.
```
saratchandrakarumuri: Point `5.1` is displayed right beside `5`. Need to add spaces at end of point `5`

```suggestion
5. Add the feature layer to the map.  
```

sbaskaran: Adding a space has no effect. Introducing a line space between the lines appears as below. Is this acceptable per samples guidelines?

Screen Shot 2019-03-29 at 11 37 22 AM

saratchandrakarumuri: Two spaces at the end of the line did the trick for me. It looks like this for me

IMG_6D759249C82C-1

sbaskaran: cool, will add 2 spaces

arcgis-ios-sdk-samples/Layers/Browse WFS layers/BrowseWFSLayers.storyboard
```diff
@@ -0,0 +1,147 @@
...
+                                <tableViewCellContentView key="contentView" opaque="NO" clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="center" preservesSuperviewLayoutMargins="YES" insetsLayoutMarginsFromSafeArea="NO" tableViewCell="Zyv-ID-psG" id="0eh-SC-W3k">
+                                    <rect key="frame" x="0.0" y="0.0" width="375" height="43.5"/>
+                                    <autoresizingMask key="autoresizingMask"/>
+                                    <subviews>
+                                        <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Swap" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="oo0-DJ-aZi">
+                                            <rect key="frame" x="16" y="11.5" width="42" height="20.5"/>
+                                            <constraints>
+                                                <constraint firstAttribute="width" relation="greaterThanOrEqual" constant="42" id="TiR-N3-rMk"/>
```
philium: Neither of these width constraints are necessary since labels have an intrinsic width based on their text. Please remove them. I would suggest adding a horizontal spacing constraint from the trailing edge of the label to the leading edge of the switch, greater than or equal to the minimum distance you want between the two (you can delete the constant value to use the standard spacing if you want).

arcgis-ios-sdk-samples/Layers/Browse WFS layers/BrowseWFSLayers.storyboard
```diff
@@ -0,0 +1,147 @@
...
+                                                <constraint firstAttribute="width" constant="42" id="o73-8d-hzX"/>
+                                            </constraints>
+                                            <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                            <nil key="textColor"/>
+                                            <nil key="highlightedColor"/>
+                                        </label>
+                                        <switch opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="750" verticalHuggingPriority="750" contentHorizontalAlignment="center" contentVerticalAlignment="center" translatesAutoresizingMaskIntoConstraints="NO" id="n0O-Fe-0GS">
+                                            <rect key="frame" x="312" y="6.5" width="51" height="31"/>
+                                            <constraints>
+                                                <constraint firstAttribute="width" constant="49" id="Ys6-BA-aAH"/>
```
philium: Neither of these width constraints are necessary since switches have an intrinsic width. Please remove them.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/README.md
philium: Adding two spaces to the preceding line doesn't appear to work with the GitHub renderer. From what I can tell, nested ordered list of the form X.Y aren't really supported. You might consider changing the sublist to an unordered list (bullet) so that it renders consistently everywhere.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/WFSLayersTableViewController.swift
```diff
@@ -0,0 +1,216 @@
...
+    private var shouldSwapCoordinateOrder = false
+    private var selectedLayerInfo: AGSWFSLayerInfo?
+    private var lastQuery: AGSCancelable?
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+    }
+    
+    // A convenience type for the table view sections.
+    private enum Section: Int {
```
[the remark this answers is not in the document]
sbaskaran: I took this example from Manage Operational Layers example. I'm inclined to leave it as-is, the change could be made in all places separately, what do you think?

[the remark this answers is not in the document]
philium: Let me think about it. I know that `TilePackagesListViewController.Section` was introduced prior `CaseIterable`'s introduction in Swift 4.2. Both `TilePackagesListViewController` and `StatisticalQueryGroupAndSortViewController` include a `Section` enumeration conforming to `CaseIterable`.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/BrowseWFSLayers.storyboard
sbaskaran: If I remove the width constraint, I don't see the text. I think it's fine as-is, most of our other samples have width for labels too. 

The horizontal spacing between the two doesn't appear necessary too. In fact, if I add it, the switch moves to the right, and gets a little bit cut in the view

arcgis-ios-sdk-samples/Layers/Browse WFS layers/BrowseWFSLayers.storyboard
sbaskaran: you're right. Removed

sbaskaran:
@saratchandrakarumuri @philium I've addressed your comments. Please take another look. FYI, I'm holding off resolving merge conflicts until approval so I just do it once.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/WFSLayersTableViewController.swift
```diff
@@ -0,0 +1,216 @@
...
+        }
+        // Create query parameters
+        let params = AGSQueryParameters()
+        params.whereClause = "1=1"
+        
+        // Show progress
+        SVProgressHUD.show(withStatus: "Querying")
+        
+        // Populate features based on query
+        self.lastQuery = wfsFeatureTable.populateFromService(with: params, clearCache: true, outFields: ["*"]) { [weak self] (result: AGSFeatureQueryResult?, error: Error?) in
```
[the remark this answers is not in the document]
philium: I'm still curious about this. The documentation for `AGSWFSFeatureTable.populateFromService(with:clearCache:outFields:completion:)` doesn't specify from which thread the completion handler will be called.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## How to use the sample
+
+Select a layer from the list to display it on the map. 
+
+Some WFS services return coordinates in X,Y order, while others return coordinates in lat/long (Y,X) order. If you don't see features rendered or you see features in the wrong location, use the swap switch to change the coordinate order and reload.
+
+## How it works
+
+* Create an instance of `AGSWFSService` with a URL to a WFS feature service.
```
philium: I'm sorry for not being clear. When I said "You might consider changing the sublist to an unordered list", what I meant was the X.1 list items ("Set the axis order…", "The sample uses…"). Please keep this list as an ordered list to match all the other samples.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/WFSLayersTableViewController.swift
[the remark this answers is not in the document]
philium: I suppose what you have is fine for now.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/WFSLayersTableViewController.swift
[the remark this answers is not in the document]
sbaskaran: I don't think the populate method needs it, I thought SVProgressHD may need it but doesn't seem like it does.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/BrowseWFSLayers.storyboard
philium: > If I remove the width constraint, I don't see the text.

In the storyboard canvas or in the view at runtime? I was able to remove the constraint and still see the text in both places.

> most of our other samples have width for labels 

That's unfortunate as the text width can change based on the system font, the content size category (if using an adaptive font), etc. With Auto Layout, it is best to let views with intrinsic content size size themselves rather than constraining them to an exact width or height. As Apple recommends in their [Rules of Thumb](https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/AutolayoutPG/WorkingwithConstraintsinInterfaceBuidler.html#//apple_ref/doc/uid/TP40010853-CH10-SW7) when Working with Constraints in Interface Builder, "Avoid giving views a fixed height or width."

> The horizontal spacing between the two doesn't appear necessary

It isn't strictly necessary, but it is a good idea. It would prevent the label from expanding into the switch's territory. That's not a concern for this sample because the label's text is so short, but samples do change over time and the label's text may not always be the same. In addition, often new samples borrow from the implementation of existing samples. This cell could be copied for use in another sample, which has been done before.

> the switch moves to the right, and gets a little bit cut in the view

Did you add a fixed constraint rather than a greater than or equal to constraint?

I'm not trying to cause you more work for nothing. Another rule of thumb from Apple is to "Create constraints between a view and its nearest neighbor." I've found that following those rules as a habit can bring unforeseen benefits.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/BrowseWFSLayers.storyboard
```diff
@@ -0,0 +1,151 @@
...
+                                            <rect key="frame" x="16" y="0.0" width="343" height="43.5"/>
+                                            <autoresizingMask key="autoresizingMask"/>
+                                            <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                            <nil key="textColor"/>
+                                            <nil key="highlightedColor"/>
+                                        </label>
+                                    </subviews>
+                                </tableViewCellContentView>
+                            </tableViewCell>
+                            <tableViewCell clipsSubviews="YES" contentMode="scaleToFill" preservesSuperviewLayoutMargins="YES" selectionStyle="none" indentationWidth="10" reuseIdentifier="SettingsCell" id="Zyv-ID-psG" customClass="SettingsTableViewCell" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target">
```
philium: The following warning is logged at runtime for this cell:

> [Warning] Warning once only: Detected a case where constraints ambiguously suggest a height of zero for a tableview cell's content view. We're considering the collapse unintentional and using standard height instead.

The issue is that the table view is using self-sizing cells, but this cell doesn't have enough constraints to specify its height. Please add enough constraints to specify the height of the cell, making that warning go away. One way of doing that would be to remove the centerY constraint for the label and add a constraint from the top margin of the superview to the top of the label and add another from the bottom of the label to the bottom margin of the superview, both with a constant of zero.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/BrowseWFSLayers.storyboard
sbaskaran: Removing the width did not display the text of the label in runtime because the horizontal spacing (it was greater than or equal to) was large. Once I reduced that, the label displays fine. Could you take another look?

However, do note that having horizontal spacing in this case would again place a restriction. For eg, if the label were to be copied over to a new sample, it would still cause issues if the text of the newer label is longer because of incorrect horizontal spacing. Samples as they are today are mostly designed and coded for the specific workflow/data. So it's not all bad to have constraints specific to the current sample instead of generic ones as long as the expectation is to modify them correctly when copied over. As for programming practices, it's something the users look for consistency and I agree we should follow patterns and best practices.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/BrowseWFSLayers.storyboard
```diff
@@ -0,0 +1,151 @@
...
+                                            <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                            <nil key="textColor"/>
+                                            <nil key="highlightedColor"/>
+                                        </label>
+                                        <switch opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="750" verticalHuggingPriority="750" contentHorizontalAlignment="center" contentVerticalAlignment="center" translatesAutoresizingMaskIntoConstraints="NO" id="n0O-Fe-0GS">
+                                            <rect key="frame" x="312" y="6.5" width="51" height="31"/>
+                                        </switch>
+                                    </subviews>
+                                    <constraints>
+                                        <constraint firstItem="oo0-DJ-aZi" firstAttribute="leading" secondItem="0eh-SC-W3k" secondAttribute="leading" constant="16" id="ITM-f7-reN"/>
```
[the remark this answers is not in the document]
sbaskaran: I honestly didn't notice :) I have made changes per suggestion.

arcgis-ios-sdk-samples/Layers/Browse WFS layers/BrowseWFSLayers.storyboard
sbaskaran: Done

sbaskaran: Done

arcgis-ios-sdk-samples/Layers/Browse WFS layers/BrowseWFSLayers.storyboard
```diff
@@ -0,0 +1,150 @@
...
+                                            <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                            <nil key="textColor"/>
+                                            <nil key="highlightedColor"/>
+                                        </label>
+                                        <switch opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="750" verticalHuggingPriority="750" contentHorizontalAlignment="center" contentVerticalAlignment="center" translatesAutoresizingMaskIntoConstraints="NO" id="n0O-Fe-0GS">
+                                            <rect key="frame" x="310" y="6.5" width="51" height="31"/>
+                                        </switch>
+                                    </subviews>
+                                    <constraints>
+                                        <constraint firstItem="n0O-Fe-0GS" firstAttribute="leading" relation="greaterThanOrEqual" secondItem="oo0-DJ-aZi" secondAttribute="trailing" constant="150" id="DYn-vV-8um"/>
```
philium: I see what's going on. The purpose of this constraint is to limit the width to which the label can grow. But with such a large constant, it is actually being used to position the switch. I'm sorry for the confusion. I obviously didn't do a good enough job explaining what this constraint was for. Because a label's width is dependent on its text, the wider the text, the wider the label. With a sufficiently large text, the label would start to overlap the switch. This constraint is meant to limit the label from doing that. It says that the distance between the two must be greater than, or equal to, the constant. The constant is meant to be a small value; I think the standard spacing is like `8`. You can use the standard value by not specifying a constant:

Screen Shot 2019-04-01 at 3 40 41 PM

That allows the label's width to be as small as needed to fit the text. It also allows the label to grow to fit the text up until it is `Standard` spacing away from the switch. Then the system says, "That's it! That is the maximum you're allowed to grow." 

Does that make sense?

sbaskaran: Ah got it 👍

sbaskaran:
Thanks @philium and @saratchandrakarumuri 
Seems like you'd have to approve again after I resolved merge conflicts. Could you please?

arcgis-ios-sdk-samples/Scenes/Scene symbols/SceneSymbolsViewController.swift
```diff
@@ -63,14 +63,14 @@ class SceneSymbolsViewController: UIViewController {
         let diamondSymbol = AGSSimpleMarkerSceneSymbol(style: .diamond, color: .random(), height: 200, width: 200, depth: 200, anchorPosition: .center)
         let sphereSymbol = AGSSimpleMarkerSceneSymbol(style: .sphere, color: .random(), height: 200, width: 200, depth: 200, anchorPosition: .center)
         let tetrahedronSymbol = AGSSimpleMarkerSceneSymbol(style: .tetrahedron, color: .random(), height: 200, width: 200, depth: 200, anchorPosition: .center)
-
```
philium: I meant to comment on this earlier. Could you please revert the changes to this file? Putting the cursor on the line and pressing Ctrl+I (Editor > Structure > Re-Indent) should indent the lines properly. I think that's how they were before.

arcgis-ios-sdk-samples/Scenes/Scene symbols/SceneSymbolsViewController.swift
```diff
@@ -70,7 +70,7 @@ class SceneSymbolsViewController: UIViewController {
             let point = AGSPoint(x: x + 0.01 * Double(offset), y: y, z: z, spatialReference: .wgs84())
             return AGSGraphic(geometry: point, symbol: symbol)
         }
-
```
philium: It looks like one of the changes went away, but GitHub is still saying that this file was modified. Could you please get rid of this one as well?

sbaskaran:
Thank you both

PR #687, opened 2019-04-03
Fix Xcode project indentation

philium:
Incorrect indentation was introduced as a result of manually resolving a merge conflict. Xcode fixed the indentation, but not right away.


philium:
@gdhakal Thank you!

PR #689, opened 2019-04-04
fix issue where invisible cell still had a checkmark after unselecting

sbaskaran:
@saratchandrakarumuri @philium please review fix for an issue I ran into. Without this, the unselected cell still has a checkmark, if originally invisible.

### DIFF
--- a/arcgis-ios-sdk-samples/Layers/Browse WFS layers/WFSLayersTableViewController.swift
+++ b/arcgis-ios-sdk-samples/Layers/Browse WFS layers/WFSLayersTableViewController.swift
@@ -75,6 +75,7 @@ class WFSLayersTableViewController: UITableViewController {
             let cell = tableView.dequeueReusableCell(withIdentifier: "LayerCell", for: indexPath)
             let layerInfo = allLayerInfos[indexPath.row]
             cell.textLabel?.text = layerInfo.title
+            cell.accessoryType = (tableView.indexPathForSelectedRow == indexPath) ? .checkmark : .none
             return cell
         case .swapCoordinateOrder:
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Layers/Browse WFS layers/WFSLayersTableViewController.swift
```diff
@@ -75,6 +75,7 @@ class WFSLayersTableViewController: UITableViewController {
             let cell = tableView.dequeueReusableCell(withIdentifier: "LayerCell", for: indexPath)
             let layerInfo = allLayerInfos[indexPath.row]
             cell.textLabel?.text = layerInfo.title
+            cell.accessoryType = (tableView.indexPathForSelectedRow == indexPath) ? .checkmark : .none
```
philium: Since you're already setting the accessory type of the selected cell in `didSelectRowAt`, I would suggest setting the accessory type of the unselected cell in `didDeselectRowAt` (which gets called automatically when another row becomes selected, when multiple selection is disallowed).

sbaskaran: Yes I do call `didDeselectRowAt`, that deselects it but `cellForRowAt:` returns `nil` if the cell is invisible. So I couldn't set the `accessoryType` to `.none` in `didDeselectRowAt`. It works for visible cells as expected

philium: Right. I missed that. Thanks!

PR #690, opened 2019-04-04
Group layers sample

ghost:


arcgis-ios-sdk-samples/Layers/Group layers/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Use case  
+
+Group layers communicate to the user that layers are related and can be managed together.
+
+In a land development project, you might group layers according to the phase of development.
+
+## How it works
+
+1. Create an empty `GroupLayer`.
```
philium:
```suggestion
1. Create a new `AGSGroupLayer` object.
```

philium (changes requested):
Thank you Garima; I can tell you put a lot of work into this sample. I gave you a lot of feedback and requested some changes. Please take a look.

saratchandrakarumuri (dismissed):
Sample is working as expected

gdhakal:
@philium I've made the changes. Please re-review.

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+    ///     - section: Position of layer on the layers list.
+    /// - Returns: A layer if it exists else it returns nil.
+    func getLayer(from layers: [AGSLayer], at section: Int) -> AGSLayer? {
+        if section >= layers.count { return nil }
+        return layers[section]
+    }
+}
+
+class GroupLayersCell: UITableViewCell {
+    @IBOutlet var layerVisibilitySwitch: UISwitch!
```
philium: As you probably have noticed, the switches no longer have the proper padding on the right of the cell. You could solve that by constraining the trailing edge to the trailing edge of the margin of the superview (with a constant of zero), but there's also another problem. The label's text is very close to the switch. Too close, in my opinion. I think your first idea of setting the switch as the accessory view was a good one. We can't do that in the storyboard (at least with a prototype cell--it can be done with static cells), but we could do that here in code:

```swift
let layerVisibilitySwitch: UISwitch

required init?(coder aDecoder: NSCoder) {
    layerVisibilitySwitch = UISwitch()
    super.init(coder: aDecoder)
    accessoryView = layerVisibilitySwitch
}
```

Screen Shot 2019-04-11 at 3 55 09 PM

The other option would be to use our own label in the storyboard so that we could add a minimum spacing constraint between our label and the switch.

arcgis-ios-sdk-samples/Content Display Logic/Views/GroupLayersSectionView.swift
```diff
@@ -0,0 +1,35 @@
...
+import UIKit
+
+final class GroupLayersSectionView: UITableViewHeaderFooterView {
+    @IBOutlet var layerNameLabel: UILabel!
+    @IBOutlet var layerVisibilitySwitch: UISwitch?
+    weak var delegate: GroupLayersSectionViewDelegate?
+    
+    static let reuseIdentifier = String(describing: self)
+    
+    static var nib: UINib {
```
philium: I know it's been said that a `UITableViewHeaderFooterView` can be created from a XIB. After all, Apple has an API for registering a header/footer view with one. But one thing that's always bothered me about that approach is that it goes against Apple's own documentation. What do I mean? Take the following:

> If you have custom content to display, create the subviews for your content and add them to the view in the `contentView` property.

That's not where the custom content ends up when it comes out of the XIB. Instead of being subviews of the content view, they end up as subviews of the header/footer view itself:

Screen Shot 2019-04-11 at 3 51 43 PM

That may be perfectly fine in practice. I don't know. Just something to keep in mind.

arcgis-ios-sdk-samples/Content Display Logic/Views/GroupLayersSectionView.xib
```diff
@@ -0,0 +1,50 @@
...
+                            </connections>
+                        </switch>
+                    </subviews>
+                </stackView>
+            </subviews>
+            <constraints>
+                <constraint firstAttribute="bottom" secondItem="X6R-rt-aHC" secondAttribute="bottom" constant="5" id="4Da-1Z-GB0"/>
+                <constraint firstAttribute="trailing" secondItem="X6R-rt-aHC" secondAttribute="trailing" constant="5" id="OI6-c5-c6y"/>
+                <constraint firstItem="X6R-rt-aHC" firstAttribute="top" secondItem="XjP-9q-BD7" secondAttribute="top" constant="5" id="mKF-H0-Gdd"/>
+                <constraint firstItem="X6R-rt-aHC" firstAttribute="leading" secondItem="XjP-9q-BD7" secondAttribute="leading" constant="5" id="wLZ-ZN-DIT"/>
```
philium: Please constrain the leading and trailing edges of the stack view to the superview's margins, with a constant of zero. That matches the inset of the rows themselves much more closely, albeit not exactly. But close enough.

You'll need to tick down the horizontal content compression resistance priority of the label by one, otherwise it'll push the switch off to the side:

image

arcgis-ios-sdk-samples/Content Display Logic/Views/GroupLayersSectionView.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+
+final class GroupLayersSectionView: UITableViewHeaderFooterView {
```
philium: Since this view is specific to this sample, please move it into the directory for the sample. `Content Display Logic` is used for views (among other things) that are reusable and/or not specific to any one sample.

philium (commented):
Thank you for your changes! I do think things are looking better. I went through the PR again and have another round of feedback for you. Please take a look.

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+        
+        // Set the switch to on or off.
+        headerView.layerVisibilitySwitch?.isOn = layer.isVisible
+        if !layer.isVisible {
+            // Disable switch of child layers if the parent is not visible.
+            updateSwitchForRows(in: section, isEnabled: false)
+        }
+        
+        // To update the visibility of operational layers on switch toggle.
+        headerView.layerVisibilitySwitch?.tag = section
```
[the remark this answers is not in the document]
gdhakal: > That seems to be a bit fragile as the header view is relying on someone outside of itself to set the tag.

Agreed. 

I liked the second approach. Thanks!

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+        
+        guard let layer = getLayer(from: layers, at: section) else { return nil }
+        // Set label.
+        headerView.layerNameLabel.text = formattedValue(of: layer.name)
+        
+        // Set the switch to on or off.
+        headerView.layerVisibilitySwitch?.isOn = layer.isVisible
+        if !layer.isVisible {
+            // Disable switch of child layers if the parent is not visible.
+            updateSwitchForRows(in: section, isEnabled: false)
```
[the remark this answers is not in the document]
gdhakal: I'm not sure if I understood what you mean. In `tableView(_:cellForRowAt:)`, I'm setting the switch on or off based on the visibility of child layers. Since child layer's visibility can be true when parent layer's visibility is false, I have "disabled" child layers switch when the parent layer is turned off. That would communicate to user that they can't toggle the child when the parent is off.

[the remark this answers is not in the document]
philium: As far as I can tell, child layer switch values are being set twice: once in `tableView(_:viewForHeaderInSection:)` (by the call to `updateSwitchForRows(in:isEnabled:)`) and again in `tableView(_:cellForRowAt:)`. I don't know which is called first, but I imagine that `tableView(_:viewForHeaderInSection:)` is called before `tableView(_:cellForRowAt:)`. In either case, it would be nice to only set each switch's value once when the table data is loaded. Could `tableView(_:cellForRowAt:)` take into account the visibility of the parent layer when setting the switch value? Perhaps with a helper function:

```swift
func isLayerVisible(at indexPath: IndexPath) -> Bool {
    let groupLayer = layers[indexPath.section] as! AGSGroupLayer
    let childLayer = groupLayer.layers[indexPath.row] as! AGSLayer
    return groupLayer.isVisible && childLayer.isVisible
}
```

and then in `tableView(_:cellForRowAt:)`:

```swift
cell.layerVisibilitySwitch.isOn = isLayerVisible(at: indexPath)
```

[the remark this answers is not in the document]
gdhakal: I see what you mean. But we need to honor child layer's visibility even when the parent is turned off. This is what I mean - 

When group layer visibility is turned off, its child layers won't be visible on map. But those child layer's `isVisible` property could still be `true`. 

`return groupLayer.isVisible && childLayer.isVisible` will turn off the switch of child layer even when the child layer's `isVisible` property is `true`, which is not correct. 

Hence, I am "disabling" the switch of child layers when the parent is turned off. Switch disable is different than swift on/off state. 

I hope I didn't confuse you.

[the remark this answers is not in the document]
philium: Ah, that makes sense. In that case, could we still do that in `tableView(_:cellForRowAt:)`?

```swift
cell.layerVisibilitySwitch.isOn = childLayer.isVisible
cell.layerVisibilitySwitch.isEnabled = layers[indexPath.section].isVisible
```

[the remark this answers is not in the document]
gdhakal: yes, thank you!

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+    // MARK: - Helper methods
+    
+    /// Ensures we cannot toggle visibility of child layers when the parent is turned off.
+    ///
+    /// - Parameters:
+    ///     - section: Section number of the table view.
+    ///     - enabled: Indicates if the switch should be enabled or disabled.
+    func updateSwitchForRows(in section: Int, isEnabled: Bool) {
+        guard let visibleRows = tableView.indexPathsForVisibleRows else { return }
+        let visibleRowsForSection = visibleRows.filter { $0.section == section }
```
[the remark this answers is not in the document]
gdhakal: yes, I saw there was unnecessary check for section in line 106. I removed that. 

I think we do not need lazy filter because there is no chaining operations.

arcgis-ios-sdk-samples/Layers/Group layers/GroupLayersSectionView.xib
```diff
@@ -0,0 +1,50 @@
...
+                <switch opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="750" verticalHuggingPriority="750" contentHorizontalAlignment="center" contentVerticalAlignment="center" on="YES" translatesAutoresizingMaskIntoConstraints="NO" id="9ku-4w-xzb">
+                    <rect key="frame" x="321" y="5" width="51" height="34"/>
+                    <connections>
+                        <action selector="toggleSwitch:" destination="-2" eventType="valueChanged" id="ndi-hb-abW"/>
+                    </connections>
+                </switch>
+            </subviews>
+            <constraints>
+                <constraint firstItem="10R-ek-huY" firstAttribute="top" secondItem="XjP-9q-BD7" secondAttribute="top" constant="5" id="1AJ-iz-MjX"/>
+                <constraint firstItem="9ku-4w-xzb" firstAttribute="leading" secondItem="10R-ek-huY" secondAttribute="trailing" constant="5" id="Lt2-qq-1br"/>
```
philium: Setting a fixed distance like this is fairly common, but it has one drawback: it makes the label wider than it needs to be. We can correct that by making this a `greaterThanOrEqual` relation. It allows a label to be as small as needed to fit its text while preventing the label from encroaching on the switch.

image

Before:

image

After:

image

Same with the table view cell.

arcgis-ios-sdk-samples/Layers/Group layers/GroupLayersSectionView.xib
```diff
@@ -0,0 +1,50 @@
...
+                <constraint firstItem="10R-ek-huY" firstAttribute="leading" secondItem="XjP-9q-BD7" secondAttribute="leading" constant="16" id="NKS-xX-JVd"/>
```
philium: Rather than hardcoding the leading space, please constrain to the superview's leading margin:

image

arcgis-ios-sdk-samples/Layers/Group layers/GroupLayersSectionView.xib
```diff
@@ -0,0 +1,50 @@
...
+                <constraint firstAttribute="bottom" secondItem="9ku-4w-xzb" secondAttribute="bottom" constant="5" id="QuA-ha-W9X"/>
+                <constraint firstAttribute="bottom" secondItem="10R-ek-huY" secondAttribute="bottom" constant="5" id="inj-lU-HBo"/>
+                <constraint firstItem="9ku-4w-xzb" firstAttribute="trailing" secondItem="XjP-9q-BD7" secondAttribute="trailing" constant="-5" id="w8s-53-RrP"/>
```
philium: Please constrain to the trailing margin of the superview:

image

Same with the table view cells.

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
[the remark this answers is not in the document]
philium: > there is no chaining operations.

Yes and no. The `for` loop isn't a chained operation _per se_, but it has the same effect. It would be functionally equivalent rewritten as a chained operation:

```swift
visibleRows.filter { $0.section == section }
    .forEach { indexPath in
        if let cell = tableView.cellForRow(at: indexPath) as? GroupLayersCell {
            cell.layerVisibilitySwitch.isEnabled = isEnabled
        }
    }
```

In either implementation, there are two operations: 1) filter, 2) iterate. The `filter` operation creates a new array containing the filtered elements. `lazy.filter`, on the other hand, does not create a new array. It returns an instance of `LazyFilterCollection<Base>`, which is a view into the original collection. The filtering hasn't actually happened. It happens on demand as the the lazy filter collection is iterated. Does that make sense?

Obviously the benefit of using a lazy collection is proportional to the size of the original collection. The size of `visibleRows` is relatively small, so it probably wouldn't make any difference in this case.

philium (changes requested):
Thank you for your changes! There are just a few remaining layout issues. Then this should be good to go.

arcgis-ios-sdk-samples/Layers/Group layers/GroupLayersSectionView.xib
gdhakal: If I constrain to the leading space of label and trailing space of switch to super's margin, it appears like this. 
Screen Shot 2019-04-15 at 5 02 41 PM

philium: Did you adjust the spacing constraint between the label and the switch? Currently it is as follows:

image

This means that the trailing edge of the label should be five points after the leading edge of the switch. It translates to the following linear equation: `label.leading = 1.0 * switch.leading + 5.0`. The label is allowed to underlap the switch by five points, which is what you're seeing in that last row.

A better constraint would be `switch.leading >= label.leading + c`, where `c` is some constant, such as the system spacing. It looks as follows in Interface Builder:

image

In order for that last row to be sized (and truncated) properly, you'll need to set the number of lines to `1`:

image

Screen Shot 2019-04-16 at 8 59 55 AM

Alternatively, you could change the relation of the switch's top and bottom constraints to `greaterThanOrEqual` and then center it vertically, allowing the label to grow to two lines, thereby increasing the height of that cell.

image

philium (changes requested):
The last commit merged `master` into this branch. Since `master` does not have the Choose Camera Controller sample, this PR now removes that sample from `v.next`. Please revert the last commit and update from `v.next` rather than `master`. I'll take another look once that issue has been fixed.

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -1543,6 +1559,19 @@
...
+			isa = PBXGroup;
+			children = (
+				12D6602E226697320005DE8A /* GroupLayers.storyboard */,
+				12D66032226697320005DE8A /* GroupLayersSectionView.swift */,
+				12D6602C226697320005DE8A /* GroupLayersSectionView.xib */,
+				12D6602F226697320005DE8A /* GroupLayersViewController.swift */,
+				12D66031226697320005DE8A /* LayersTableViewController.swift */,
+			);
+			name = "Group layers";
+			path = "New Group";
```
philium: Please fix the path of the group and the group's members.

image

philium (changes requested):
One thing left to fix and then this should be good to go.

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
gdhakal: done

philium: You fixed the path of the group, but not the path of the group's members:

image

gdhakal: Interesting. Group member's path look good at my end. 

Screen Shot 2019-04-17 at 9 29 45 AM

Screen Shot 2019-04-17 at 9 29 56 AM

gdhakal: nevermind. I saw where the problem is.

philium (approved):
Thank you!

PR #691, opened 2019-04-05
Display a KML Tour: Fix toolbar width

philium:
I thought for sure bar leading and trailing edges should be constrained to the safe area, but either something has changed or I was mistaken. The leading and trailing edges are now constrained to the superview. This ensures that the toolbar goes from edge to edge horizontally.


philium:
@pgruenler Thank you!

PR #693, opened 2019-04-09
Remove PortalItems.plist from Resources

philium:
It was accidentally a member of the app target. I thought I removed it, but obviously not completely. Xcode keeps recommending this change.


philium:
@rolson Thank you!

PR #697, opened 2019-04-16
Display Layer View State: Improvements to status handling

philium:
- Create status string as needed.
- Include all statuses in the status string.
- Reload rows only for layers whose status has changed.

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Display layer view state/LayerStatusViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Display layer view state/LayerStatusViewController.swift
@@ -22,8 +22,6 @@ class LayerStatusViewController: UIViewController, UITableViewDataSource, UITabl
     
     private var map: AGSMap!
     
-    private var viewStatusArray = [String]()
-    
     override func viewDidLoad() {
         super.viewDidLoad()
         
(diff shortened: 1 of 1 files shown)


philium:
@sbaskaran Could you please review this at your earliest convenience?

arcgis-ios-sdk-samples/Maps/Display layer view state/LayerStatusViewController.swift
```diff
@@ -70,43 +66,42 @@ class LayerStatusViewController: UIViewController, UITableViewDataSource, UITabl
...
-        case AGSLayerViewStatus.outOfScale:
-            return "Out of Scale"
-        case AGSLayerViewStatus.loading:
-            return "Loading"
-        case AGSLayerViewStatus.error:
-            return "Error"
-        default:
+        var statuses = [String]()
+        if status.contains(.active) {
+            statuses.append("Active")
```
sbaskaran: @philium I got a couple of questions
- Since the status at a time can only be one of these, can `if... else` be used? In that case, is there an advantage of the if statement over the switch case?
- Do we need to handle for the same status getting appended twice?

philium: > the status at a time can only be one of these

Since `AGSLayerViewStatus` is an option set (instead of an enum), it can have more than one status at a time. Therefore these do need to be separate `if` statements so that all of the statuses in the set are shown to the user.

When running the sample, it does briefly show multiple statuses.

> Do we need to handle for the same status getting appended twice?

No. Each of the statuses (`active`, `notVisible`, etc.) should be unique, so `statuses` should never contain duplicates.

sbaskaran: Ok good to know. Thanks

philium:
@sbaskaran Thank you!

PR #698, opened 2019-04-17
Make ListViewController.list non-optional

philium:
This class was so small that I also went and turned `selectAction` into a non-IUO.


philium:
I do prefer the delegate pattern. In this case, there were a couple of reasons why I left it the way it was. First, I wanted to keep this PR small and focused--implementing a delegate would require a change to at least one other source file. Second, I'm not sure `ListViewController` is worth keeping around. I might just end up replacing it with an action sheet.

PR #700, opened 2019-04-18
ContentTableViewController: Remove unused method parameter

philium:


philium:
@pgruenler Thank you!

PR #705, opened 2019-04-23
Rename BlendRendererSettingsVC to BlendRendererSettingsViewController

philium:


arcgis-ios-sdk-samples/Layers/Blend renderer/BlendRendererSettingsViewController.swift
```diff
@@ -16,19 +16,19 @@
 import UIKit
 import ArcGIS
 
-protocol BlendRendererSettingsVCDelegate: AnyObject {
-    func blendRendererSettingsVC(_ blendRendererSettingsVC: BlendRendererSettingsVC, selectedAltitude altitude: Double, azimuth: Double, slopeType: AGSSlopeType, colorRampType: AGSPresetColorRampType)
+protocol BlendRendererSettingsViewControllerDelegate: AnyObject {
+    func blendRendererSettingsViewController(_ blendRendererSettingsVC: BlendRendererSettingsViewController, selectedAltitude altitude: Double, azimuth: Double, slopeType: AGSSlopeType, colorRampType: AGSPresetColorRampType)
```
njarecha: Do you think, `blendRendererSettingsVC` should be changed to `blendRendererSettingsViewController` as well?

philium: I thought about that, but in the end I just didn't feel strongly either way about the parameter name since it isn't part of the symbol name (`BlendRendererSettingsViewControllerDelegate.blendRendererSettingsViewController(_:selectedAltitude:azimuth:slopeType:colorRampType:)`). I could change it if you'd like.

njarecha: Let's change it

PR #711, opened 2019-05-07
Add Authenticate with OAuth sample

philium:


arcgis-ios-sdk-samples/Cloud & Portal/Authenticate with OAuth/AuthenticateWithOAuthViewController.swift
```diff
@@ -0,0 +1,85 @@
...
+private let portalURL = URL(string: "https://www.arcgis.com")!
+/// The Client ID for an app registered with the server. The provided ID is for
+/// a public app created by the ArcGIS Runtime team.
+private let portalItemID = "e5039444ef3c48b8a8fdc9227f9be7c1"
+
+/// The identifier with which this application was registered with the portal.
+private let clientID = "lgAdHkYZYlwwfAhC"
+/// The URL for redirecting after a successful authorization (this must be
+/// configured in the Info plist).
+private let redirectURL = URL(string: "my-ags-app://auth")!
```
njarecha: You need to pass `redirectURL` as string in the OAuth config then why this is declared as `URL` instead of `String`?

njarecha: Also, why are you not using `clientID: "xHx4Nj7q1g19Wh6P", redirectURL: "iOSSamples://auth"` which is being used by other samples?

philium: > why this is declared as URL instead of String?

My practice has been to store all URLs as instances of `URL` rather than as raw strings, regardless of how it is used. I've found the validation provided by `URL.init(string:)` to be helpful in catching my mistakes.

> why are you not using…which is being used by other samples?

I'm using what is specified in the design. I'll send you the link.

njarecha: > My practice has been to store all URLs as instances of URL rather than as raw strings, regardless of how it is used. I've found the validation provided by URL.init(string:) to be helpful in catching my mistakes.

I don't see anything here to validate. The API requires it as string. There is no point creating URL with string and then get string from URL to pass in the API. I feel, you should change it to String.

philium: > I don't see anything here to validate.

I'm talking about URLs in general, not just this one. Some URLs are copied and pasted, others are manually entered. All of them are checked by the initializer to be well formed: "Returns nil if a URL cannot be formed with the string (for example, if the string contains characters that are illegal in a URL, or is an empty string)."

> I feel, you should change it to String.

I'll go ahead and change it at your request.

philium:
The team is looking into the OAuth settings issue you brought up. Either the sample design will be updated (in which case this sample will be updated on each platform), the other samples will be updated, or neither. In any case, I don't know that it is worth holding up this PR.

arcgis-ios-sdk-samples/Cloud & Portal/Authenticate with OAuth/AuthenticateWithOAuthViewController.swift
njarecha: If user of the sample wants to use their own `redirectURL` then they will have to use the same string which they have registered with the `clientID`. Hence, I don't see a value of creating URL for the validation you intend to do.

philium:
@njarecha Thank you!

hardikc07:
verified the sample pr and works as expected

PR #715, opened 2019-05-24
Adheres Section to CaseIterable so as to leverage allCases.count

esreli:
Adheres Section to CaseIterable so as to leverage allCases.count in place of a static integer.


esreli:
Good suggestions. Have another look?

PR #716, opened 2019-05-29
Refactor DemoTouchManager

philium:
This turns the demo touch manager into a shared instance of a class rather than an enum, moving almost all of the logic out of `UIApplication` and into the manager itself.

@njarecha Could you please review when you get a chance?
@saratchandrakarumuri Could you please verify that the Show Touch feature still works properly?


philium:
@saratchandrakarumuri Thank you!
@mhdostal Could you please review when you get a chance?

philium:
@mhdostal Thank you!

PR #717, opened 2019-05-29
Add Bristol3DModel as dependency of Choose Camera Controller sample

philium:
Fixes a bug where the model could not be found at runtime.


philium:
@rolson Thank you!

PR #723, opened 2019-06-28
Migrate to Swift 5

philium:
Most of the changes are to address new warnings. Xcode now warns when passing an IUO to a method that takes `Any` (not `Any?`), so I switched to using a non-optional local variable in most of those cases. A few of them I was able to make the property non-optional instead.

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Identify layers/IdentifyLayersViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Identify layers/IdentifyLayersViewController.swift
@@ -20,9 +20,6 @@ class IdentifyLayersViewController: UIViewController, AGSGeoViewTouchDelegate {
     
     private var map: AGSMap!
     
-    private var featureLayer: AGSFeatureLayer!
-    private var mapImageLayer: AGSArcGISMapImageLayer!
-    
     override func viewDidLoad() {
         super.viewDidLoad()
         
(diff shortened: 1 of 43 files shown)


arcgis-ios-sdk-samples/Maps/Identify layers/IdentifyLayersViewController.swift
```diff
@@ -33,25 +30,25 @@ class IdentifyLayersViewController: UIViewController, AGSGeoViewTouchDelegate {
         self.map = AGSMap(basemap: .topographic())
         
         //map image layer
-        self.mapImageLayer = AGSArcGISMapImageLayer(url: URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/SampleWorldCities/MapServer")!)
+        let mapImageLayer = AGSArcGISMapImageLayer(url: URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/SampleWorldCities/MapServer")!)
         
         //hide Continent and World layers
-        self.mapImageLayer.load { [weak self] (error: Error?) in
+        mapImageLayer.load { [weak mapImageLayer] (error: Error?) in
```
nixta: If the completion block is called, doesn't that make it implicit that `mapImageLayer` exists? So is there any point in bothering with the `weak`?

philium: > doesn't that make it implicit that `mapImageLayer` exists?

It does. 

> So is there any point in bothering with the `weak`?

Good question. The point is more about what happens if the completion handler isn't called. Without looking at the implementation, it is conceivable that `mapImageLayer` (or something owned by `mapImageLayer`) keeps a strong reference to the completion handler. If that's true, keeping a strong reference to the map image layer inside of the closure would introduce a retain cycle for the lifetime of the closure. If the view controller were to go away (quickly opened and closed, for example) before the load operation completes, that retain cycle might not be broken (assuming there is one). `unowned` would probably work too in this case, but I wanted to err on the side of caution.

philium:
@nixta Thank you!

PR #727, opened 2019-07-09
Fix bug in Unique Value Renderer

philium:
`self.featureLayer` was being reference before its value had been set.

The first commit fixes the bug. The second commit updates the sample to match the style used in more recent sample implementations.


philium:
@sbaskaran Thank you!

PR #728, opened 2019-07-15
Create terrain surface from a local raster

vquach2404:

PR #729, opened 2019-07-17
Update download data script

philium:
- Adds support for ZIP archived portal items.
- Uses filename suggested by server. The filenames in `PortalItems.plist` are no longer necessary and will be removed at a later date.
- File types are now stored outside of the script.

### DIFF
--- a/Scripts/download-portal-item-data.swift
+++ b/Scripts/download-portal-item-data.swift
@@ -14,11 +14,34 @@
 // limitations under the License.
 //
 
-// This scripts downloads data for portal items. It takes two arguments. The
+// This scripts downloads data for portal items. It takes three arguments. The
 // first is a path to a plist file defining the portal items to download. The
-// second is a path to the download directory.
+// is a path to a plist file defining how various file types should be
+// organized. The third is a path to the download directory.
+//
+// A mapping of item identifiers to relative paths is maintained inside the
+// download directory. This mapping affords efficiently checking whether an
+// item has already been downloaded.
 
 import Foundation
+
+protocol URLProvider {
+    func makeURL(filename: String) -> URL
+}
+
+struct DestinationURLProvider: URLProvider {
+    let downloadDirectory: URL
+    let fileTypes: [String: [String]]
+    
+    func makeURL(filename: String) -> URL {
+        var url = downloadDirectory
+        if let subdirectory = fileTypes.first(where: { $0.value.contains((filename as NSString).pathExtension) })?.key {
+            url.appendPathComponent(subdirectory, isDirectory: true)
+        }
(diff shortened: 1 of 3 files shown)


philium:
The [Create Terrain Surface from a Local Raster](https://github.com/Esri/arcgis-runtime-samples-ios/pull/728) implementation depends on the changes in this PR.

Scripts/download-portal-item-data.swift
```diff
@@ -100,49 +138,104 @@ struct PortalItem: Decodable {
...
+            dispatchGroup.enter()
+            let sourceURL = makeDataURL(portalURL: portalURL, itemIdentifier: portalItem.identifier)
+            downloadFile(at: sourceURL, destinationURLProvider: destinationURLProvider) { (result) in
+                switch result {
+                case .success(let url):
+                    // ' + 1' removes the leading path separator.
+                    downloadedItems[portalItem.identifier] = url.lastPathComponent
+                    dispatchGroup.leave()
+                case .failure(let error):
+                    print("warning: Error downloading item \(portalItem.identifier): \(error)")
```
rolson: need a `dispatchGroup.leave()` here?

Scripts/download-portal-item-data.swift
```diff
@@ -36,55 +59,70 @@ func makeDataURL(portalURL: URL, itemIdentifier: String) -> URL {
...
+    let process = Process()
+    process.executableURL = URL(fileURLWithPath: "/usr/bin/unzip", isDirectory: false)
+    process.arguments = ["-p", sourceURL.path]
+    process.standardOutput = fileHandle
+    try process.run()
+    process.waitUntilExit()
+}
+
+func downloadFile(at sourceURL: URL, destinationURLProvider: URLProvider, completion: @escaping (Result<URL, Error>) -> Void) {
+    let downloadTask = URLSession.shared.downloadTask(with: sourceURL) { (temporaryURL, response, error) in
```
rolson: if you are downloading something off a portal, you should use `AGSRequestOperation` so that it takes care of all authentication for you. You can give it an output url which will allow it to save the download somewhere on disk instead of in memory.

Scripts/download-portal-item-data.swift
philium: Not necessary since we go on to call `exit(_:)`, which returns `Never`.

Scripts/download-portal-item-data.swift
philium: It was a conscious decision to not have `ArcGIS` as a dependency of this script. All of the portal items downloaded by this script are public.

philium:
@rolson Thank you!

PR #736, opened 2019-07-25
Dictionary renderer with feature layer

vquach2404:


arcgis-ios-sdk-samples/Layers/Dictionary renderer with feature layer/DictionaryRendererViewController.swift
```diff
@@ -0,0 +1,77 @@
...
+        generatedGeodatabase.load { [weak self] (error: Error?) in
+            if let error = error {
+                self?.presentAlert(error: error)
+            } else {
+                for featureTable in generatedGeodatabase.geodatabaseFeatureTables {
+                    // Create a feature layer from the feature table.
+                    let featureLayer = AGSFeatureLayer(featureTable: featureTable)
+                    // Set the minimum visibility scale.
+                    featureLayer.minScale = 1000000
+                    featureLayer.load { [weak self] (error: Error?) in
```
philium: FYI: A closure will capture the same kind of reference to `self` as the enclosing scope by default. Since `self` is `weak` in the enclosing scope, this closure will weakly capture `self` without needing to specify that in the capture list.

```suggestion
                    featureLayer.load { (error: Error?) in
```

arcgis-ios-sdk-samples/Layers/Dictionary renderer with feature layer/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+![Dictionary renderer with feature layer](dictionary-renderer-with-feature-layer.png)
+
+## Use case
+
+A dictionary renderer uses a style file along with a rule engine to display advanced symbology. This is useful for displaying features using precise military symbology.
+
+## How to use the sample
+
+When launched, the sample will display military symbology. Pan and zoom to explore.
```
philium: The design has "Simply run the sample." for this section. But since all samples are contained within a single app, neither "launch" nor "run" are appropriate verbs. Maybe something like this?

```suggestion
This sample displays military symbology. Pan and zoom to explore the map.
```

@mi

PR #735, opened 2019-07-24
Use new AGSDictionarySymbolStyle initializer

philium:
Also makes the style an ODR.

PR #737, opened 2019-07-25
Update portal item identifier of mil2525d.stylx

philium:
Hosting of the style file has moved to https://www.arcgis.com/home/item.html?id=c78b149a1d52414682c86a5feeb13d30.

### DIFF
--- a/arcgis-ios-sdk-samples/Content Display Logic/PortalItems.plist
+++ b/arcgis-ios-sdk-samples/Content Display Logic/PortalItems.plist
@@ -6,7 +6,7 @@
 	<array>
 		<dict>
 			<key>identifier</key>
-			<string>5de1ff8d2aa243558ba057fa922beb84</string>
+			<string>c78b149a1d52414682c86a5feeb13d30</string>
 			<key>filename</key>
 			<string>mil2525d.stylx</string>
 		</dict>


rolson:
I'm not sure how this is being used, but I'm pretty sure that if you update to the new style file you need to use the new constructor of AGSDictionarySymbolStyle. Please check with Ralf Gottschalk

philium:
@rolson That has already be done as part of https://github.com/Esri/arcgis-runtime-samples-ios/pull/735. As I said, this is just changing the endpoint of the style file as requested. The file is the same at both the old and the new endpoint.

rolson:
if the file hasn't changed, just the endpoint, then the old (deprecated) constructor should be used.

philium:
Did you look at the linked PR? The endpoint of the _new_ style file has changed.

ralfgottschalk:
This is the new style file.  If using the new style file you should change to the new constructor as @rolson said.  If you use the old constructor the dictionary logic to assemble the symbols will be the old logic which is tightly coupled to the old style file.  The new style file has a different set of keys and would result in a whole bunch of unresolved symbols.

The nice thing about the new style files is the logic for the dictionary is inside the style file, which makes it easy to customize.

PR #738, opened 2019-07-26
Metadata PR for Maps

ChrisFrenchPDX:
@philium here are the `Maps` category metadata files, could you take a look and let me know if you see any errors? If not, I'll make PRs fro the other metadata categories. Thanks!

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Create and save a map/README.metadata.json
+++ b/arcgis-ios-sdk-samples/Maps/Create and save a map/README.metadata.json
@@ -1,0 +1,25 @@
+{
+    "category": "Maps",
+    "description": "@@TODO: This sample demonstrates how to create a map and save it to your portal",
+    "ignore": false,
+    "images": [
+        "image4.png",
+        "image1.png",
+        "image2.png",
+        "image3.png"
+    ],
+    "keywords": [
+        "@@TODO: *** PLEASE ADD ADDITIONAL KEYWORDS ***"
+    ],
+    "redirect_from": "/ios/latest/swift/sample-code/create-and-save-a-map.htm",
+    "relevant_apis": [
+        "@@TODO: *** PLEASE ADD RELEVANT APIS ***"
+    ],
+    "snippets": [
+        "@@TODO: *** PLEASE VERIFY THE INCLUSION OF THESE SNIPPETS ***",
+        "SaveAsViewController.swift",
+        "CreateSaveMapViewController.swift",
+        "CreateOptionsViewController.swift"
+    ],
+    "title": "@@TODO: Create and save a map"
+}
--- a/arcgis-ios-sdk-samples/Maps/Generate offline map (basemap by reference)/README.metadata.json
+++ b/arcgis-ios-sdk-samples/Maps/Generate offline map (basemap by reference)/README.metadata.json
@@ -1,0 +1,27 @@
+{
+    "category": "Maps",
+    "description": "@@TODO: Use the AGSOfflineMapTask to take a web map offline, but instead of downloading an online basemap, use one which is already on the device.",
+    "ignore": false,
+    "images": [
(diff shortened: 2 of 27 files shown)


arcgis-ios-sdk-samples/Maps/Create and save a map/README.metadata.json
```diff
@@ -0,0 +1,25 @@
+{
+    "category": "Maps",
+    "description": "@@TODO: This sample demonstrates how to create a map and save it to your portal",
+    "ignore": false,
+    "images": [
+        "image4.png",
```
philium: I noticed that these images are not in the same order as they appear in the README. We might want to preserve that ordering if we can.

arcgis-ios-sdk-samples/Maps/Map reference scale/README.metadata.json
```diff
@@ -0,0 +1,28 @@
...
+    "category": "Maps",
+    "description": "@@TODO: Set the map's reference scale and which feature layers should honor the reference scale.",
+    "ignore": false,
+    "images": [
+        "image1.png",
+        "image2.png"
+    ],
+    "keywords": [
+        "Maps",
+        "and",
```
philium: Instead of converting "&" to "and", it might be better to just omit it as a keyword altogether.

arcgis-ios-sdk-samples/Maps/Generate offline map (basemap by reference)/README.metadata.json
```diff
@@ -0,0 +1,27 @@
+{
+    "category": "Maps",
+    "description": "@@TODO: Use the AGSOfflineMapTask to take a web map offline, but instead of downloading an online basemap, use one which is already on the device.",
+    "ignore": false,
+    "images": [
+        "image1.png"
+    ],
+    "keywords": [
+        "Offline",
+        "AGSOfflineMapTask",
```
philium: It looks like the relevant APIs are being appended to the keywords. Might it be useful to keep those separate? They could be combined later at the point of use, correct?

arcgis-ios-sdk-samples/Maps/Create and save a map/README.metadata.json
ChrisFrenchPDX: Is this necessary for the iOS sample viewer? If it matters at all, my Python script turns the md files in ast trees, and iterates through the ast, and I just appends the image files to an array in the order they appear in the md file, so I'm pretty sure the images listed are ordered, but from last to first.

arcgis-ios-sdk-samples/Maps/Map reference scale/README.metadata.json
ChrisFrenchPDX: oh, good catch - I forget to add back a processing step which I took out for the android metadata files.

arcgis-ios-sdk-samples/Maps/Generate offline map (basemap by reference)/README.metadata.json
ChrisFrenchPDX: Yes, this is by design -- the keywords is used by page SEO, so I think we want to the class names in the page metadata so the sample pages will have a higher ranking on google, etc.

arcgis-ios-sdk-samples/Maps/Create and save a map/README.metadata.json
philium: I don't know how the images are going to be used, but if they are going to be displayed on the dev site, there is a benefit to preserving the order from the README. In this case, the images show a progression through the sample. The README has them in numerical order:

```markdown
![](image1.png)
![](image2.png)
![](image3.png)
![](image4.png)
```

That's probably going to be the case when there are multiple images, but I can't guarantee that.

PR #742, opened 2019-08-05
Add Control Annotation Sublayer Visibility sample

philium:


philium:
@saratchandrakarumuri Thanks for that observation. The behavior you described was an intentional design choice so that users could see the sublayers (in particular the one limited to a certain range) while adjusting the map scale. Do you think it should be removed?

> This behavior is inconsistent compared to other samples.

I think there used to be more samples with this behavior, but now there is only one other one, namely Change Atmosphere Effect. There are two others (Change Basemap and ArcGIS Vector Tiled Layer (URL)) that attempt to have this feature, but fail:

```swift
controller.popoverPresentationController?.passthroughViews?.append(mapView)
```

I think the issue is that `passthroughViews` is `nil` by default.

saratchandrakarumuri:
If it is intentional I guess it is ok and it makes sense in case of this sample and couple of other that you have mentioned.

philium:
Thank you @saratchandrakarumuri and @sbaskaran!

PR #743, opened 2019-08-08
Make SourceCodeViewController.filenames non-optional

philium:
Along with a few other properties in the filenames chain.


philium:
@njarecha Thank you!

PR #745, opened 2019-08-08
DemoTouchView -> DemoTouchManager

philium:
`DemoTouchView` no longer exists, if it ever existed in the first place. The name of the salient class in the renamed source file is [DemoTouchManager](https://github.com/Esri/arcgis-runtime-samples-ios/blob/4569d14586f0e1eea5d128085ea445434c073566/arcgis-ios-sdk-samples/Shared%20resources/Libraries/DemoTouchView/DemoTouchView.swift#L124):

```swift
class DemoTouchManager {
    ...
}
```

PR #739, opened 2019-07-29
Update Mil2525DMessages.xml

philium:
Fixes issues with the underlying data. Unfortunately the indentation changed, making it difficult to see the relevant changes. Basically the `<sic>` tag was removed and a few of the `<_control_points>` tags were updated.

PR #746, opened 2019-08-08
Resolved failing test in MessageParserTests

stevebaranski:
### Context 
A failing test was encountered during review of #745. In conversation with @philium, we ascertained that the failure was introduced via #739.

This PR fixes the failing test. 

### Testing

1. Run existing tests (via Cmd+U) against `v.next`, noting a failure on [this line](https://github.com/Esri/arcgis-runtime-samples-ios/blob/a0efcf50a37c086c543b69ce0ee639783427ecae/arcgis-ios-sdk-samplesTests/Sample%20Tests/GraphicsOverlayDictionaryRenderer3DViewControllerTests.swift#L55).
1. Checkout the PR branch, re-run the test, and confirm that it now passes.
1. Review the code in the revised test, to guard against unintended regressions.

### DIFF
--- a/arcgis-ios-sdk-samplesTests/Sample Tests/GraphicsOverlayDictionaryRenderer3DViewControllerTests.swift
+++ b/arcgis-ios-sdk-samplesTests/Sample Tests/GraphicsOverlayDictionaryRenderer3DViewControllerTests.swift
@@ -67,11 +67,11 @@ class MessageParserTests: XCTestCase {
         XCTAssertNotNil(firstMessage)
         
         let expectedPointCount = 36
-        let actualFirstMessagePointCount = firstMessage!.points.count
+        let actualFirstMessagePointCount = firstMessage?.points.count ?? 0
         XCTAssertEqual(actualFirstMessagePointCount, expectedPointCount)
-        
+
         let expectedMinimumAttributeCount = 7
-        let actualFirstMessageAttributeCount = firstMessage!.attributes.count
-        XCTAssertTrue(actualFirstMessageAttributeCount >= expectedMinimumAttributeCount)
+        let actualFirstMessageAttributeCount = firstMessage?.attributes.count ?? 0
+        XCTAssertGreaterThanOrEqual(actualFirstMessageAttributeCount, expectedMinimumAttributeCount)
     }
 }
--- a/arcgis-ios-sdk-samplesTests/Sample Tests/GraphicsOverlayDictionaryRenderer3DViewControllerTests.swift
+++ b/arcgis-ios-sdk-samplesTests/Sample Tests/GraphicsOverlayDictionaryRenderer3DViewControllerTests.swift
@@ -42,18 +42,36 @@ class GraphicsOverlayDictionaryRenderer3DViewControllerTests: XCTestCase {
 }
 
 class MessageParserTests: XCTestCase {
-    let messagesURL = Bundle.main.url(forResource: "Mil2525DMessages", withExtension: "xml")!
-    
     func testParseMessagesAtURL() {
-        let parser = MessageParser()
-        do {
-            let messagesData = try Data(contentsOf: messagesURL)
-            let messages = try parser.parseMessages(from: messagesData)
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samplesTests/Sample Tests/GraphicsOverlayDictionaryRenderer3DViewControllerTests.swift
```diff
@@ -42,18 +42,36 @@ class GraphicsOverlayDictionaryRenderer3DViewControllerTests: XCTestCase {
...
+        
+        let expectedMessageCount = 33
+        let actualMessageCount = messages.count
+        XCTAssertEqual(actualMessageCount, expectedMessageCount)
+
+        let firstMessage = messages.first
+        XCTAssertNotNil(firstMessage)
+        
+        let expectedPointCount = 36
+        let actualFirstMessagePointCount = firstMessage!.points.count
```
philium: Assuming there are no messages, force unwrapping `firstMessage` would cause a crash.

```suggestion
        let actualFirstMessagePointCount = firstMessage?.points.count
```

arcgis-ios-sdk-samplesTests/Sample Tests/GraphicsOverlayDictionaryRenderer3DViewControllerTests.swift
```diff
@@ -42,18 +42,36 @@ class GraphicsOverlayDictionaryRenderer3DViewControllerTests: XCTestCase {
...
+        XCTAssertEqual(actualFirstMessagePointCount, expectedPointCount)
+        
+        let expectedMinimumAttributeCount = 7
+        let actualFirstMessageAttributeCount = firstMessage!.attributes.count
```
philium:
```suggestion
        let actualFirstMessageAttributeCount = firstMessage?.attributes.count ?? 0
```

arcgis-ios-sdk-samplesTests/Sample Tests/GraphicsOverlayDictionaryRenderer3DViewControllerTests.swift
```diff
@@ -42,18 +42,36 @@ class GraphicsOverlayDictionaryRenderer3DViewControllerTests: XCTestCase {
...
+        XCTAssertTrue(actualFirstMessageAttributeCount >= expectedMinimumAttributeCount)
```
philium:
```suggestion
        XCTAssertGreaterThanOrEqual(actualFirstMessageAttributeCount, expectedMinimumAttributeCount)
```

arcgis-ios-sdk-samplesTests/Sample Tests/GraphicsOverlayDictionaryRenderer3DViewControllerTests.swift
stevebaranski: I introduced a call to `XCTAssertNotNil(firstMessage)` on line 67 to account for the fact that `firstMessage` could be `nil`. These assertions occur after that potential failure.

stevebaranski: You're absolutely right, however, that the test could still crash post-assertion if the XML file or the parser were to change subsequently. I'll tidy this up.

philium: That is all true. However, we're not setting `continueAfterFailure` and the default value is `true`, so this line would be reached, even if the assertion on line `67` failed. Apple introduced `XCTUnwrap` to solve this problem, but it is only available when using Xcode 11. I'll send you a Xcode 10 friendly implementation you can use if you'd like to go that route.

stevebaranski:
@philium thanks for the valuable input. I opted _not_ to proceed with `XCTUnwrap` in this case, pushing an additional commit that should address each element of your feedback.

I believe this is ready for another look.

PR #748, opened 2019-08-08
Remove GPKGLayerTableCell

philium:
And use a basic style cell instead.  The cell was being used to map index paths to layers, but that is the responsibility of the table view's data source.


stevebaranski:
I took the branch for a spin, and can confirm that it builds and existing tests pass. I understand the rationale, and the proposed changes look good to me. 

The only question that comes to mind is as follows : can we also remove both the `GPKGLayerTableCell.swift` file and Xcode project reference?

philium:
@stevebaranski No, thank you!

PR #749, opened 2019-08-11
Improve SwiftLint config

philium:
Removes unrecognized rule (`unused_private_declaration`) and analyzer rule in wrong place (`unused_import`). The analyzer rule may be reintroduced in the right place at a later date.


philium:
@rolson Thank you!

PR #750, opened 2019-08-12
Vivian edit sync v next

vquach2404:


arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,265 @@
...
+        if self.selectedFeature != nil {
+            let point = mapView.screen(toLocation: screenPoint)
+            if AGSGeometryEngine.geometry(point, intersects: areaOfInterest) {
+                selectedFeature.geometry = point
+                selectedFeature.featureTable?.update(selectedFeature) { [weak self] (error: Error?) in
+                    if let error = error {
+                        self!.presentAlert(error: error)
+                    } else {
+                        self!.syncButton.isHidden = false
+                        self!.syncButton.isEnabled = true
```
philium: In general, repeatedly accessing a `weak` value is not a good idea since the value could become `nil` in between accesses to it. We know that can't happen here, but `self` is also being force unwrapped multiple times. Instead, please either unwrap `self` at the top of the closure or extract the contents of the closure into a method so that `self` only needs to be referenced once (to call the method).

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,265 @@
...
+                    }
+                }
+            } else {
+                self.presentAlert(title: "Cannot move feature outside downloaded area")
+            }
+        } else { // Identify which feature was tapped and select it.
+            mapView.identifyLayers(atScreenPoint: screenPoint, tolerance: 22.0, returnPopupsOnly: false, maximumResultsPerLayer: 1) { (results: [AGSIdentifyLayerResult]?, error: Error?) in
+                if let error = error {
+                    self.presentAlert(error: error)
+                } else {
```
philium: I would suggest optionally unwrapping `results` here so as not to force unwrap it later:

```swift
                } else if let results = results {
```

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,265 @@
...
+                    self.tapFeatureLabel.isHidden = true
+                    self.moveFeatureLabel.isHidden = false
+                    self.tapSyncLabel.isHidden = true
+                    if results!.isEmpty == false {
```
philium: I would suggest combining this line and the next line:

```swift
if let result = results.first {
    ...
}
```

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,265 @@
...
+                        let firstResult = results!.first
+                        let layerContent = firstResult!.layerContent
+                        
+                        // Check that the result is a feature layer and has elements.
+                        if layerContent.isKind(of: AGSFeatureLayer.self) && firstResult!.geoElements.isEmpty == false {
```
philium: Please use optional binding for the layer and the geo element:

```swift
if let layer = layerContent as? AGSFeatureLayer, let feature = result.geoElements.first as? AGSFeature {
    ...
}
```

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeatures.storyboard
```diff
@@ -0,0 +1,114 @@
...
+                                <color key="backgroundColor" red="0.80479327282929791" green="0.80479327282929791" blue="0.80479327282929791" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                <color key="tintColor" white="0.33333333333333331" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                                <state key="normal" title="Generate Geodatabase"/>
+                                <connections>
+                                    <action selector="generateGeodatabase" destination="x5g-SP-gLy" eventType="touchUpInside" id="cfm-H4-NDS"/>
+                                </connections>
+                            </button>
+                            <label hidden="YES" opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Tap on a feature" textAlignment="center" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="Ukm-A2-tL3" userLabel="Tap on a Feature Label">
```
philium: Rather than having a separate label per message, I would suggest using a single label and just updating its text as needed.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeatures.storyboard
```diff
@@ -0,0 +1,114 @@
...
+                                <rect key="frame" x="0.0" y="44" width="414" height="21"/>
+                                <color key="backgroundColor" white="0.33333333329999998" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                                <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                <color key="textColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                                <nil key="highlightedColor"/>
+                            </label>
+                            <button hidden="YES" opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="roundedRect" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="rvk-82-DQP" userLabel="Sync Button">
```
philium: The button doesn't respect the safe area:

image

However, floating buttons like this look a little out of place on iOS. So in many of the other samples, we put buttons in a toolbar at the bottom.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,265 @@
...
+        // Add the source code button item to the right of navigation bar.
+        (self.navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["EditAndSyncFeaturesViewController"]
+    }
+}
+
+// Allows the user to interactively select and move features on the map.
+extension EditAndSyncFeaturesViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        // Move the selected feature to the tapped location and update it in the feature table.
+        if self.selectedFeature != nil {
```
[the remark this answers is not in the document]
vquach2404: I tried to implement this suggestion but it didn't fix the force unwrapping issue and also created a new variable that I had no use for in that closure. So I left it as is and removed the`[weak self]` in the later closure to resolve the force unwrapping issue. Please let me know if there is a preferred, alternate solution.

vquach2404:
Thanks for the feedback! Ready for a re-review.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,258 @@
...
+            let point = mapView.screen(toLocation: screenPoint)
+            if AGSGeometryEngine.geometry(point, intersects: areaOfInterest) {
+                selectedFeature.geometry = point
+                selectedFeature.featureTable?.update(selectedFeature) { (error: Error?) in
```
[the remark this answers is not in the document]
vquach2404: I see, I thought I needed to access `selectedFeature` in order for the function to work. But you were right, I can use feature and it works just as well.

vquach2404:
Made the requested changes, ready for a re-review.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeatures.storyboard
```diff
@@ -0,0 +1,104 @@
...
+                                    <barButtonItem style="plain" systemItem="flexibleSpace" id="RT5-mr-U1C"/>
+                                    <barButtonItem title="Generate Geodatabase" id="1os-c0-2Eg" userLabel="Generate Bar Button Item">
+                                        <connections>
+                                            <action selector="generateGeodatabase" destination="x5g-SP-gLy" id="WYG-N0-KcY"/>
+                                        </connections>
+                                    </barButtonItem>
+                                    <barButtonItem style="plain" systemItem="flexibleSpace" id="Sb4-0b-W7L"/>
+                                </items>
+                            </toolbar>
+                            <toolbar hidden="YES" opaque="NO" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="xXA-zf-wRd" userLabel="Sync Toolbar">
```
philium: Just like you're reusing the a single label, could you reuse a single toolbar with a single bar button item? You should be able to just change the title of the bar button item.

Also, at least one of the toolbars is not extending into the safe area:

image

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeatures.storyboard
```diff
@@ -0,0 +1,104 @@
...
+                            </view>
+                            <view userInteractionEnabled="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="r5p-5s-jhB" userLabel="Extent View">
+                                <rect key="frame" x="72.5" y="157" width="269" height="582"/>
+                                <color key="backgroundColor" white="0.0" alpha="0.0" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                                <accessibility key="accessibilityConfiguration">
+                                    <accessibilityTraits key="traits" notEnabled="YES"/>
+                                </accessibility>
+                            </view>
+                            <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="" textAlignment="center" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="HcO-gi-Dd8" userLabel="Instructions Label">
```
philium: The label doesn't extend into the safe area either:

image

vquach2404: I'm not sure what's wrong — I have alignment constraints that make the label leading and trailing equal to the safe area leading and trailing respectively. Is there something else I need to change?

philium: In this case I would align the label's leading and trailing edges to the leading and trailing edges of its superview, not the safe area. We want the label's background to extend edge-to-edge.

vquach2404:
Ready for another review. Also @sbaskaran please review 🙂 thanks!

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,261 @@
...
+        didSet {
+            // Initialize map with a basemap.
+            let tileCache = AGSTileCache(name: "SanFrancisco")
+            let tiledLayer = AGSArcGISTiledLayer(tileCache: tileCache)
+            let map = AGSMap(basemap: AGSBasemap(baseLayer: tiledLayer))
+            // Assign the map to the map view.
+            mapView.map = map
+            
+            // Create a geodatabase sync task using the feature service URL.
+            let featureServiceString = "https://sampleserver6.arcgisonline" + ".com/arcgis/rest/services/Sync/WildfireSync/FeatureServer"
```
sbaskaran: @vquach2404 wonder why it's concatenation of two strings here? 

I was positive I left a comment on this last week, but can't find it now. Wonder if not adding it as part of a review had to do anything with it.

vquach2404: Looking back now, I'm not sure. I may have just done that because this sample was modeled after Java and that's how they also implemented it. I will remove the concatenation if it doesn't cause any errors.

vquach2404:
> @philium @vquach2404 What is the purpose of this sample when we already have `Offline edit and sync` which does almost same things and bit more. They both use same data too
> 

@saratchandrakarumuri Thanks for your feedback, I have made changes accordingly. Regarding your question about the purpose of this sample, this question was addressed in the sample design description: "this is more complex with the popup and multiple layers, and some SDK teams do not yet have an editable popup. It would be good to show a very simple sample that takes a single point layer offline, allows you to tap on the feature, then tap somewhere else to move its geometry, then a sync button will become enabled after editing a feature."

vquach2404:
@saratchandrakarumuri I've implemented changes. Please re-review! Thank you!

saratchandrakarumuri:
> Selection is not rendering on selecting the first feature. It works as expected for all eventual selections  

All the issues except for this are fixed. I am still able to repro this

vquach2404:
> All the issues except for this are fixed. I am still able to repro this

@saratchandrakarumuri  do you mean the first feature you select is not highlighted? And what steps do you take to repro this?

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+    // Clears selection in all layers of the map.
+    private func clearSelection() {
+        for layer in mapView.map!.operationalLayers {
+            if let layer = layer as? AGSFeatureLayer {
+                layer.clearSelection()
+            }
+        }
+    }
+    
+    private func didMove() {
```
nixta: Hi @vquach2404. I would rename this to something like `resetUIState()` or similar (I haven't checked the logic in depth - there might be a much better description of what this does). But as discussed, `didMove()` doesn't describe what's moved in the context of the class it's on. Since it doesn't itself manipulate the feature or really do anything with the feature explicitly, it's better to describe what it does for the sample's workflow.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+        self.selectedFeature = nil
+        self.barButtonItem.isEnabled = true
+        self.barButtonItem.title = "Sync geodatabase"
+        self.instructionsLabel.text = String("Tap the sync button")
```
nixta: You do not need the explicit `String()` constructor here. Just setting the label to "Tap the sync button" as a literal will work.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+    private func clearSelection() {
+        for layer in mapView.map!.operationalLayers {
+            if let layer = layer as? AGSFeatureLayer {
+                layer.clearSelection()
+            }
+        }
+    }
+    
+    private func didMove() {
+        self.selectedFeature = nil
```
nixta: All of these do not need the `self.` prefix. When you call this function with `self?.didMove()` Swift is already taking care of the reference to `self`.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+                    } else {
+                        // Create a new feature layer from the table and add it to the map.
+                        let featureLayer = AGSFeatureLayer(featureTable: geodatabaseFeatureTable)
+                        self?.mapView.map?.operationalLayers.add(featureLayer)
+                    }
+                }
+            }
+            self.generateJob = nil
+            self.barButtonItem.isEnabled = false
+            self.instructionsLabel.text = String("Tap on a feature")
```
nixta: Again, please use the direct literal in place of the explicit constructor.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+            if let error = error {
+                self?.presentAlert(error: error)
+            } else {
+                self?.geodatabaseDidSync()
+            }
+        })
+    }
+    
+    @IBAction func generateOrSync() {
+        if barButtonItem.title == "Sync geodatabase" {
```
nixta: Using the string to determine state is somewhat fragile.

Ideally, one might define the state of the app with either a boolean or an enum, and just reflect the state with a string rather than using the string itself to define the state.

However, in this case I would be Ok with it if you defined the "Sync geodatabase" string as a `private let` file level or class level let, and then replace the instances of "Sync geodatabase" with a reference to that.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+                    self?.presentAlert(error: error)
+                } else if let results = results {
+                    self?.instructionsLabel.text = String("Tap on the map to move the feature")
+                    if let firstResult = results.first {
+                        let layerContent = firstResult.layerContent
+                        
+                        // Check that the result is a feature layer and has elements.
+                        if let featureLayer = layerContent as? AGSFeatureLayer, let feature = firstResult.geoElements.first as? AGSFeature {
+                            featureLayer.select(feature)
+                            self?.selectedFeature = feature
```
nixta: Having self.selectedFeature both represent application state AND reflect the map's state is a little fragile (and may be behind what @saratchandrakarumuri is seeing).

One option could be to use a `didSet` on `selectedFeature` to ensure that the map's selection matches the `selectedFeature` value/state.

Another would be to make `selectedFeature` a calculated property which works with the map's operationalLayers to determine what's selected, although the unknown there is what to do if more than one thing is selected.

saratchandrakarumuri:
> do you mean the first feature you select is not highlighted? And what steps do you take to repro this?  

@vquach2404 Just generate the geodatabase, select any feature. selection is not displayed for that feature. Once you move that a new location and then select another feature, now you can see the selection rendered as expected. I have attached a gif in previous comments showing this behavior

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
vquach2404: I don't think I'll be able to define it as a `private let` at file or class level because the title is set to "Generate geodatabase" in the storyboard. I believe setting it to "Sync geodatabase" at file/class level as a `private let` would not allow me to change the title when needed.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
vquach2404: I'm not sure what you mean by using a `didSet` on `selectedFeature`. Could you elaborate?

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
nixta: Hi @vquach2404: See https://github.com/Esri/arcgis-runtime-samples-ios/commit/6484386c4ed54d9caf01b0a8a011503b6bc1b322#diff-857df2a38391d55c8b3258543acb5541 to see what I mean.

If this is also set in the Storyboard, then this should really be redone as a boolean state variable, or an enum (though the enum would only have 2 options).

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
nixta: See https://github.com/Esri/arcgis-runtime-samples-ios/commit/8ac621fc804690c44a2be69ba6646325dd7cfb05#diff-857df2a38391d55c8b3258543acb5541 for details.

You can execute a `didSet` closure on a variable whenever that variable is updated. In this case I've consolidated all the logic which selects a feature in the map view to the `didSet` on `selectedFeature`, which gives just one entry point to manage both the sample's variable that tracks the selected feature AND the updating of the map view to reflect this.

nixta: @saratchandrakarumuri if you try with branch [nixta/suggestions-for-pr-750](https://github.com/Esri/arcgis-runtime-samples-ios/tree/nixta/suggestions-for-pr-750) do you see the same behavior?

saratchandrakarumuri: @nixta Your branch fixes the behavior I am seeing. First feature is selected as expected

vquach2404:
@nixta I just merged your PR and made a small change to the name of the button title variable. I appreciate your help, thank you!

vquach2404:
@sbaskaran @nixta Please review once more so I can merge after getting everyone's approval. Thank you 🙂

arcgis-ios-sdk-samples/Edit data/Edit and sync features/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+![Edit and sync features](edit-and-sync-features.png)
+
+## Use case
+
+A survey worker who works in an area without an internet connection could take a geodatabase of survey features offline at their office, make edits and add new features to the offline geodatabase in the field, and sync the updates with the online feature service after returning to the office.
+
+## How to use the sample
+
+Pan and zoom to position the red rectangle around the area to be taken offline. Tap "Generate geodatabase" to take the area offline. To edit features, tap to select a feature, and tap again anywhere else on the map to move the selected feature to the clicked location. To sync the edits with the feature service, tap the "Sync geodatabase" button.
```
sbaskaran: clicked location -> tapped location

arcgis-ios-sdk-samples/Edit data/Edit and sync features/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## How it works
+
+1. Create a geodatabase sync task from a URL to a feature service.
+2. Generate the geodatabase sync task to create geodatabase parameters by passing in an envelope extent as the parameter.
+3. Create a geodatabase job, passing in the parameters and a path to where the geodatabase should be downloaded locally.
+4. Start the job and get a geodatabase as a result.
+5. Load the geodatabase and get its feature tables. Create feature layers from the feature tables and add them to the map's operational layers collection.
+6. Set the sync direction.
```
sbaskaran: Between points 5 and 6, should there be a mention of the local edits? (Point 8 talks about the edits)

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,280 @@
...
+                self.generateJob = generateGeodatabaseJob
+                generateGeodatabaseJob.start(
+                    statusHandler: { (status: AGSJobStatus) in
+                        SVProgressHUD.show(withStatus: status.statusString()) //Show job status.
+                    },
+                    completion: { [weak self] (_, error: Error?) in
+                        SVProgressHUD.dismiss()
+                        
+                        if let error = error {
+                            self?.presentAlert(error: error)
```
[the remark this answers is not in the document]
nixta: A sensible approach which is now supported in Swift is to do this:

``` Swift
```

From that point on you have a strong non-optional reference to `self`, having confirmed it has not been deallocated.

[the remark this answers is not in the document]
philium: @sbaskaran If we only reference `weak self` once in any given path through the closure, it is fine to just use optional chaining. Otherwise we unwrap it first using the pattern @nixta showed above.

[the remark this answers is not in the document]
sbaskaran: @philium I agree. I've given that proposal at another place in the PR. My suggestion here is to be consistent. Optional chaining is fine, I noticed `self!` somewhere, maybe change that?

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,280 @@
+//
+// Copyright © 2019 Esri.
```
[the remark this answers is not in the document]
vquach2404: This seems to be present in most of the newer samples. Is there someone who can confirm?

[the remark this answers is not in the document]
nixta: The copyright notice _should_ be there. It should reflect when the sample was first added (that is, we do not subsequently update it).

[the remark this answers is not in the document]
sbaskaran: @nixta should the copyright _symbol_ © be there?

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,280 @@
...
+        // Hide the unnecessary items.
+        barButtonItem.isEnabled = false
+        extentView.isHidden = true
+        
+        // Get the area outlined by the extent view.
+        areaOfInterest = self.extentViewFrameToEnvelope()
+        
+        geodatabaseSyncTask.defaultGenerateGeodatabaseParameters(withExtent: areaOfInterest) { [weak self] (params: AGSGenerateGeodatabaseParameters?, error: Error?) in
+            if let params = params,
+                let self = self {
```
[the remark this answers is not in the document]
vquach2404: I believe I did that to unwrap `params`. Which method (`guard` and unwrapping params twice with `?` or leaving as is unwrapping once at the top) is preferred?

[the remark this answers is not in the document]
nixta: In a case like this, I would recommend this:

``` Swift
if let error = error {
  // handle the error in the normal way
}

guard let params = params else { return }
```

That way the bulk of the method is not indented too far. This helps with readability too as you're using `guard` to explicitly explain what's important for the sample to continue. And this `guard` statement will `unwrap` params into a non-optional, which is a good thing.

As @philium mentioned elsewhere, if you're just using an optional with inline chaining like `self?...` for calling a function or two, it's ok, but in general for anything longer, like this, (or if you'll be passing that optional in to other functions which don't accept an optional) then you should use an `if let` or `guard let` to unwrap them. I prefer a `guard let` in a case like this because of the reasons given above.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,280 @@
...
+                    }
+                )
+            } else {
+                self!.presentAlert(title: "Could not generate default parameters: \(error!)")
+            }
+        }
+    }
+    
+    func syncGeodatabase() {
+        barButtonItem.isEnabled = false
```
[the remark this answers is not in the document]
vquach2404: I just tried this method but I don't think it will work because the geodatabase does not yet exist upon the first instance of generating the geodatabase.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/README.md
vquach2404: Local edits are now mentioned in step 6. I tried to phrase it while still addressing a developer as the audience. Let me know what you think!

sbaskaran: Did you push a commit? I don't see it

vquach2404: @sbaskaran Sorry, just pushed!

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -158,43 +161,40 @@ class EditAndSyncFeaturesViewController: UIViewController {
...
-                            self?.presentAlert(error: error)
-                        } else {
-                            // Load the geodatabase when the job is done.
-                            self?.geodatabase = generateGeodatabaseJob.result
-                            self?.geodatabase.load { [weak self] (error: Error?) in
-                                self?.geodatabaseDidLoad()
-                            }
+            guard let self = self else { return }
+            
+            guard let params = params else { return }
```
sbaskaran: I think you'd want to present the alert here?

arcgis-ios-sdk-samples/Edit data/Edit and sync features/README.md
```diff
@@ -10,16 +10,16 @@ A survey worker who works in an area without an internet connection could take a
...
 
 1. Create a geodatabase sync task from a URL to a feature service.
 2. Generate the geodatabase sync task to create geodatabase parameters by passing in an envelope extent as the parameter.
 3. Create a geodatabase job, passing in the parameters and a path to where the geodatabase should be downloaded locally.
 4. Start the job and get a geodatabase as a result.
-5. Load the geodatabase and get its feature tables. Create feature layers from the feature tables and add them to the map's operational layers collection.
-6. Set the sync direction.
+5. Set the sync direction.
+6. To enable editing, load the geodatabase and get its feature tables. Create feature layers from the feature tables and add them to the map's operational layers collection.
 7. Create geodatabase job passing in the parameters and geodatabase as arguments.
```
sbaskaran: Can you mention `sync job` just to make it really clear from generate geodatabase job?

sbaskaran:
Couple more comments, @vquach2404

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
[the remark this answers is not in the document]
sbaskaran: why would the sync method be called before generating geodatabase?

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -0,0 +1,280 @@
...
+        
+        // Return the envenlope covering the entire extent frame.
+        return AGSEnvelope(min: minPoint, max: maxPoint)
+    }
+    
+    private func addFeatureLayers() {
+        // Iterate through the layers in the service.
+        geodatabaseSyncTask?.load { [weak self] (error) in
+            if let error = error {
+                print("Could not load feature service \(error)")
```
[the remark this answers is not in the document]
sbaskaran: can you please address this as well?

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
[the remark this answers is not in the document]
sbaskaran: @vquach2404 you are right, there are samples with and without the symbol. Let's leave it as-is

vquach2404:
@sbaskaran implemented those changes!

ldanzinger:
@sbaskaran is this good to merge?

sbaskaran:
@ldanzinger There are a couple of open questions. I've marked others as Resolved.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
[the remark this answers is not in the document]
philium: The Sync button doesn't become enabled until after a feature has been moved, so checking for whether the geodatabase has local edits would be superfluous.

PR #765, opened 2019-08-29
Enabled additional SwiftLint rule : trailing_closure

stevebaranski:
### Context

This PR proposes enabling the `trailing_closure` SwiftLint rule, and appeases the existing warnings. [The rule](https://github.com/realm/SwiftLint/blob/master/Rules.md#trailing-closure) is _not_ enabled by default, unlike the rules for [`multiple_closures_with_trailing_closure`](https://github.com/realm/SwiftLint/blob/master/Rules.md#multiple-closures-with-trailing-closure) & [`empty_parentheses_with_trailing_closure`](https://github.com/realm/SwiftLint/blob/master/Rules.md#empty-parentheses-with-trailing-closure).

### Impetus

This was rightly called out in a [recent PR review](https://github.com/Esri/arcgis-runtime-samples-ios/pull/758#discussion_r315765353), but also overlooked in a [separate review](https://github.com/Esri/arcgis-runtime-samples-ios/pull/751/files#diff-69ab99ba03b38a0a6688fc7ec682b742R123). I thought it would be great to promote some additional consistency.

### Notes

- This does not need to be part of the 100.6 release.
- I smoke-tested the impacted samples, and noted no ill effects.

PR #766, opened 2019-08-29
Enabled additional SwiftLint rule : unused_closure_parameter

stevebaranski:
### Context

This PR proposes enabling the `unused_closure_parameter` SwiftLint rule, appeases the existing warnings. SwiftLint enables [this rule](https://github.com/realm/SwiftLint/blob/master/Rules.md#unused-closure-parameter) by default, but it is currently explicitly disabled in this project, as observed in this recent [PR review](https://github.com/Esri/arcgis-runtime-samples-ios/pull/758#discussion_r317627980).

### Notes

- Because this change is being proposed in parallel with another closure-related SwiftLint [change](https://github.com/Esri/arcgis-runtime-samples-ios/pull/765), I have expressed this PR against _that_ base branch (i.e., `sab/lint-trailing_closure`) for review clarity. While this _could_ be merged into that branch prior to merging #765 into `v.next`, we should probably just change the base branch of this PR (i.e., to `v.next`) once that PR has been merged.
- This does not need to be part of the 100.6 release.
- I smoke-tested the impacted samples, and noted no ill effects.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -25,7 +25,6 @@ disabled_rules: # rule identifiers to exclude from running
   - trailing_whitespace
   - type_body_length
   - type_name
-  - unused_closure_parameter
 
 opt_in_rules: # some rules are only opt-in
   - anyobject_protocol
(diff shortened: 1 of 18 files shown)


stevebaranski:
As discussed offline, this rule was previously considered for the [macOS sample project](https://github.com/Esri/arcgis-runtime-samples-macos/pull/240).

That review notes that the `_` might obscure the API in some cases. If you find that to be the case for some of the closures amended here, then perhaps we can consider them on a case-by-case basis during this review? For example, we can add a comment explaining how / why some of those were unused in the specific sample.

PR #778, opened 2019-09-30
Added instructions for cocoapods

dg0yal:
- Minor wording update to the top paragraph. Added image of sample app from app store.
- Added instructions for using the installed SDK
- Added instructions and podfile for cocoapods


dg0yal:
Good call @philium

PR #780, opened 2019-10-02
Integrating toolkit into samples repo

dg0yal:
This is in prep for the upcoming AR samples that will need the toolkit.

The toolkit repo has been added as a submodule. This will help users who use the installed sdk.

For those using cocoapods, a podfile has been added.

Note - We will need to update the podfile and the submodule to point to 100.6 toolkit when it is released.

Currently I am referencing 100.5 version of the toolkit.

@mhdostal @philium


podfile
```diff
@@ -1,4 +1,4 @@
 platform :ios, '11.0'
 target 'ArcGIS Runtime SDK Samples' do
-  pod 'ArcGIS-Runtime-SDK-iOS', '~> 100.6'
+  pod 'ArcGIS-Runtime-Toolkit-iOS', '~> 100.5'
```
mhdostal: This will need to be updated to 100.6 once the Toolkit is released.

dg0yal: Yup

dg0yal: And also the submodule reference

dg0yal:
@philium done

PR #784, opened 2019-10-05
Open a Scene (Portal Item): Fix scene view constraints

philium:
Constrains the scene view's leading and trailing edges to the superview rather than the safe area so that the scene extends into the safe area.


philium:
@njarecha Please review at your earliest convenience.

Issue #777, opened 2019-09-30
Update app for iOS 13


esreli:
The iOS samples app has not yet updated for iOS 13. Some areas to consider include:

### [UINavigationBarAppearance](https://developer.apple.com/documentation/uikit/uinavigationbarappearance)

On iOS 13, the navigation bar is unusable.

<img width="475" alt="Screen Shot 2019-09-30 at 9 38 01 AM" src="https://user-images.githubusercontent.com/33433113/65898577-bab36880-e366-11e9-92ce-e521fc51456f.png">

### Dark Mode

A horizontal spacing between cells is not visible.

<img width="474" alt="Screen Shot 2019-09-30 at 9 46 55 AM" src="https://user-images.githubusercontent.com/33433113/65898852-46c59000-e367-11e9-9556-69684d4fe42f.png">

dg0yal:
We will be fixing the nav bar problems asap.
But won't be adding support for dark mode until later

esreli:
I've identified another element for consideration in this effort due to a redesign in iOS 13, the `UISegmentControl`. 

![Screen Shot 2019-10-04 at 1 13 20 PM](https://user-images.githubusercontent.com/33433113/66237139-31bf6880-e6a9-11e9-93d0-e62125cbb50f.png)

esreli:
I've identified another UI element that needs design consideration for iOS 13, the nav bar's search bar.

<img width="393" alt="Screen Shot 2019-10-10 at 10 18 22 AM" src="https://user-images.githubusercontent.com/33433113/66591321-80588100-eb47-11e9-995b-6aacb3bcda30.png">

rmuellerjr:
Using IOS 13 with iPhone Pro, trying to use "Edit and sync" via Sample SDK app. When i open the sample, I see "Create geodatabase". When click create button app closes without error.

yo1995:
Also, do we consider update the URL atop the repo to https://developers.arcgis.com/ios/latest/ ?

![image](https://user-images.githubusercontent.com/9660181/80522171-fdf7a900-8940-11ea-86b2-73b516c11261.png)

yo1995:
Update the README and above to reflect changes in 100.8

- [ ] `Requirements` section
- [ ] README versions
- [ ] Podfile dependency version
- [ ] `Building samples using installed SDK` section
- [ ] URLs in `Licensing` and `Contributing` sections are not working.

Jaseven777:
[...]

PR #785, opened 2019-10-08
Update navigation bar appearance for iOS 13

philium:
This fixes and issue where the background color and tint color were the same. Also disabled dark mode as we don't support it yet.


esreli:
I've identified two more UI elements that will need consideration re: iOS 13. 

https://github.com/Esri/arcgis-runtime-samples-ios/issues/777#issuecomment-538547048
https://github.com/Esri/arcgis-runtime-samples-ios/issues/777#issuecomment-540686516

PR #787, opened 2019-10-09
Update master with V.next

dg0yal:
AR samples & iOS 13 nav bar fix


philium:
@dg0yal You don't want any of the changes in the two open PRs to be part of this release?

dg0yal:
Those can come later. I do not want to delay the AR samples any longer.
They were supposed to go out last week

philium:
Hold on. There's one issue that Suganya and I have seen this morning that needs to be fixed. Cloning (including pulling the submodule) and building results in the following error:

image

Thankfully the fix is trivial. I'll put up a PR.

PR #805, opened 2019-11-14
Make Xcode happy

philium:
Xcode keeps suggesting this change whenever a change is made to the project file.


philium:
Yes, it still builds. The identifier `4C11DBA2234D05C00040EC1C` only appears on this one line, so it appears to be vestigial.

PR #807, opened 2019-12-01
Add Apply Scheduled Updates to Preplanned Map Area sample

philium:


arcgis-ios-sdk-samples/Maps/Apply scheduled updates to preplanned map area/ApplyScheduledUpdatesToPreplannedMapAreaViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+            try FileManager.default.copyItem(at: mobileMapPackageURL, to: temporaryMobileMapPackageURL)
+            mobileMapPackage = AGSMobileMapPackage(fileURL: temporaryMobileMapPackageURL)
+        } catch {
+            print("Error setting up mobile map package: \(error)")
+            mobileMapPackage = nil
+        }
+        
+        super.init(coder: coder)
+        
+        mobileMapPackage?.load { [weak self] (error) in
```
mhdostal: `mobileMapPackage` is defined as non-optional.  Does the compiler give a warning if you leave off the `?`?

arcgis-ios-sdk-samples/Maps/Apply scheduled updates to preplanned map area/ApplyScheduledUpdatesToPreplannedMapAreaViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+    }
+    
+    /// Called in response to the mobile map package load operation completing.
+    /// - Parameter result: The result of the load operation.
+    func mobileMapPackageDidLoad(with result: Result<Void, Error>) {
+        switch result {
+        case .success:
+            let map = self.mobileMapPackage.maps.first!
+            self.loadViewIfNeeded()
+            self.mapView.map = map
```
mhdostal: Do you need to use `self.` in this line and the two previous lines?

arcgis-ios-sdk-samples/Maps/Apply scheduled updates to preplanned map area/ApplyScheduledUpdatesToPreplannedMapAreaViewController.swift
philium: It is defined as an implicitly unwrapped optional (IUO). The only way it will be `nil` is if creation of the temporary copy fails, here in the initializer. So we need to use optional chaining on the off chance that happened. We don't need to use optional chaining elsewhere as the only way that code can be reached is if the mobile map package was loaded and therefore exists. Does that make sense?

arcgis-ios-sdk-samples/Maps/Apply scheduled updates to preplanned map area/ApplyScheduledUpdatesToPreplannedMapAreaViewController.swift
philium: No. The references to `self` came from these lines originally being inside a closure. I've removed them.

philium:
Thank you @saratchandrakarumuri and @mhdostal !

PR #810, opened 2019-12-04
Sort categories and samples alphabetically by name

philium:


philium:
> Will sorting the categories and samples require redoing the screen shots in the Readme?

I don't know that it's a requirement, but it would be nice to update it. @dg0yal How did you get the image of screenshots for the README?

> Is there any other doc that needs to change because they're now sorted?

Not that I know of.

> Are all the other platforms currently sorting their categories/samples?

I don't know about all of them, but some of them do. iOS may be the only holdout.

dg0yal:
Please hold off on this.

PR #811, opened 2019-12-04
Update metadata files for renamed files

philium:
I forgot to do this along the way. I'll add this to my rename checklist.


philium:
@mhdostal Thank you!

PR #812, opened 2019-12-06
Copy READMEs into a single directory

philium:
Rather than putting each README in a separate directory, this puts them all in the same directory. This helps to clean up the app bundle.


yo1995:
cn

This is done as part of the new sample viewer. As this repo will be stabilized in the following month, what's your thoughts on applying the changes in this PR? Are there anything we missed that blocked this PR from being merged @vquach2404 thoughts are welcome.

I personally don't see any harm, and am in favor of keeping the bundle structure clean even if no user would see it. So I'm ok with merging this PR.

vquach2404:
Sounds good to me!

philium:
This PR never got reviewed, which is why it was never merged. At this point, I don't think it is worth spending any more time on this project than is necessary. Is there even another release planned?

I am glad to hear that this was implemented in the new sample viewer app!

vquach2404:
No release planned, we can close this issue.

yo1995:
Confirmed, no more major release in this repo so I'll do some housekeeping to close stale issues.

PR #813, opened 2019-12-13
Replace usage of AGSSurfacePlacement.draped with drapedBillboarded

philium:
The case `draped` has been deprecated.


philium:
Thank you @pgruenler and @njarecha!

PR #820, opened 2020-01-08
Surface Placements: Update camera to better emphasize draped flat

philium:
Before | After
------ | -----
Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-07 at 22 02 04 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-08 at 08 59 02


philium:
Thank you @njarecha!

PR #828, opened 2020-03-10
Search bar text field background color changed to differentiate from navigation bar

yo1995:
This PR solves a previous issue that causes search bar background color blend with the navigation bar.

Before: 
<p align="center"> Color difference - left iOS 12, right iOS 13</p>

After:
<p align="center"> Background color and cursor color changed - left iOS 12, right iOS 13</p>


arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -45,7 +45,12 @@ class ContentCollectionViewController: UICollectionViewController, UICollectionV
         searchBar.autocapitalizationType = .none
         
         if #available(iOS 13, *) {
-            // Do nothing, on iOS 13 the settings below have no effect.
+            // Change the backgroundColor to cope with #1517 and #1326
```
philium: Please remove these issue numbers as the issues themselves are not publicly visible.

PR #826, opened 2020-02-27
Display subtype feature layer for iOS

vquach2404:
Implemented sample and fixed SwiftLint warning in CutGeometryViewController.

PR #834, opened 2020-03-17
Add back the accidentally removed plist entry in the app

yo1995:
In #830 at [this commit](https://github.com/Esri/arcgis-runtime-samples-ios/pull/830/commits/f02101ca34441b2c6d10e69e74b4fbc7158c82ee), I removed the entry by accident. 

Add it back now, and would need to require @saratchandrakarumuri to do another check for the example in #826 , to make sure it is working correctly right now.

### DIFF
--- a/arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
+++ b/arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
@@ -289,6 +289,14 @@
 		<array>
 			<dict>
 				<key>displayName</key>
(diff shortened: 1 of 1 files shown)


yo1995:
Sorry for the trouble caused 😓 I should've taken another look at files before hitting on the merge button.

yo1995:
> I was about to look into that since I couldn't find this sample in our tests. Thanks for the fix

Great! I'll play it safe this time 😅 Please merge after you verify it's OK.

PR #835, opened 2020-03-17
Fix strong reference with weak capture in closures

yo1995:
Refactor previous code that does not add a `weak` keyword to the capture list, which could hold a strong reference to the ViewController itself.

### DIFF
--- a/arcgis-ios-sdk-samples/Analysis/Analyze hotspots/AnalyzeHotspotsViewController.swift
+++ b/arcgis-ios-sdk-samples/Analysis/Analyze hotspots/AnalyzeHotspotsViewController.swift
@@ -79,11 +79,7 @@ class AnalyzeHotspotsViewController: UIViewController, HotspotSettingsVCDelegate
         }, completion: { [weak self] (result: AGSGeoprocessingResult?, error: Error?) in
             //dismiss progress hud
             SVProgressHUD.dismiss()
-            
-            guard let self = self else {
-                return
-            }
-            
+            guard let self = self else { return }
             if let error = error {
                 //show error
                 self.presentAlert(error: error)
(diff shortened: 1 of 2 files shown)


arcgis-ios-sdk-samples/Analysis/Analyze hotspots/AnalyzeHotspotsViewController.swift
```diff
@@ -96,11 +92,14 @@ class AnalyzeHotspotsViewController: UIViewController, HotspotSettingsVCDelegate
...
-                    if error == nil {
+                (self.mapView.map?.operationalLayers.firstObject as! AGSLayer).load { [weak self] (error: Error?) in
+                    guard let self = self else { return }
+                    if let error = error {
+                        self.presentAlert(error: error)
+                    } else {
                         //set viewpoint as the extent of the mapImageLayer
                         if let extent = result?.mapImageLayer?.fullExtent {
-                            self.mapView.setViewpointGeometry(extent, completion: nil)
+                            self.mapView.setViewpointGeometry(extent)
```
philium: I know the temptation to clean up things as you go. I really do! But I've learned the hard way that it is better to leave unrelated things the way that they are than to change them and then find out later that the cleanup was not as harmless as I expected. I just want to caution you to be careful when making these sorts of changes. Thankfully this one is fine, but it isn't always. For example, take the `setViewpoint()` family of methods. The API defines one that takes a viewpoint and another that takes both a viewpoint and a completion handler:

```objc
- (void)setViewpoint:(AGSViewpoint*)viewpoint;
- (id<AGSCancelable>)setViewpoint:(AGSViewpoint*)viewpoint completion:(nullable void(^)(BOOL finished))completion;
```

The first method (without a completion handler) sets the viewpoint without animation. The second method (with a completion handler) sets the viewpoint with animation. The problem is, it is translated into Swift with a default value of `nil` for the `completion` parameter. So it would seem like omitting the `completion` parameter would just use the default value, but it actually changes to calling the first method. It causes a behavior change, going from animating the viewpoint change to not animating it.

Does that make sense? The rule I follow is this: when in doubt, hold off on unrelated cleanup changes and maybe submit a separate PR for them later. No PR is too small!

arcgis-ios-sdk-samples/Analysis/Analyze hotspots/AnalyzeHotspotsViewController.swift
```diff
@@ -96,11 +92,14 @@ class AnalyzeHotspotsViewController: UIViewController, HotspotSettingsVCDelegate
...
                         }
                     }
```
philium: A weak variable doesn't always need to be unwrapped with `guard`. If it is only ever accessed once for any given pathway, then it can just be unwrapped inline. But if it can be accessed repeatedly, then it should be unwrapped, since it is generally not good to repeatedly access a `weak` reference. So it would have been fine to use optional chaining both here.

arcgis-ios-sdk-samples/Analysis/Analyze hotspots/AnalyzeHotspotsViewController.swift
yo1995: 😨 😨 😨   
I see... I'll take smaller steps next time touching the things that I don't fully understand. Thanks for the detailed explanation!! 🙏

yo1995:
Double-checked the API and animation behavior after the change and it is correct. Next time will be more careful with those changes!
```swift
func setViewpointGeometry(_ geometry: AGSGeometry, completion: ((Bool) -> Void)? = nil) -> AGSCancelable
```

PR #837, opened 2020-03-20
Navigate route

yo1995:
Implementation of the `Navigate route` example. A brief demo below.

|Navigation.|Recenter to current navigation.|Reset to starting point.|
|------ | ------| ------|
||||

A link to current branch folder: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/NavigateRoute_vNext/arcgis-ios-sdk-samples/Route%20%26%20Directions/Navigate%20route


yo1995:
Keep a list of questions for coding style

1. Can I make the `routeTask` an `inout` variable to reduce instance properties?
2. Where should I call the solveRoute method, either in `viewDidLoad` or in `makeMap`
3. Where should I nullify the `speechSynthesizer`? In `viewWillDisappear` or somewhere else - what is the deinit logic in this app - I experimented and wonder where to look at.
4. When should I explicitly set variables to nil?

yo1995:
> 4. Add `AGSStop`s to the parameters `stops` collection for each destination.

This line is confusing.

yo1995:
Hi, I've a few questions listed above, and would love to have some input on this implementation! Feel free to let me know your thoughts!

I believe there will be more issues and things that I should be aware of for this example, comparing to the previous ones. So take your time to have a closer look at it. 😺

yo1995:
During the review meeting this morning it was mentioned that the navigation info label can be dark background with light text.   
While I think it makes sense, I'd like to know the common color scheme and the plan for migration to dark mode for the app.

vquach2404:
> During the review meeting this morning it was mentioned that the navigation info label can be dark background with light text.
> While I think it makes sense, I'd like to know the common color scheme and the plan for migration to dark mode for the app.

The Edit and sync features and Offline edit and sync samples are the only ones I know of that have  similar labels. Both of those use "Dark Gray" as the background color and "White" as the text color.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,296 @@
...
+                    mapView.setViewpoint(viewPoint, completion: nil)
+                } else {
+                    mapView.setViewpointGeometry(routeGeometry) { [weak self] _ in
+                        // Get the initial zoomed view point.
+                        self?.defaultViewPoint = self?.mapView.currentViewpoint(with: .centerAndScale)
+                    }
+                }
+            }
+        }
+    }
```
philium: Again, with a name like `setRouteGraphics(_:)`, I expected this method to take an array of graphics, just like `AGSRouteParameters.setStops(_:)` takes an array of stops. It also seems like it is trying to do too many things. Here are a few thoughts:

- One of the things this method does is create graphics and add them to the graphics overlay. When the route is reset, all the graphics are removed and then this method is called to recreate them. Unless I'm missing something, all of the stop graphics are always the same, so they don't need to be recreated every time. And `routeAheadGraphic` and `routeTraveledGraphic` are always the same object, just their geometries are updated. That got me thinking: could all of this be extracted into a method that creates the graphics, creates a graphics overlay and adds the graphics to it, and then returns the graphics overlay (`func makeRouteOverlay() -> AGSGraphicsOverlay`? Then, when the route is solved, the graphics overlay could be created and added to the map view, eliminating the need for the `graphicsOverlay` property.

- Another thing this method does is update the geometry of `routeAheadGraphic` and `routeTraveledGraphic`. Perhaps that could be extracted into a method that updates the route graphics? It could take a tracking progress: `func updateRouteGraphics(trackingProgress: AGSTrackingProgress)`.

- The third thing it does is zoom update the viewpoint. Perhaps that should be a third method.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,296 @@
...
+    
+    /// Resets to the starting location for location display.
+    func resetToStartingLocation() {
+        guard let initialLocation = mockDataSource?.locations?[0] else { return }
+        mockDataSource?.didUpdate(initialLocation)
+    }
```
philium: A few thoughts:

- This method is only being called from `reset()` and it doesn't really encapsulate anything, so I would just move the body of it into `reset()`.
- `if let` is being used to unwrap an optional, so there's no need to subscript into the `locations` array to get the first member. Just use `first` instead.
- `mockDataSource` is being unwrapped twice.

Putting those all together, this is what it might look like in `reset()`:

```swift
if let dataSource = mockDataSource, let location = dataSource.locations?.first {
    dataSource.didUpdate(location)
}
```

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,296 @@
...
+        mapView.locationDisplay.autoPanModeChangedHandler = { [weak self] _ in self?.recenterButtonItem.isEnabled = true }
+        // Start the location data source and location display.
+        mapView.locationDisplay.start(completion: nil)
+    }
+    
+    /// Called in response to the Reset button being tapped.
+    @IBAction func reset() {
+        // Stop the speech, if there is any.
+        speechSynthesizer?.stopSpeaking(at: .immediate)
+        speechSynthesizer = nil
```
philium: No need for this to be optional; just stopping the speech is fine.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,296 @@
...
+    }
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        // Nullify the properties to avoid lingering tasks running in the background.
+        super.viewWillDisappear(animated)
+        defaultViewPoint = nil
+        directionsList = nil
+        routeTracker = nil
+        mockDataSource = nil
+        speechSynthesizer = nil
```
philium: I don't think any of these should be unset when the view disappears. If the view disappears because the user opens the readme, all of these things would need to be recalculated/recreated again. If the view disappears because the sample is closed, as you pointed out, the view controller will get deallocated when another sample is opened.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,296 @@
...
+                self.presentAlert(error: error)
+            } else {
+                // Unwrap the AGSRouteParameters if there is no error.
+                guard let params = params else { return }
+                // Explicitly set values for parameters.
+                params.returnDirections = true
+                params.returnStops = true
+                params.returnRoutes = true
+                params.outputSpatialReference = .wgs84()
+                params.setStops(self.makeStops())
```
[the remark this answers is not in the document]
yo1995: This would be needed as it is called in an escaping closure for `AGSRouteTask.defaultRouteParameters(completion:)`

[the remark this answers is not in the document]
philium: Right. My mistake.

yo1995:
Hi Philip, I've addressed the issues mentioned above.

Also in terms of the README, I'm not quite sure the wording for delegate calls. I've appended my thoughts inline, and would like to have you edit them so that we can follow a norm for delegate methods in the `How it works` section. 🙂

philium:
@yo1995 Thanks for making those changes! I'll take another look sometime today.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,275 @@
...
+            if let error = error {
+                self.presentAlert(error: error)
+            } else if let params = params {
+                // Explicitly set values for parameters.
+                params.returnDirections = true
+                params.returnStops = true
+                params.returnRoutes = true
+                params.outputSpatialReference = .wgs84()
+                params.setStops(self.makeStops())
+                self.routeTask.solveRoute(with: params, completion: self.solveRouteCompletion(with:error:))
```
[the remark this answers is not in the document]
yo1995: Ahh my bad. Missed this one when refactoring that function. 😓

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,275 @@
...
+            }
+        }
+        // Add the source code button item to the right of navigation bar.
+        (self.navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["NavigateRouteViewController"]
+    }
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        // Stop the speech immediately.
+        speechSynthesizer.stopSpeaking(at: .immediate)
```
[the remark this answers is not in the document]
yo1995: Got it. That is also why I initially tried to set everything to `nil` when  `viewWillDisappear`, as I don't want anything that I don't understand run in the background when the VC is popped.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,275 @@
...
+        }
+    }
+    
+    func updateTrackingStatusDisplay(with status: AGSTrackingStatus) {
+        var statusText: String
+        switch status.destinationStatus {
+        case .notReached, .approaching:
+            let formatter = DateComponentsFormatter()
+            formatter.allowedUnits = [.hour, .minute, .second]
+            formatter.unitsStyle = .full
```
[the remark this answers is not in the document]
yo1995: Got it. Thanks for pointing that out! 😄 

Ref.: https://stackoverflow.com/questions/8832768/why-is-allocating-or-initializing-nsdateformatter-considered-expensive

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRoute.storyboard
```diff
@@ -0,0 +1,80 @@
...
+                                            <action selector="reset" destination="JxV-EA-CYk" id="6P0-Qg-upN"/>
+                                        </connections>
+                                    </barButtonItem>
+                                </items>
+                            </toolbar>
+                        </subviews>
+                        <color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
+                        <constraints>
+                            <constraint firstItem="EC7-F9-vN9" firstAttribute="top" secondItem="YVM-3B-TME" secondAttribute="top" id="2Vh-u4-Cba"/>
+                            <constraint firstItem="IhO-bQ-arF" firstAttribute="leading" secondItem="FzH-kw-6eq" secondAttribute="leading" id="Kxc-Jh-jye"/>
```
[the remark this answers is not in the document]
yo1995: Oops I omitted it as the label was added later. 😓   
I'll review the layout frames relationships to get a better understanding of how it looks like on a round-edge device.

Ref.: https://developer.apple.com/documentation/uikit/uiview/positioning_content_relative_to_the_safe_area. 
https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/adaptivity-and-layout/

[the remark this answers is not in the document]
yo1995: Wow that is cool. I don't know the app will become a split view on landscape for iPhone! Fixed.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/README.md
```diff
@@ -0,0 +1,62 @@
...
+* AGSRouteParameters
+* AGSRouteTask
+* AGSRouteTracker
+* AGSSimulatedLocationDataSource
+* AGSStop
+* AGSVoiceGuidance
+
+## Offline data
+
+None
```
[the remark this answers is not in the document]
yo1995: Removed.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRoute.storyboard
```diff
@@ -0,0 +1,80 @@
...
+                                            <action selector="reset" destination="JxV-EA-CYk" id="6P0-Qg-upN"/>
+                                        </connections>
+                                    </barButtonItem>
+                                </items>
+                            </toolbar>
+                        </subviews>
+                        <color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
+                        <constraints>
+                            <constraint firstItem="EC7-F9-vN9" firstAttribute="top" secondItem="YVM-3B-TME" secondAttribute="top" id="2Vh-u4-Cba"/>
+                            <constraint firstItem="IhO-bQ-arF" firstAttribute="leading" secondItem="YVM-3B-TME" secondAttribute="leading" id="Kxc-Jh-jye"/>
```
philium, line 55: The background now extends into the safe area, but so does the text. The text should be within the safe area, just like the attribution bar at the bottom of the map view:

image

That is why I suggested having two views: one for the background that is edge-to-edge and one for the text.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,284 @@
...
+            mapView.graphicsOverlays.add(makeRouteOverlay())
+        }
+    }
+    
+    /// A wrapper function for operations after the route is solved by an AGSRouteTask.
+    ///
+    /// - Parameters:
+    ///   - routeResult: The result from `AGSRouteTask.solveRoute(with:completion:)`.
+    ///   - error: The error from `AGSRouteTask.solveRoute(with:completion:)`.
+    func solveRouteCompletion(with routeResult: AGSRouteResult?, error: Error?) {
```
philium, line 67: This method name reads more like a command, like the method is going to be solving a route. Consider remaining it to something that removes that ambiguity, like `didSolveRoute`.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,284 @@
...
+    
+    /// Called in response to the Navigate button being tapped.
+    @IBAction func startNavigation() {
+        navigateButtonItem.isEnabled = false
+        resetButtonItem.isEnabled = true
+        mapView.locationDisplay.autoPanMode = .navigation
+        // If the user navigates the map view away from the location display, activate the recenter button.
+        mapView.locationDisplay.autoPanModeChangedHandler = { [weak self] _ in self?.recenterButtonItem.isEnabled = true }
+        // Start the location data source and location display.
+        mapView.locationDisplay.start(completion: nil)
```
philium, line 161:
```suggestion
        mapView.locationDisplay.start()
```

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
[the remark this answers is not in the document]
philium: Right. I misunderstood. Thanks for making that change!

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,284 @@
...
+            }
+        }
+        // Add the source code button item to the right of navigation bar.
+        (self.navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["NavigateRouteViewController"]
+    }
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        // Stop the speech immediately.
+        speechSynthesizer.stopSpeaking(at: .immediate)
```
philium, line 221: `reset()` calls `stopSpeaking(at:)`, so you don't need to call it here too.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRoute.storyboard
yo1995: Hmm, that is really a great point. I'll do a bit research on that. Also later I'll open another PR to fix the other examples with the same problem.

yo1995: For the Connected Trace example, it embeds the UILabel in a stack view (seems that you implemented that one). I think that would be a good practice to follow.

yo1995:
Just spotted and fixed that when there is no network connection, the example will crash because the reset function try to unwrap nil optionals.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRoute.storyboard
yo1995: To unify the styles in current examples, proposed #845 . That PR will make the `UILabel` extend into safe area. The text in the labels are not explicitly aligned to safe area, because their widths are short and are all center aligned.

yo1995:
> Sample looks good but when the license watermark is on upper right corner, directions labels overlaps with it and looks clumsy like this
> 
> IMG_4F265999193F-1
> 
> It is better to add some mapView inset like following to avoid the overlap
> 
> `mapView.contentInset.top`
> 
> IMG_D44C6F36AC08-1

Thanks for pointing that out! 😃  I also noticed that and didn't know how to solve it! I'll try with it and let you know when it is ready.

yo1995:
Set `mapView.contentInset.top` for the map so that its content won't overlap with the label text.

The inset was calculated to match the maximum height of the label. However, there is no easy way to calculate the real height of a label with line spacing, so I simply multiply the number of lines by the font height as an approximate.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -0,0 +1,287 @@
...
+        mapView.locationDisplay.autoPanMode = .navigation
+        recenterButtonItem.isEnabled = false
+    }
+    
+    // MARK: UIViewController
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Avoid the overlap between the label and the map content.
+        mapView.contentInset.top = CGFloat(statusLabel.numberOfLines) * statusLabel.font.lineHeight
```
philium, line 197: I didn't notice that the status label was given an explicit number of lines. While not wrong, that property is usually used for labels in a table view cell where you want to be able to assign an arbitrary string to a label and have the label only grow so big. Since that isn't a major concern here, I would expect the number of lines to be zero. As for the content inset, it could be updated dynamically based on the height of the label:

```swift
override func viewDidLayoutSubviews() {
    super.viewDidLayoutSubviews()
    // Avoid the overlap between the label and the map content.
    mapView.contentInset.top = statusLabel.frame.height
}
```

yo1995: iPhone SE

I set the `numberOfLines` because initially I want to limit the height of the `UILabel`, so that it won't cover too much content on a smaller screen such as iPhone SE. But for this particular example, I also agree that it's OK to leave it as 0.

yo1995: In terms of setting the inset on each turn, I also tried that, but found the UI is a bit clumsy. There's no transition animation between each turn if the inset is update every time. Check the gif below.

each-turn

philium: Yeah, that's unfortunately. I guess what you have is fine then.

PR #845, opened 2020-04-03
UILabel safe area fix

yo1995:
In the 3 AR examples, the info `UILabel`s' background does not extend into the safe area correctly. Also, the `ArcGISARView` the `Collect data AR` example overlap with the toolbar and labels incorrectly. This PR address those issues.

Below attached the screenshot comparisons between original and fixed versions.

## Collect data AR
|Original|After fix|
|------ | ------|
|||

## Display scenes in tabletop AR
|Original|After fix|
|------ | ------|
|||

## Explore scenes in flyover AR
|Original|After fix|
|------ | ------|
|||


yo1995:
Thanks for the review! I'll double check and merge afterwards.

PR #848, opened 2020-04-09
Fix: missing feature layer in Download preplanned map area sample

yo1995:
This PR fixes the issue that in `Download preplanned map area` sample, no feature layer is displayed for offline preplanned maps. This is due to the `AGSDownloadPreplannedOfflineMapParameters.updateMode` need to be set explicitly.

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Download preplanned map/MapSelectionTableViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Download preplanned map/MapSelectionTableViewController.swift
@@ -194,23 +194,30 @@ class MapSelectionTableViewController: UITableViewController {
             offlineTask = AGSOfflineMapTask(onlineMap: onlineMap)
         }
         
-        guard let task = offlineTask else { return }
-        
+        let task = offlineTask!
         try? FileManager.default.removeItem(at: area.mapPackageURL)
-        
-        let job = task.downloadPreplannedOfflineMapJob(with: area, downloadDirectory: area.mapPackageURL)
-        currentJobs[area] = job
-        
-        let cell = tableView.cellForRow(at: indexPath) as? PreplannedMapAreaTableViewCell
-        cell?.progressView.observedProgress = job.progress
-        
-        job.start(statusHandler: nil) { [weak self] (result, error) in
+        let cell = self.tableView.cellForRow(at: indexPath) as? PreplannedMapAreaTableViewCell
+        // Create download parameters with default values.
+        task.defaultDownloadPreplannedOfflineMapParameters(with: area) { [weak self] (params, error) in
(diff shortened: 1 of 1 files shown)


yo1995:
|Before | After|
|-|-|
|Before|After|

yo1995:
One more thing to mention is that I try to fix the UI where the checkmark `accessoryView` covers the `progressView`. However, after doing some researches online, I found the steps to fix that are not trivial, so I left it as it is right now.

Jietu20200409-112556

arcgis-ios-sdk-samples/Maps/Download preplanned map/MapSelectionTableViewController.swift
```diff
@@ -194,23 +194,30 @@ class MapSelectionTableViewController: UITableViewController {
...
         try? FileManager.default.removeItem(at: area.mapPackageURL)
-        
-        let job = task.downloadPreplannedOfflineMapJob(with: area, downloadDirectory: area.mapPackageURL)
-        currentJobs[area] = job
-        
-        let cell = tableView.cellForRow(at: indexPath) as? PreplannedMapAreaTableViewCell
-        cell?.progressView.observedProgress = job.progress
-        
-        job.start(statusHandler: nil) { [weak self] (result, error) in
+        let cell = self.tableView.cellForRow(at: indexPath) as? PreplannedMapAreaTableViewCell
```
philium, line 199: Please move this to inside the completion handler. Since cells are reused, this cell may or may not be the cell we want when `task.defaultDownloadPreplannedOfflineMapParameters` completes. That is why the method takes an index path, not a cell.

yo1995: Thanks for pointing that out, I fixed it.  
I don't fully understand the mechanism of `dequeueReusableCell` though. 🤔  Previously what I did is to refer to a cell first, and update it async afterwards, in this case in an escaping closure. If that cell that I referred to moves out of the view before the completion closure get executed, will it be reused and put into the queue, which cause the problem (not the same cell) you mentioned? 

Am I understanding it correctly or there is another reason that the cell might change? Thanks!

philium: That's correct. If a cell is scrolled offscreen, it is put back into the reuse queue and most likely dequeued for use by a different row. Retrieving the cell before performing an asynchronous operation and then updating it afterwards could lead to updating the cell for the wrong row, causing both the expected row and the accidentally modified row to be incorrect.

yo1995:
Will merge in this sprint according to the discussion with Lucas and Mike. Further fixes for all platforms will be done in other PRs.

PR #850, opened 2020-04-13
Trace utility network

yo1995:
Implementation of the `Trace utility network` sample. A brief demo below.

A link to current `v.next` branch folder: 
- https://github.com/Esri/arcgis-runtime-samples-ios/blob/v.next/arcgis-ios-sdk-samples/Utility%20Network/Connected%20Trace

- https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/TraceUtilityNetwork_vNext/arcgis-ios-sdk-samples/Utility%20Network/Trace%20utility%20network

|`Downstream`|`Upstream`|
|:------:|:------:|
|||
|`Connected`|`Subnetwork`|
|||


yo1995:
Track my questions here
---

- [ ] Why not differ the medium voltage network and low voltage network with different colors? It doesn't seem distinguishable with same color and line/dash style.
- [ ] Should I refactor the code according to the latest format? E.g. init of mapView, remove access control modifiers, etc.?

yo1995:
title image

Propose to have this image as the title image. Does it look OK?

yo1995:
Will need to read and understand how do other trace types work, so that I can check their behaviors.

yo1995:
Hi Phil and Nick, I invited both of you to review as you have better idea of this sample.  
Please feel free to leave comments and answer my questions listed above. 🙂  
Also, you can leave the README and json for now, as they are subject to change after my addressing the review.

philium:
> There is a lingering fileprivate global variable atop the source code

Is it actually being used anywhere else? If not, I would just remove it.

philium:
> Should I remove all the print functions in the source code?

I would say yes. If they are reporting an error, they should be changed to display an alert instead.

yo1995:
> > There is a lingering fileprivate global variable atop the source code
> 
> Is it actually being used anywhere else? If not, I would just remove it.

It is only used twice in the class. I checked the file version history and it is there since the first implementation. https://github.com/Esri/arcgis-runtime-samples-ios/blob/7002cfbe92f8c963ef0f0e545c8aa1e6f2310c63/arcgis-ios-sdk-samples/Utility%20Network/Connected%20Trace/ConnectedTraceViewController.swift#L18

I think it is left there probably because it allows people to change that attribute string easier? 

Anyway I will move it inside the class. Probably Nick can share some thoughts on why it's there.

nixta:
> Probably Nick can share some thoughts on why it's there.

It's needed to define the UniqeValueRenderer and to set an attribute value on barrier graphics to drive that UVA. There's probably a much better place and way to declare that. I probably noticed at the last minute that it was a hard-coded string and decided it needed to be declared once and was being lazy.

nixta:
Actually, the `barrierUniqueValue` should probably be declared with `values: [InteractionMode.addingBarriers.traceLocationType]`

https://github.com/Esri/arcgis-runtime-samples-ios/blob/7002cfbe92f8c963ef0f0e545c8aa1e6f2310c63/arcgis-ios-sdk-samples/Utility%20Network/Connected%20Trace/ConnectedTraceViewController.swift#L48-L51

but that smells like it deserves some sort of refactor because we're mixing an app UX enum with Graphics attributes.

yo1995:
https://github.com/Esri/arcgis-runtime-samples-ios/blob/357c5519b4117d91ef64b88d76f5cdd576a220d9/arcgis-ios-sdk-samples/Utility%20Network/Trace%20utility%20network/TraceUtilityNetworkViewController.swift#L177-L184

👆The question related to capture list that I asked earlier. I wonder why/why not to include an instance property in the list, as long as `self` is already captured.

nixta:
As I see it, there are two things going on.

1. `self` is being used in the `guard` to defend against the sample being switched out.

2. `currentMode` is captured to ensure that when the callback is entered, the value of `captureMode` at the time `selectTerminal()` was _called_ is used and not potentially an updated value if `selectTerminal()` takes a long time to call back, and the UI allows it. It also simplifies the call to `self.add()`.

It's questionable whether `currentValue` could ever change in the time it takes `selectTerminal()` to call back, but I think it's a reasonable and resilient pattern.

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -188,8 +214,13 @@ class ConnectedTraceViewController: UIViewController, AGSGeoViewTouchDelegate {
     
     // MARK: Perform Trace
     @IBAction func traceNetwork(_ sender: Any) {
-        SVProgressHUD.show(withStatus: "Running connected trace…")
-        let parameters = traceParameters
+        SVProgressHUD.show(withStatus: "Running \(traceType.0) trace…")
+        let parameters = AGSUtilityTraceParameters(traceType: traceType.1, startingLocations: traceParameters.startingLocations)
+        traceParameters.barriers.forEach { barrier in
+            parameters.barriers.append(barrier)
+        }
```
philium, line 221:
```suggestion
        parameters.barriers.append(contentsOf: traceParameters.barriers)
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -252,6 +281,12 @@ class ConnectedTraceViewController: UIViewController, AGSGeoViewTouchDelegate {
     }
     
     // MARK: Terminal Selection UI
+    /// Popup an action sheet to select one from multiple terminals, or return if there is only one.
```
philium, line 284:
```suggestion
    // MARK: Terminal Selection UI
    
    /// Popup an action sheet to select one from multiple terminals, or return if there is only one.
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -252,6 +281,12 @@ class ConnectedTraceViewController: UIViewController, AGSGeoViewTouchDelegate {
...
+    ///
+    /// - Parameters:
+    ///   - terminals: A list of terminals.
```
philium, line 287:
```suggestion
    ///   - terminals: An array of terminals.
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -252,6 +281,12 @@ class ConnectedTraceViewController: UIViewController, AGSGeoViewTouchDelegate {
...
+    ///   - mapPoint: The location tapped on the map.
+    ///   - completion: Completion clousre to pass the selected terminal.
```
philium, line 289:
```suggestion
    ///   - completion: Completion closure to pass the selected terminal.
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -316,12 +351,35 @@ class ConnectedTraceViewController: UIViewController, AGSGeoViewTouchDelegate {
...
+            ("Upstream", AGSUtilityTraceType.upstream),
+            ("Downstream", AGSUtilityTraceType.downstream)
+        ]
+        types.forEach { type in
+            let action = UIAlertAction(title: type.0, style: .default) { [unowned self] _ in
+                self.traceType = type
+                self.setStatus(message: "Trace type \(type.0) selected.")
+            }
+            alertController.addAction(action)
+        }
```
philium, line 369: Referencing to tuple members without labels really harms readability. It would be helpful to add labels to the tuple members:

```suggestion
        let types: [(name: String, type: AGSUtilityTraceType)] = [
            ("Connected", .connected),
            ("Subnetwork", .subnetwork),
            ("Upstream", .upstream),
            ("Downstream", .downstream)
        ]
        types.forEach { (name, type) in
            let action = UIAlertAction(title: name, style: .default) { [unowned self] _ in
                self.traceType = type
                self.setStatus(message: "Trace type \(name) selected.")
            }
            alertController.addAction(action)
        }
```

This suggestion requires a couple other changes. The type of `traceType` would need to be `AGSUtilityTraceType`, which I don't think will cause any problems.

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
yo1995: Cool! 🤩

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
yo1995: I keep the name string of the trace type so that it can be displayed in the progress HUD with correct type name. Previously it was hardcoded as "Connected" here: https://github.com/Esri/arcgis-runtime-samples-ios/blob/c86a0f2155c7e84d716fb9c5c24fdcb8629fced2/arcgis-ios-sdk-samples/Utility%20Network/Connected%20Trace/ConnectedTraceViewController.swift#L191

arcgis-ios-sdk-samples/Utility Network/Trace utility network/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## How it works
+
+1. Create an `AGSMapView` and listen for `didTap` events on the `AGSGeoViewTouchDelegate`.
+2. Create an `AGSMap` that contains `AGSFeatureLayer`(s) that are part of a utility network.
+3. Create and load an `AGSUtilityNetwork` with the same feature service URL and map.
+4. Add an `AGSGraphicsOverlay` with symbology that distinguishes starting locations from barriers.
+5. Identify tapped features on the map and add an `AGSGraphic` that represents its purpose (starting point or barrier) at the tapped location.
+6. Create an `AGSUtilityElement` for the identified feature.
+7. Determine the type of the identified feature using `AGSUtilityNetworkSource.sourceType` passing its table name.
```
philium, line 23: "passing its table name" doesn't seem to make sense since `AGSUtilityNetworkSource.sourceType` is a property.

```suggestion
7. Determine the type of the identified feature using `AGSUtilityNetworkSource.sourceType`.
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/README.md
```diff
@@ -0,0 +1,57 @@
...
+8. If the type is a `junction`, display a terminal picker when more than one terminal is found and create an `AGSUtilityElement` using the selected terminal, or the single terminal if there is only one.
```
philium, line 24: Since you're using a code symbol here for `junction`, I would remove the _a_. If it were a Boolean value, you wouldn't say "if the type is a `true` or a `false`".

```suggestion
8. If the type is `junction`, display a terminal picker when more than one terminal is found and create an `AGSUtilityElement` using the selected terminal, or the single terminal if there is only one.
```

Alternatively, you could not use the code symbol:

```suggestion
8. If the network source is a junction, display a terminal picker when more than one terminal is found and create an `AGSUtilityElement` using the selected terminal, or the single terminal if there is only one.
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/README.md
```diff
@@ -0,0 +1,57 @@
...
+9. If the type is an `edge`, create an `AGSUtilityElement` from the identified feature and compute how far along the edge the user tapped using `class AGSGeometryEngine.fraction(alongLine:to:tolerance:)`.
+10. Add this `AGSUtilityElement` to a collection of starting locations or barriers.
+11. Create `AGSUtilityTraceParameters` with the selected trace type along with the collected starting locations and barriers (if applicable).
+12. Set the `AGSUtilityTraceConfiguration` with the utility tier's `traceConfiguration` property.
+13. Run `AGSUtilityNetwork.trace(with:completion:)` with the specified starting points and (optionally) barriers.
+14. Group the `AGSUtilityElementTraceResult.elements` by their `networkSource.name`.
+15. For every `AGSFeatureLayer` in this map with trace result elements, select features by converting `AGSUtilityElement`(s) to `AGSArcGISFeature`(s) using `AGSUtilityNetwork.features(for:completion:)`
```
philium, line 31: Missing a period:

```suggestion
15. For every `AGSFeatureLayer` in this map with trace result elements, select features by converting `AGSUtilityElement`(s) to `AGSArcGISFeature`(s) using `AGSUtilityNetwork.features(for:completion:)`.
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+1. Create an `AGSMapView` and listen for `didTap` events on the `AGSGeoViewTouchDelegate`.
+2. Create an `AGSMap` that contains `AGSFeatureLayer`(s) that are part of a utility network.
+3. Create and load an `AGSUtilityNetwork` with the same feature service URL and map.
+4. Add an `AGSGraphicsOverlay` with symbology that distinguishes starting locations from barriers.
+5. Identify tapped features on the map and add an `AGSGraphic` that represents its purpose (starting point or barrier) at the tapped location.
+6. Create an `AGSUtilityElement` for the identified feature.
+7. Determine the type of the identified feature using `AGSUtilityNetworkSource.sourceType` passing its table name.
+8. If the type is a `junction`, display a terminal picker when more than one terminal is found and create an `AGSUtilityElement` using the selected terminal, or the single terminal if there is only one.
+9. If the type is an `edge`, create an `AGSUtilityElement` from the identified feature and compute how far along the edge the user tapped using `class AGSGeometryEngine.fraction(alongLine:to:tolerance:)`.
```
philium, line 25:
```suggestion
9. If the type is `edge`, create an `AGSUtilityElement` from the identified feature and compute how far along the edge the user tapped using `class AGSGeometryEngine.fraction(alongLine:to:tolerance:)`.
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -15,45 +15,60 @@
...
+                                                        values: [5])
+                let lowVoltageValue = AGSUniqueValue(description: "N/A",
+                                                     label: "Low voltage",
+                                                     symbol: AGSSimpleLineSymbol(style: .dash, color: darkCyan, width: 3),
+                                                     values: [3])
+                layer.renderer = AGSUniqueValueRenderer(fieldNames: ["ASSETGROUP"],
+                                                        uniqueValues: [mediumVoltageValue, lowVoltageValue],
+                                                        defaultLabel: "",
+                                                        defaultSymbol: AGSSimpleLineSymbol()
+                )
```
philium, line 50: I know having things pushed so far to the right is rather common in Objective-C, but in Swift, we usually do things a little differently:

```suggestion
                let mediumVoltageValue = AGSUniqueValue(
                    description: "N/A",
                    label: "Medium voltage",
                    symbol: AGSSimpleLineSymbol(style: .solid, color: darkCyan, width: 3),
                    values: [5]
                )
                let lowVoltageValue = AGSUniqueValue(
                    description: "N/A",
                    label: "Low voltage",
                    symbol: AGSSimpleLineSymbol(style: .dash, color: darkCyan, width: 3),
                    values: [3]
                )
                layer.renderer = AGSUniqueValueRenderer(
                    fieldNames: ["ASSETGROUP"],
                    uniqueValues: [mediumVoltageValue, lowVoltageValue],
                    defaultLabel: "",
                    defaultSymbol: AGSSimpleLineSymbol()
                )
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -117,20 +138,25 @@ class ConnectedTraceViewController: UIViewController, AGSGeoViewTouchDelegate {
...
             self.addStartElementOrBarrier(for: feature, at: mapPoint)
         }
     }
-
+    
+    /// Based on the selection mode, the tapped utility element is added either to the starting locations or barriers for the trace parameters.
+    /// An appropriate graphic is created at the tapped location to mark the element as either a starting location or barrier.
+    ///
+    /// - Parameters:
+    ///   - feature: The geoelement retrieved as an AGSArcGISFeature.
```
philium, line 158:
```suggestion
    ///   - feature: The geoelement retrieved as an `AGSArcGISFeature`.
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -117,20 +138,25 @@ class ConnectedTraceViewController: UIViewController, AGSGeoViewTouchDelegate {
...
+    ///   - location: The AGSPoint used to identify utility elements in the utility network.
```
philium, line 159:
```suggestion
    ///   - location: The `AGSPoint` used to identify utility elements in the utility network.
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -188,8 +214,12 @@ class ConnectedTraceViewController: UIViewController, AGSGeoViewTouchDelegate {
     
     // MARK: Perform Trace
     @IBAction func traceNetwork(_ sender: Any) {
-        SVProgressHUD.show(withStatus: "Running connected trace…")
-        let parameters = traceParameters
+        SVProgressHUD.show(withStatus: "Running \(traceType.name) trace…")
```
philium, line 217: Should the name be lowercased to match the previous behavior?

```suggestion
        SVProgressHUD.show(withStatus: "Running \(traceType.name.lowercased()) trace…")
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -252,6 +280,13 @@ class ConnectedTraceViewController: UIViewController, AGSGeoViewTouchDelegate {
     }
     
     // MARK: Terminal Selection UI
+    
+    /// Popup an action sheet to select one from multiple terminals, or return if there is only one.
```
philium, line 284:
```suggestion
    /// Presents an action sheet to select one from multiple terminals, or return if there is only one.
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -316,12 +351,35 @@ class ConnectedTraceViewController: UIViewController, AGSGeoViewTouchDelegate {
...
+            let action = UIAlertAction(title: name, style: .default) { [unowned self] _ in
+                self.traceType = (name, type)
+                self.setStatus(message: "Trace type \(name) selected.")
+            }
+            alertController.addAction(action)
+        }
+        let cancelAction = UIAlertAction(title: "Cancel", style: .cancel, handler: nil)
+        alertController.addAction(cancelAction)
+        alertController.popoverPresentationController?.barButtonItem = typeButton
+        present(alertController, animated: true, completion: nil)
```
philium, line 373:
```suggestion
        present(alertController, animated: true)
```

arcgis-ios-sdk-samples/Utility Network/Trace utility network/README.md
yo1995, line 31:
```suggestion
15. For every `AGSFeatureLayer` in this map with trace result elements, select features by converting `AGSUtilityElement`(s) to `AGSArcGISFeature`(s) using `AGSUtilityNetwork.features(for:completion:)`.
```
Missing trailing dot here.

arcgis-ios-sdk-samples/Utility Network/Trace utility network/TraceUtilityNetworkViewController.swift
yo1995: Right! I see it pretty awkwardly aligned, but forget where I should cut off the first line 😅 and auto re-indent tool does not fix it.

yo1995:
Fixed a few more issues and re-indented some exceedingly long lines.

yo1995:
Hi Sarat, @saratchandrakarumuri please take a look at this sample! Thanks 😃

yo1995:
> Sample looks good but I have only one following question
> 
> * Even though we are modifying an existing sample, with the total restructure and rename, this looks like a new sample. Shouldn't we change the copyright header on swift file to `2020` @philium ?

Hi Sarat, please refer to this https://github.com/Esri/arcgis-runtime-samples-qt/pull/964#discussion_r373265500, and also common-samples 1672. According to their discussion, it is suggested to keep the original year info in the legal/license part.

PR #857, opened 2020-04-23
Animate images with image overlay

yo1995:
Implementation of the `Animate images with image overlay` sample. A brief demo below.

- https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/AnimateImage_vNext/arcgis-ios-sdk-samples/Scenes/Animate%20image%20overlay

|`Ever changing world`|
|:------:|
||


yo1995:
UI design: How to make sth like a drop down menu with value displayed. Change title of a button seems awkward to me.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+    // MARK: Instance properties
+    
+    /// A circular loop counter within a length.
+    private var loopCounter: LoopCounter!
+    /// The graphics overlay to show a graphic at the tapped point.
+    let imageOverlay = AGSImageOverlay()
+    /// The frame rate in the unit of frame per second.
+    var frameRate = 60.0
+    /// An array to hold the URLs for the overlay images.
+    var imageURLs: [URL] = []
```
yo1995, line 54: During the meeting on May 1st, Trevor mentioned it might be better to preload all the images/image frames into an array in the memory for Kotlin, so that the disk/file loading performance won't lag the animation.

I tested with loading all 240 images into the memory as `UIImage`, and a rough profiling gives me ~40 more MB memory used, which is acceptable. However, loading from disk on-the-fly does not downgrade the overall performance. So I'll leave it here for further discussion.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+    }
+    
+    /// Set the image at current index to the image overlay.
+    func setImageFrame() {
+        let imageFrame = AGSImageFrame(url: imageURLs[loopCounter.currentIndex], extent: pacificSouthwestEnvelope)
+        imageOverlay.imageFrame = imageFrame
+        loopCounter.next()
+    }
+    
+    func getOpacityString() -> String {
```
yo1995, line 122: This function might be of less value, but I'd like to keep format control characters within a single place.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+    @IBAction func playPauseButtonTapped(_ button: UIBarButtonItem) {
+        if isAnimating {
+            frameRateTimer?.invalidate()
+        } else {
+            startAnimation(frameRate)
+        }
+        isAnimating.toggle()
+    }
+    
+    @IBAction func frameRateButtonTapped(_ button: UIBarButtonItem) {
```
yo1995, line 142: Should I only create once and keep a reference to the alertVC, as it won't change for this sample?

yo1995: No as the memory footprint is pretty small.

yo1995:
In this morning review, Jared mentioned that to be more strict with wording, the `FPS` might not be accurate enough. Either use `16.66 millisecond to display each image frame`, or use more vague descriptor as slow medium and fast.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+import UIKit
+import ArcGIS
+
+class AnimateImagesWithImageOverlayViewController: UIViewController {
+    // MARK: Storyboard views
+    
+    /// The button to toggle image overlay animation.
+    @IBOutlet weak var playPauseButton: UIBarButtonItem!
+    /// The button to choose a frame rate for the animation
+    @IBOutlet weak var frameRateButton: UIBarButtonItem!
```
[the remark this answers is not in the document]
philium: Please set the possible titles of this bar item as well. That may fix a bug I found where the button item couldn't be tapped on an iPhone after rotating to landscape

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+class AnimateImagesWithImageOverlayViewController: UIViewController {
+    // MARK: Storyboard views
+    
+    /// The button to toggle image overlay animation.
+    @IBOutlet weak var playPauseButton: UIBarButtonItem!
```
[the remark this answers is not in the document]
yo1995: Sounds good. I'll experiment on using systemBBI and expect your feedback in next review.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+    }
+    
+    deinit {
+        // Invalidates the timer before exiting the sample.
+        frameRateTimer?.invalidate()
+    }
+}
+
+/// A circular loop counter within a length.
+private struct LoopCounter {
```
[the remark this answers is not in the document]
yo1995: Sounds great! I'll try with this idea. 😍

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+        }
+        self.frameRateTimer = timer
+        // Add the timer to common mode run loop, so the timer is not effected by UI events.
+        RunLoop.main.add(timer, forMode: .common)
+    }
+    
+    /// Load the URLs for the overlay images.
+    ///
+    /// - Parameter completion: If it loads successfully, enable the UI components in the completion closure.
+    func loadImageURLs(completion: () -> Void) {
```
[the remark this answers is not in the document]
yo1995: Initially I write this non-escaping completion closure for updating UI after loading complete, so that I don't incorporate unrelated logic into loading data function. I'll try with another way to put it.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
yo1995: I've done further profiling and decided to keep an array of `UIImage`s instead.

Keeping an array of `AGSImageFrame` consume more than 2GB memory, whereas `[UIImage]` uses less than 100MB.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,235 @@
...
+    /// The toolbar in the view controller.
+    @IBOutlet weak var toolbar: UIToolbar!
+    /// The label to display opacity level.
+    @IBOutlet weak var opacityLabel: UILabel!
+    
+    /// The slider to change opacity level, from transparent 0% to opaque 100%.
+    @IBOutlet weak var opacitySlider: UISlider! {
+        didSet {
+            imageOverlay.opacity = opacitySlider.value
+            opacityLabel.text = getOpacityPercentageString(percentage: opacitySlider.value)
```
yo1995, line 36: There might be better way to initialize the slider's value and text.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,235 @@
...
+    var isAnimating = false {
+        didSet {
+            if isAnimating {
+                let index = toolbar.items!.firstIndex(of: playButtonItem)!
+                toolbar.items![index] = pauseButtonItem
+            } else {
+                let index = toolbar.items!.firstIndex(of: pauseButtonItem)!
+                toolbar.items![index] = playButtonItem
+            }
+        }
```
yo1995, line 91: This is a bit cumbersome, but I'm trying to avoid modifying the bar button at a certain index.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,235 @@
...
+    @IBAction func speedButtonTapped(_ button: UIBarButtonItem) {
+        let alertController = UIAlertController(
+            title: "Choose playback speed.",
+            message: nil,
+            preferredStyle: .actionSheet
+        )
+        let speedChoices: [(name: String, fps: Int)] = [
+            ("Fast", 60),
+            ("Medium", 30),
+            ("Slow", 15)
```
yo1995, line 174: According to the new design, the frame rate is not shown explicitly.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,235 @@
...
+    
+    /// Set current image to the image overlay.
+    @objc
+    func setImageFrame() {
+        let frame = AGSImageFrame(image: imagesIterator.next()!, extent: pacificSouthwestEnvelope)
+        imageOverlay.imageFrame = frame
+    }
+    
+    func getOpacityPercentageString(percentage: Float) -> String {
+        return "Opacity\n" + percentageFormatter.string(from: NSNumber(value: percentage))!
```
philium, line 146: FYI: You don't have to change this, but I wanted to point something out. When part of the string is static, particularly when it is on a separate line, it is more efficient to put the static portion in a separate label. "Opacity" could be in one label and the percent in another. That way, only the part that changes, the percent, is redrawn each time. "Opacity" would only be drawn once. Does that make sense?

philium, line 146: I don't know if one is more efficient than the other, but you can also cast to `NSNumber` (`percentage as NSNumber`) instead of going through the initializer. I suspect casting might be more efficient, but I don't really know for sure.

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,235 @@
...
+        let scene = AGSScene(basemap: AGSBasemap(baseLayer: worldDarkGrayBasemap))
+        scene.baseSurface = surface
+        return scene
+    }
+    
+    func startAnimation(fps: Int) {
+        // Invalidate display link to stop previous ongoing animation.
+        displaylink?.invalidate()
+        // Create a new display link.
+        displaylink = CADisplayLink(target: self, selector: #selector(setImageFrame))
```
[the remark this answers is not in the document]
yo1995: I'll try with that!

arcgis-ios-sdk-samples/Scenes/Animate image overlay/AnimateImagesWithImageOverlayViewController.swift
yo1995: Yes I believe that is the right way. 

I'm afraid of adding redundant or wrong constraints when setting up multiple custom `UILabel`s within a `UIBarButtonItem`. Nevertheless I'll give it a try 🤔

yo1995: This is interesting. 

My knowledge for type casting still lives with C/C++. I couldn't find the details about how Swift types cast to ObjC types (most posts suggest it is Apple's compiling "magic") after digging a bit. Nevertheless it makes sense that compile time type casting `as` should be faster than runtime `init`. So I'll adopt this change.

arcgis-ios-sdk-samples/Scenes/Animate images with image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,215 @@
...
+    var displaylink: CADisplayLink!
+    
+    /// An iterator to hold and loop through the overlay images.
+    private lazy var imagesIterator: CircularIterator<UIImage> = {
+        // Get the URLs to images added to the project's folder reference.
+        var imageURLs = Bundle.main.urls(forResourcesWithExtension: "png", subdirectory: "PacificSouthWest") ?? []
+        imageURLs.sort { $0.lastPathComponent < $1.lastPathComponent }
+        return CircularIterator(
+            elements: imageURLs.map { UIImage(contentsOfFile: $0.path)! }
+        )
```
philium, line 66: This isn't a big deal right here, but I've found it beneficial to use immutable values whenever possible. `imageURLs` doesn't need to be mutable.

```suggestion
        let imageURLs = Bundle.main.urls(forResourcesWithExtension: "png", subdirectory: "PacificSouthWest") ?? []
        let images = imageURLs
            .sorted { $0.lastPathComponent < $1.lastPathComponent }
            .map { UIImage(contentsOfFile: $0.path)! }
        return CircularIterator(elements: images)
```

arcgis-ios-sdk-samples/Scenes/Animate images with image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,215 @@
...
+    }
+    
+    @IBAction func playPauseButtonTapped(_ button: UIBarButtonItem) {
+        let index = toolbar.items!.firstIndex(of: button)!
+        if !displaylink.isPaused {
+            toolbar.items![index] = playButtonItem
+        } else {
+            toolbar.items![index] = pauseButtonItem
+            // Set framerate of the display link before starting to play.
+            displaylink.preferredFramesPerSecond = frameRate
```
philium, line 129: Does the display link require that this property be set before being paused? If not, setting it each time the animation is resumed it overkill. It really should only be set once when the user changes the speed.

arcgis-ios-sdk-samples/Scenes/Animate images with image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,215 @@
...
+            let camera = AGSCamera(location: point, heading: 353.994, pitch: 48.5495, roll: 0)
+            sceneView.setViewpointCamera(camera)
+            sceneView.imageOverlays.add(imageOverlay)
+        }
+    }
+    
+    // MARK: Instance properties
+    
+    /// The frame rate in the unit of frame per second.
+    var frameRate = 60
```
philium, line 53: I'm wondering if you could get rid of this property so that the display link is the source of truth for what the current frame rate is. Here's how that could be done:

```swift
let displaylink = CADisplayLink(target: self, selector: #selector(setImageFrame))
displaylink.preferredFramesPerSecond = self.displaylink?.preferredFramesPerSecond ?? 60
displaylink.isPaused = true
// Add to main thread common mode run loop, so it is not effected by UI events.
displaylink.add(to: .main, forMode: .common)
self.displaylink = displaylink
```

I haven't tried it, but it seems like that should work.

arcgis-ios-sdk-samples/Scenes/Animate images with image overlay/AnimateImagesWithImageOverlayViewController.swift
```diff
@@ -0,0 +1,215 @@
...
+    /// The image overlay to show image frames.
+    let imageOverlay = AGSImageOverlay()
+    /// A timer to synchronize image frame animation to the refresh rate of the display.
+    var displaylink: CADisplayLink!
```
philium, line 57: Please rename this to `displayLink`.

yo1995:
@saratchandrakarumuri Please take a look at this sample! Thanks. 🤗

yo1995:
Oops it seems I break the project file structure 😓 . I'll revert before respond!

yo1995:
> Animation is automatically getting paused on going away from sceneView to info page... Is that an expected behavior?

Yes this is expected, according to https://github.com/Esri/arcgis-runtime-samples-ios/pull/857#discussion_r431241702 . The animation is paused when it is invisible, so that clock resources won't be consumed for the animation timer. 🙂  Also the project file conflict is fixed!

PR #859, opened 2020-05-01
Various project changes

philium:
These are changes Xcode wants to make whenever the project file is modified. And whenever the version number is modified, Xcode moves it out of the info plist and into the project.


philium:
> I was thinking if there is a way to prevent Xcode from making those project file changes

I think they happen as a result of resolving merge conflicts in the project file or plist. Whenever that happens, it is a good idea to open the project and do something that forces Xcode to resave the project or plist file.

PR #865, opened 2020-05-07
Update download script for archive with multiple files

yo1995:
This PR updates the `download-portal-item-data.swift` script which currently handles the offline data on-demand download for the app. It
- Improves support for ZIP archived portal items.
- Added logics to detect if a file already exists.

A few concerns below.

---

- [x] 1. The script still relies on the filename in `PortalItems.plist`, which run against what being proposed in update #729.

- [x] 2. ~~The script **WILL** create an enclosing folder for the archive to extract to, so that it can ensure all the files from an archive stay in the same folder, rather than scattering under sub-root. See below.~~

**Edit**: With "-j" argument, no enclosing folder will be created.

`enclosing`
```
.
|---enclosing_folder
      |---1.png
      |---2.png
      | ...
```
`flat`
```
.
|---1.png
|---2.png
| ...
```

One downside is that if the files in archive are already wrapped by an enclosing folder, this method will create an effect such as below.
```
.
|---created_enclosing_folder
      |---original_enclosing_folder
            |---1.png
            |---2.png
            | ...
```
Unless the script can handle the case that subfolder is only created for a "flat" archive.

- [x] 3. ~~Some unused functions stays in the file in case they are needed in the future.~~

**Edit**: Removed unused functions.

- [x] 4. ~~Currently the extracted files are not recorded in `.downloaded_items` dictionary. Instead, the zip archive itself is kept track of.~~

**Edit**: There are two cases. First, if the downloaded item is a single file or an archive containing a single file, then the filename of that single file will be recorded. Second, if the downloaded item is an archive, then the `{archive_name}.zip` will be recorded.

Before download, the script will check if the filename-identifier mapping exists, and get the existing file's name. The file will only be downloaded when it doesn't exist on disk.

- [x] ~~5. In this script, both the .zip archive and the extracted files are kept in the storage, which wastes some disk space. Because the sample relies on the `on-demand resource tag`, and the tag is given to the zip archive rather than the extracted folder.~~

  This doesn't apply anymore.

- [x] 6. The extracted files will need to be add to the project file's folder reference, so that it can be retrieved by the sample.

### DIFF
--- a/Scripts/download-portal-item-data.swift
+++ b/Scripts/download-portal-item-data.swift
@@ -138,7 +138,7 @@ func downloadFile(at sourceURL: URL, destinationURLProvider: URLProvider, comple
                 if isArchive {
                     let fileCount = try countFilesInArchive(at: temporaryURL)
                     print("File count in the archive is \(fileCount)")
-                    extractURL = downloadURL.deletingLastPathComponent()
+                    extractURL = destinationURLProvider.makeSubFolderForArchive(folderName: (suggestedFilename as NSString).deletingPathExtension)
                 }
                 
                 try FileManager.default.createDirectory(at: downloadURL.deletingLastPathComponent(), withIntermediateDirectories: true)
@@ -149,9 +149,10 @@ func downloadFile(at sourceURL: URL, destinationURLProvider: URLProvider, comple
                 
                 if isArchive {
                     try uncompressArchive(at: temporaryURL, to: extractURL!)
+                } else {
+                    try FileManager.default.moveItem(at: temporaryURL, to: downloadURL)
                 }
-                try FileManager.default.moveItem(at: temporaryURL, to: downloadURL)
-
+                
                 completion(.success(downloadURL))
             } catch {
                 completion(.failure(error))
@@ -238,8 +239,9 @@ portalItems.forEach { (portalURLString, portalItems) in
         
         // Check if it is a non-archive single file.
         let isFileExist: Bool = FileManager.default.fileExists(atPath: destinationURLProvider.makeURL(filename: filename).path)
-
-        // Check if there is a sub-folder for the corresponding archive, and the sub-folder is empty or not
+        
+        // Check if there is a sub-folder for the corresponding archive, and the sub-folder is empty or not.
+        // The corresponding sub-folder has the same name as the archive, but without the extension.
         let subFolderURL = destinationURLProvider.makeSubFolderForArchive(folderName: (filename as NSString).deletingPathExtension)
         let paths = try? FileManager.default.contentsOfDirectory(
             at: subFolderURL,
@@ -251,7 +253,7 @@ portalItems.forEach { (portalURLString, portalItems) in
         /// nil: the folder does not exist or not accessible
         let isEmptySubFolder = paths?.isEmpty
         
-        if isFileExist && isEmptySubFolder != nil && !isEmptySubFolder! {
+        if isEmptySubFolder != nil && !isEmptySubFolder! {
             // One exception to this case is that the content in the folder is changed. There is no way to detect that
             // unless we keep track of each extracted file in the .downloaded_items dictionary.
             print("Item \(portalItem.identifier) has already been downloaded, and is extracted to an folder")
--- a/Scripts/download-portal-item-data.swift
+++ b/Scripts/download-portal-item-data.swift
@@ -27,6 +27,7 @@ import Foundation
 
 protocol URLProvider {
     func makeURL(filename: String) -> URL
+    func makeSubFolderForArchive(folderName: String) -> URL
 }
 
 struct DestinationURLProvider: URLProvider {
@@ -39,6 +40,19 @@ struct DestinationURLProvider: URLProvider {
             url.appendPathComponent(subdirectory, isDirectory: true)
         }
         url.appendPathComponent(filename, isDirectory: false)
+        return url
+    }
+    
+    /// Make a sub-folder path for the extracted files from an archive.
+    ///
+    /// - Parameter folderName: The name of the folder.
+    /// - Returns: A URL of the folder.
+    func makeSubFolderForArchive(folderName: String) -> URL {
+        var url = downloadDirectory
+        if let subdirectory = fileTypes.first(where: { $0.value.contains("zip") })?.key {
+            url.appendPathComponent(subdirectory, isDirectory: true)
+        }
+        url.appendPathComponent(folderName, isDirectory: true)
         return url
     }
 }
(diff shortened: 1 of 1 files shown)


yo1995:
This PR also blocks #857 . After the download script is updated, `Animate images with image overlay` can be reviewed and tested by then.

philium:
> Because the sample relies on the on-demand resource tag, and the tag is given to the zip archive rather than the extracted folder.

@yo1995 I don't think that does what you think it does. The on-demand resources should be the images themselves, not a zip file. I don't see any reason to keep the zip file around, just like we don't keep it for archives containing a single file.

yo1995:
OK I see. Let me try it in another way.

yo1995:
I've updated the script a bit, and it seems to work fine with both #857 and #866 . Please take another look. 🙏

Scripts/download-portal-item-data.swift
```diff
@@ -76,47 +90,70 @@ func nameOfFileInArchive(at url: URL) throws -> String {
...
 func uncompressArchive(at sourceURL: URL, to destinationURL: URL) throws {
-    FileManager.default.createFile(atPath: destinationURL.path, contents: nil)
-    let fileHandle = try FileHandle(forWritingTo: destinationURL)
+    let outputPipe = Pipe()
     let process = Process()
     process.executableURL = URL(fileURLWithPath: "/usr/bin/unzip", isDirectory: false)
-    process.arguments = ["-p", sourceURL.path]
-    process.standardOutput = fileHandle
+    // Unzip the archive into the specified sub-folder.
+    process.arguments = [sourceURL.path, "-d", destinationURL.path]
```
philium, line 119: It is unfortunate that the files in `PacificSouthWest` are in a directory. It would be nice to get that changed, but for now, maybe we could pass the `-j` option? That may need to be changed in the future, but it looks like it would work for now.

Scripts/download-portal-item-data.swift
```diff
@@ -200,16 +237,38 @@ portalItems.forEach { (portalURLString, portalItems) in
...
+        /// true: the folder exists and is empty;
+        /// false: the folder exists and has content;
+        /// nil: the folder does not exist or not accessible
+        let isEmptySubFolder = paths?.isEmpty
+        
+        if isEmptySubFolder != nil && !isEmptySubFolder! {
+            // One exception to this case is that the content in the folder is changed. There is no way to detect that
+            // unless we keep track of each extracted file in the .downloaded_items dictionary.
+            print("Item \(portalItem.identifier) has already been downloaded, and is extracted to an folder")
+            downloadedItems[portalItem.identifier] = filename
```
philium, line 261: This seems like overkill. The archive name should be unique, so I don't think we need to check that the directory is empty. If it exists, we can assume it has been downloaded.

Scripts/download-portal-item-data.swift
yo1995: Yes, `-j` would work for now, but we might need to set up a standard on how archives should be structured, to avoid the folder within subfolder situation.

yo1995:
Thanks for the review. Will make the changes on next Monday.

philium:
@yo1995 Thanks for working to get this script updated to support archives with multiple files!

Scripts/download-portal-item-data.swift
```diff
@@ -200,16 +237,38 @@ portalItems.forEach { (portalURLString, portalItems) in
...
+        } else if isFileExist && isEmptySubFolder == nil {
             print("Item \(portalItem.identifier) has already been downloaded")
-            // This is a temporary measure for users who currently don't have a
-            // downloaded items file.
+            // This is a temporary measure for users who currently don't have a downloaded items file.
```
philium, line 264: I forgot to mention that the line break here was intentional to keep the comment to a readable width, using the default page guide at column 80:

image


  |  
--- | ---
Before | image
After | image

I try to wrap all of my comments to the page guide, but especially documentation comments.

yo1995: Oh thanks for pointing that out! I once asked someone the line wrapping width but it wasn't clear so I kept around 120... I'll change it to 80!

Scripts/download-portal-item-data.swift
```diff
@@ -76,47 +90,70 @@ func nameOfFileInArchive(at url: URL) throws -> String {
...
-    FileManager.default.createFile(atPath: destinationURL.path, contents: nil)
-    let fileHandle = try FileHandle(forWritingTo: destinationURL)
+    let outputPipe = Pipe()
     let process = Process()
     process.executableURL = URL(fileURLWithPath: "/usr/bin/unzip", isDirectory: false)
-    process.arguments = ["-p", sourceURL.path]
-    process.standardOutput = fileHandle
+    // Unzip the archive into the specified sub-folder.
+    process.arguments = [sourceURL.path, "-d", destinationURL.path]
+    process.standardOutput = outputPipe
```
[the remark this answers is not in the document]
yo1995: Yep I just left here in case the output need to be redirected to another pipe or stdout. 😸 I'll silence the output then.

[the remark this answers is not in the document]
yo1995: Or to totally suppress output printing, "-qq" can be used. ref: https://linux.die.net/man/1/unzip

Scripts/download-portal-item-data.swift
```diff
@@ -76,47 +90,70 @@ func nameOfFileInArchive(at url: URL) throws -> String {
...
-                if !isArchive {
-                    downloadFilename = suggestedFilename
-                } else {
-                    downloadFilename = try nameOfFileInArchive(at: temporaryURL)
+                downloadURL = destinationURLProvider.makeURL(filename: suggestedFilename)
+                if isArchive {
+                    let fileCount = try countFilesInArchive(at: temporaryURL)
+                    print("File count in the archive is \(fileCount)")
+                    // Extract to a sub-folder with the same name as the archive without the extension.
+                    extractURL = destinationURLProvider.makeSubFolderURLForArchive(folderName: (suggestedFilename as NSString).deletingPathExtension)
```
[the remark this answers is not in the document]
yo1995: This ternary operator is awesome 🤩 .

yo1995:
Oops it seems the downloaded item is not correct. I'll take another look before requesting for review. 🤔

Scripts/download-portal-item-data.swift
```diff
@@ -200,10 +260,19 @@ portalItems.forEach { (portalURLString, portalItems) in
...
-        if FileManager.default.fileExists(atPath: destinationURLProvider.makeURL(filename: filename).path) {
+        let tempURL = destinationURLProvider.makeURL(filename: filename)
+        let pathURL = tempURL.pathExtension == "zip" ? tempURL.deletingPathExtension() : tempURL
+        
+        // Check if a single file or a folder exists.
+        var isDirectory = ObjCBool(false)
+        let isFileExist = FileManager.default.fileExists(atPath: pathURL.path, isDirectory: &isDirectory)
+        
+        if isFileExist && isDirectory.boolValue {
+            print("Item \(portalItem.identifier) has already been downloaded, and is extracted to an folder")
```
philium, line 271: It is easy enough to tell whether the archive item was extracted to a directory by looking inside of `Portal Data`. Could this just be the same message ("Item \(portalItem.identifier) has already been downloaded")? Then the `isDirectory` check wouldn't be needed.

Scripts/download-portal-item-data.swift
```diff
@@ -200,10 +260,19 @@ portalItems.forEach { (portalURLString, portalItems) in
     portalItems.forEach { (portalItem) in
         // Have we already downloaded the item?
         let filename = downloadedItems[portalItem.identifier] ?? portalItem.filename
-        if FileManager.default.fileExists(atPath: destinationURLProvider.makeURL(filename: filename).path) {
+        let tempURL = destinationURLProvider.makeURL(filename: filename)
+        let pathURL = tempURL.pathExtension == "zip" ? tempURL.deletingPathExtension() : tempURL
```
philium, line 264: Since paths and URLs are two different things, I would suggest renaming `pathURL` to `fileURL`.

Scripts/download-portal-item-data.swift
yo1995: Yes, this make sense.

yo1995:
> Looks good! Thank you so much.

Thanks for all the suggestions! 😸  
 It seems to require 2 approvals to merge now. Should I ask Nick or Ryan to also review this (they were mentioned in #729 )?

yo1995:
Will still need another approval from a reviewer with write access to merge 😅 . Thanks for the review!

philium:
@yo1995 I've given @pgruenler write access. You should be good to merge now!

yo1995:
> @yo1995 I've given @pgruenler write access. You should be good to merge now!

Thank you! 🎉

PR #866, opened 2020-05-08
Identify raster cell

yo1995:
Implementation of the `Identify raster cell` sample. A brief demo below.

- https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/IdentifyRasterCell-vNext/arcgis-ios-sdk-samples/Layers/Identify%20raster%20cell

|`Tap`|`Long press drag`|
|:------:|:------:|
|||


yo1995:
Should the source code for the stack view also be included in the info pane?

yo1995:
There seems not a way to disable or change the callout leader, and this makes it difficult to keep up with the latest design.

arcgis-ios-sdk-samples/Layers/Identify raster cell/IdentifyRasterCellViewController.swift
```diff
@@ -0,0 +1,118 @@
...
+                return
+            }
+            var calloutText: String = ""
+            var xyCoordinates: String = ""
+            identifyResult.geoElements.filter { $0 is AGSRasterCell }.forEach { cell in
+                cell.attributes.forEach { attribute in
+                    calloutText.append("\(attribute.key): \(attribute.value)\n")
+                }
+                xyCoordinates.append(String(format: "X: %.3f\nY: %.3f", cell.geometry!.extent.xMin, cell.geometry!.extent.yMin))
+            }
```
yo1995, line 70: It doesn't seem to be the best approach to write info into a string... Need some feedback.

arcgis-ios-sdk-samples/Layers/Identify raster cell/README.md
```diff
@@ -0,0 +1,39 @@
...
+## How to use the sample
+
+Tap or long press drag on an area of the raster to identify it and see the raster cell attributes information displayed in a callout.
+
+## How it works
+
+1. Use various delegate methods in `AGSGeoViewTouchDelegate` to get the screenPoint where a user tapped or long pressed on the map.
+2. On tap or long press drag:
+   * Dismiss the `AGSCallout`, if one is showing.
+   * Call `AGSGeoView.identifyLayer(_:screenPoint:tolerance:returnPopupsOnly:completion:)` passing in the raster layer, screen point, tolerance, and if only layers containing popups should be returned.
```
philium, line 20:
```suggestion
   * Call `AGSGeoView.identifyLayer(_:screenPoint:tolerance:returnPopupsOnly:completion:)` passing in the raster layer, screen point, tolerance, and whether only layers containing popups should be returned.
```

arcgis-ios-sdk-samples/Layers/Identify raster cell/IdentifyRasterCellViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+    /// Identify a pixel on the raster layer and display the result in a callout.
+    ///
+    /// - Parameters:
+    ///   - screenPoint: The point at which to identify features.
+    ///   - offset: An offset to the point to identify to display the callout.
+    func identifyPixel(at screenPoint: CGPoint, offset: CGPoint = .zero) {
+        mapView.identifyLayer(rasterLayer, screenPoint: screenPoint, tolerance: 1.0, returnPopupsOnly: false) { [weak self] identifyResult in
+            guard let self = self else { return }
+            if identifyResult.geoElements.isEmpty {
+                // If there are no geoelements identified, e.g. not on a raster layer, dismiss the callout.
```
philium, line 68:
```suggestion
                // If there are no geo elements identified, e.g. not on a raster layer, dismiss the callout.
```

arcgis-ios-sdk-samples/Layers/Identify raster cell/IdentifyRasterCellViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+                calloutText.append("\(attribute.key): \(attribute.value)")
+            }
+            let x = ("X", cell.geometry!.extent.xMin)
+            let y = ("Y", cell.geometry!.extent.yMin)
+            [x, y].forEach { (name, value) in
+                xyCoordinates.append("\(name): \(self.formatter.string(from: value as NSNumber)!)")
+            }
+            
+            self.calloutStackView.attributesLabel.text = calloutText.joined(separator: "\n")
+            self.calloutStackView.coordinatesLabel.text = xyCoordinates.joined(separator: "\n")
```
philium, line 87: I fear I may have given you not-so-great advice last time. Now that I see better what is going on, take a look at this way and let me know what you think:

```suggestion
            let calloutText = cell.attributes
                .map { "\($0.key): \($0.value)" }
                .joined(separator: "\n")
            let extent = cell.geometry!.extent
            let xyCoordinates = """
            X: \(self.formatter.string(from: extent.xMin as NSNumber)!)
            Y: \(self.formatter.string(from: extent.yMin as NSNumber)!)
            """
            
            self.calloutStackView.attributesLabel.text = calloutText
            self.calloutStackView.coordinatesLabel.text = xyCoordinates
```

Note: The screenshot has blank space between `calloutText` and `xyCoordinates`. Instead of adding a newline as was there before, could you just adjust the spacing between the two labels?

yo1995: Sure! I'll fix that.

philium:
> Should I include the other stack view source file in metadata and copyfile phase?

Either that or just put the `IdentifyRasterCellStackView` class in `IdentifyRasterCellViewController.swift`. I think I'd prefer the latter.

arcgis-ios-sdk-samples/Layers/Identify raster cell/IdentifyRasterCell.storyboard
```diff
@@ -0,0 +1,69 @@
...
+                        </label>
+                        <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Label" textAlignment="natural" lineBreakMode="tailTruncation" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="UA8-ez-M5c">
+                            <rect key="frame" x="0.0" y="51" width="120" height="59"/>
+                            <fontDescription key="fontDescription" type="system" pointSize="12"/>
+                            <nil key="textColor"/>
+                            <nil key="highlightedColor"/>
+                        </label>
+                    </subviews>
+                    <constraints>
+                        <constraint firstItem="UA8-ez-M5c" firstAttribute="top" secondItem="RaL-Yh-iw7" secondAttribute="bottom" constant="10" id="AhA-KH-PKH"/>
```
philium, line 56: Instead of adding a constraint, could you just modify the spacing property of the stack view?

image

arcgis-ios-sdk-samples/Layers/Identify raster cell/IdentifyRasterCell.storyboard
```diff
@@ -0,0 +1,69 @@
...
+                        </constraints>
+                        <viewLayoutGuide key="safeArea" id="cQH-id-UEf"/>
+                    </view>
+                    <connections>
+                        <outlet property="calloutStackView" destination="Wqr-jb-4el" id="iS2-bn-zE1"/>
+                        <outlet property="mapView" destination="cIA-Y7-Zgc" id="Hqg-Rn-lha"/>
+                    </connections>
+                </viewController>
+                <placeholder placeholderIdentifier="IBFirstResponder" id="P5h-DC-Dtg" userLabel="First Responder" customClass="UIResponder" sceneMemberID="firstResponder"/>
+                <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" distribution="fillProportionally" spacing="10" translatesAutoresizingMaskIntoConstraints="NO" id="Wqr-jb-4el" customClass="IdentifyRasterCellStackView" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target">
```
philium, line 39: I think you can set the alignment to Leading so that the labels are only as wide as they need to be, instead of the full width of the stack view:

image

arcgis-ios-sdk-samples/Layers/Identify raster cell/IdentifyRasterCell.storyboard
yo1995: - [x] Learned a new thing 😄

arcgis-ios-sdk-samples/Layers/Identify raster cell/IdentifyRasterCell.storyboard
yo1995: This is fixed.

In order to keep the width of callout a constant, a constraint was added to the stack view, so that when dragging on the map, the callout won't keep adjusting its width according to the width of the label (otherwise looks a bit glitchy 🤔).
 
I'm not sure about the best practice of setting up views in a stack view.  
Currently I set all the subview's height to 0, and only add a width constraint to the stack view, expecting autolayout will provide proper height for each view at runtime. However, the views does not display well in the xib because their heights are set to zero. Is this the correct approach for auto adjusting heights for the labels? Thanks.

yo1995: Sounds good. I'll revert this change.

yo1995:
@saratchandrakarumuri Please take a look at this sample 🙂 thanks!

arcgis-ios-sdk-samples/Layers/Identify raster cell/README.metadata.json
```diff
@@ -0,0 +1,36 @@
...
+    "keywords": [
+        "band",
+        "cell",
+        "cell value",
+        "continuous",
+        "discrete",
+        "identify",
+        "pixel",
+        "pixel value",
+        "raster"
```
yo1995, line 17: Eww 😞 How could I miss this?!

PR #876, opened 2020-05-12
Fix: OptionsTableViewController cannot dismiss properly

yo1995:
Problem: Cannot dismiss the `OptionsTableViewController` by tapping on the scene.

Cause: This issue was caused by not setting the `passthroughViews` property of the popover VC properly. See the screenshots below.

Fix: Setting the `passthroughViews` properly.

Also went through the app and checked all other samples that used `OptionsTableViewController` and didn't find another problem like this.

|`Bug`|`Fixed`|
|:------:|:------:|
|||


yo1995:
Also, please refer to https://github.com/Esri/arcgis-runtime-samples-ios/blob/cb63692756056d855c95b112f39d771c0949cf43/arcgis-ios-sdk-samples/Maps/Change%20basemap/SwitchBasemapViewController.swift#L81

or 

https://github.com/Esri/arcgis-runtime-samples-ios/blob/cb63692756056d855c95b112f39d771c0949cf43/arcgis-ios-sdk-samples/Layers/ArcGIS%20vector%20tiled%20layer%20(URL)/VectorTileLayerViewController.swift#L94 to see the correct versions.

PR #891, opened 2020-06-04
Fix: Nearest vertex wraparound

yo1995:
**Description**: Similar to the `Convex hull` sample, this sample also has trouble providing correct coordinates after wrapping around meridian. This bugfix normalizes the point before calling `GeometryEngine`. Also clean up formatting a bit. Refer to #882 .

**Bug**: Calculating a nearest vertex after wrapping around the meridian has incorrect behavior with points in web mercator.

**Fix**: Fixed by normalizing the map point before computing.

**Ref**: /common-samples/issues/2003

|`Before fix`|`After fix`|
|:------:|:------:|
|||

### DIFF
--- a/arcgis-ios-sdk-samples/Geometry/Nearest vertex/NearestVertexViewController.swift
+++ b/arcgis-ios-sdk-samples/Geometry/Nearest vertex/NearestVertexViewController.swift
@@ -138,9 +138,6 @@ extension NearestVertexViewController: AGSGeoViewTouchDelegate {
                     rotateOffsetWithMap: false,
                     animated: true
                 )
-            } else {
-                // Display the error as an alert if there is a problem with `AGSGeometryEngine` operations.
-                presentAlert(title: "Error", message: "Geometry Engine Failed!")
             }
         } else {
             // Dismiss the callout and reset geometry for all simple marker graphics.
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Geometry/Nearest vertex/NearestVertexViewController.swift
```diff
@@ -90,31 +101,46 @@ class NearestVertexViewController: UIViewController {
...
+                    rotateOffsetWithMap: false,
+                    animated: true
+                )
             } else {
-                // Display the error as an alert if there is a problem with nearestVertex and nearestCoordinate operation.
-                let alertController = UIAlertController(title: nil, message: "Geometry Engine Failed!", preferredStyle: .alert)
-                alertController.addAction(UIAlertAction(title: "OK", style: .default))
-                present(alertController, animated: true)
+                // Display the error as an alert if there is a problem with `AGSGeometryEngine` operations.
+                presentAlert(title: "Error", message: "Geometry Engine Failed!")
```
philium, line 143: This message could almost certainly be improved. It isn't that the geometry engine failed; it is that it was given bad input. That shouldn't ever happen, as far as I can tell, but the failure condition(s) are not documented. At the very least, it would be good to say what failed: "Could not normalize tapped point."

yo1995: In this case, I'll just totally drop the error message condition.

PR #894, opened 2020-06-10
do not show touches for indirect pointers

sbaskaran:
Show Touches is broken in Samples app with magic mouse or trackpad. Tapping/Clicking on multiple points doesn't remove the previous touches. The pointer is already visible, so we don't need to show touches in that scenario.


philium:
@vquach2404 or @yo1995: Could one of you please review this PR?

sbaskaran:
Thanks @yo1995 and @philium

PR #896, opened 2020-06-11
Update: Show location history with simulated data source

yo1995:
**Description**: This PR update the sample with `AGSSimulatedLocationDataSource`, so that it can run without actually getting live location from device. Branch: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/Fix-ShowLocationHistorySimulatedDataSource-vNext/arcgis-ios-sdk-samples/Maps/Show%20location%20history

**Ref**: /common-samples/issues/1859

|`Demo`|
|:------:|
||


yo1995:
A few things I'm uncertain about
---
- [ ] The original implementation wasn't quite right: it keeps adding graphics to the overlay when new locations are generated, which will result in out-of-memory when the count of graphics grow to certain amount. Fixed by refactoring the code.
- [ ] The Simulated location polyline source JSON is hardcoded into the sample. Should I save it as a file and add it to the project reference?
- [ ] The coding style doesn't match current standard. Should I refactor it in total? It would be hard to review, although doesn't require huge effort.

yo1995:
Please feel free to leave comments on this update! Vivian please elaborate on any other stuff that I missed in the description and questions above. Thanks! 😸

arcgis-ios-sdk-samples/Maps/Show location history/LocationHistoryViewController.swift
```diff
@@ -82,67 +122,85 @@ class LocationHistoryViewController: UIViewController {
     
     // MARK: Private behavior
     
+    private func makeDataSource(from polylineJSONString: String) -> AGSSimulatedLocationDataSource {
```
philium, line 125: FYI: Not a big deal since this is a private API, but from the [design guidelines](https://swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage):

image

So following the guidelines, this method should be named `makeDataSource(polylineJSONString: String)` and the following method should be named `makeOverlay(symbol: AGSSymbol)` or `makeOverlay(rendererSymbol: AGSSymbol)`.

arcgis-ios-sdk-samples/Maps/Show location history/LocationHistoryViewController.swift
```diff
@@ -82,67 +122,85 @@ class LocationHistoryViewController: UIViewController {
...
         }
     }
     
     private func setupMapView() {
         let map = AGSMap(basemap: .lightGrayCanvasVector())
         mapView.map = map
-
         map.load { [weak self, unowned map] (_) in
+            self?.pointBuilder = AGSMultipointBuilder(spatialReference: map.spatialReference)
             self?.trackBuilder = AGSPolylineBuilder(spatialReference: map.spatialReference)
```
philium, line 184: Repeated use of `weak` variable:

```suggestion
            guard let self = self else { return }
            let spatialReference = map.spatialReference
            self.pointBuilder = AGSMultipointBuilder(spatialReference: spatialReference)
            self.trackBuilder = AGSPolylineBuilder(spatialReference: spatialReference)
```

yo1995:
@saratchandrakarumuri Since this update involves changes to the behavior of the sample, I'd like to get feedback from you! 🙂

PR #897, opened 2020-06-12
New: Configure subnetwork trace

vquach2404:
I experimented with different types of iOS UI that were new to me. Looking forward to feedback!


vquach2404:
I made additional commits after requesting reviews to fix the project files. But now they are fixed and ready for reviews 🙂

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,391 @@
...
+        } else if let andCondition = expression as? AGSUtilityTraceAndCondition {
+            return "(\(expressionToString(expression: andCondition.leftExpression)!)) AND\n(\(expressionToString(expression: andCondition.rightExpression)!))"
+        } else if let orCondition = expression as? AGSUtilityTraceOrCondition {
+            return "(\(expressionToString(expression: orCondition.leftExpression)!)) AND\n(\(expressionToString(expression: orCondition.rightExpression)!))"
+        }
+        return nil
+    }
+    
+    // Convert the values to matching data types.
+    func convertToDataType(otherValue: Any, dataType: AGSUtilityNetworkAttributeDataType) -> Any? {
```
yo1995, line 339: It is tricky to convert string values into numbers. I tried with an evil 👿  case to input `010101010101010101010` to the value's row, and the force unwrap will crash. Guess that is overflowed.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,391 @@
...
+    }
+    
+    func trace() {
+        if utilityNetwork == nil || startingLocation == nil {
+            presentAlert(title: "Error", message: "Could not trace utility network.")
+            return
+        } else {
+            // Create utility trace parameters for the starting location.
+            let startingLocations = [startingLocation]
+            let parameters = AGSUtilityTraceParameters(traceType: .subnetwork, startingLocations: startingLocations as! [AGSUtilityElement])
```
yo1995, line 275:
```suggestion
            let parameters = AGSUtilityTraceParameters(traceType: .subnetwork, startingLocations: [startingLocation])
```
Maybe do not need casting here.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,391 @@
...
+        //Reset the labels.
+        attributeLabel.text = nil
+        comparisonLabel.text = nil
+        valueLabel.text = nil
+    }
+    
+    func trace() {
+        if utilityNetwork == nil || startingLocation == nil {
+            presentAlert(title: "Error", message: "Could not trace utility network.")
+            return
```
yo1995, line 271:
```suggestion
```
This return seem to be redundant.

Also, could the availability of utility network be decided during its `load`?

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,391 @@
...
+            }
+        })
+        // Add cancel button.
+        alert.addAction(UIAlertAction(title: "Cancel", style: .cancel))
+        
+        // Present the alert.
+        self.present(alert, animated: true, completion: nil)
+    }
+    
+    func addCondition() {
```
yo1995, line 224: After setting up all the required rows, and tap on add condition button several times, the conditions will get added to the textView below multifold. Is it possible to only add the same condition once?

image

It seems to me the logic is that user can add different conditions to one trace, so I guess it might be good to have a `Set` for conditions.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,391 @@
...
+    let assetGroupName = "Circuit Breaker"
+    let assetTypeName = "Three Phase"
+    let globalID = UUID(uuidString: "1CAF7740-0BF4-4113-8DB2-654E18800028")
+    // For creating the default trace configuration.
+    let domainNetworkName = "ElectricDistribution"
+    let tierName = "Medium Voltage Radial"
+    // An array of the types of AGSUtilityAttributeComparisonOperators as strings.
+    let comparisonsStrings = ["Equal", "NotEqual", "GreaterThan", "GreaterThanEqual", "LessThan", "LessThanEqual", "IncludesTheValues", "DoesNotIncludeTheValues", "IncludesAny", "DoesNotIncludeAny"]
+    // An array of the types of AGSUtilityAttributeComparisonOperators.
+    let comparisons: [AGSUtilityAttributeComparisonOperator] = [.equal, .notEqual, .greaterThan, .greaterThanEqual, .lessThan, .lessThanEqual, .includesTheValues, .doesNotIncludeTheValues, .includesAny, .doesNotIncludeAny]
```
yo1995, line 59: Might be better to have a dictionary than 2 arrays?

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,391 @@
...
+    var utilityNetwork: AGSUtilityNetwork?
+    // Utility element to start the trace from.
+    var startingLocation: AGSUtilityElement!
+    // Holding the initial conditional expression.
+    var initialExpression: AGSUtilityTraceConditionalExpression?
+    // The trace configuration.
+    var configuration: AGSUtilityTraceConfiguration?
+    // The source tier of the utility network.
+    var sourceTier: AGSUtilityTier?
+    // Arrays of attribues, values, and their respective labels.
```
yo1995, line 72:
```suggestion
    // Arrays of attributes, values, and their respective labels.
```

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,391 @@
...
+    // Feature service for an electric utility network in Naperville, Illinois.
+    let featureServiceURL = URL(string: "https://sampleserver7.arcgisonline.com/arcgis/rest/services/UtilityNetwork/NapervilleElectric/FeatureServer")!
+    // For creating the default starting location.
+    let deviceTableName = "Electric Distribution Device"
+    let assetGroupName = "Circuit Breaker"
+    let assetTypeName = "Three Phase"
+    let globalID = UUID(uuidString: "1CAF7740-0BF4-4113-8DB2-654E18800028")
+    // For creating the default trace configuration.
+    let domainNetworkName = "ElectricDistribution"
+    let tierName = "Medium Voltage Radial"
```
yo1995, line 55: Since most of these constants are only used during the init of utility network, it is also OK to create them as local variables. But it also makes sense to put all the required information atop, to make them more noticeable. I'll leave it to Philip.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,391 @@
...
+    @IBOutlet weak var traceButton: UITableViewCell!
+    // Reference to the text view.
+    @IBOutlet weak var textView: UITextView!
+    // References to the switch actions.
+    @IBAction func barriersSwitchAction(_ sender: UISwitch) {
+        sourceTier?.traceConfiguration?.includeBarriers = sender.isOn
+    }
+    @IBAction func containersSwitchAction(_ sender: UISwitch) {
+        sourceTier?.traceConfiguration?.includeContainers = sender.isOn
+    }
```
yo1995, line 45: I'm not sure, but it is possible to get rid of these 2 actions, and directly access `barriersSwitch.isOn` property when it is needed.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,391 @@
...
+        alert.addAction(UIAlertAction(title: "Done", style: .default) { [weak self] _ in
+            let textfield = alert.textFields![0]
+            // Check for invalid input characters.
+            if !(CharacterSet(charactersIn: "0123456789").isSuperset(of: CharacterSet(charactersIn: textfield.text!))) {
+                // Present alert to explain error.
+                self?.presentAlert(title: "This field accepts only numeric entries.")
+            } else {
+                guard let self = self else { return }
+                self.valueLabel?.text = textfield.text
+                self.tableView.reloadRows(at: [IndexPath(row: 3, section: 1)], with: .none)
```
yo1995, line 213: Aha, nearly forgot this. Will mention it to Phil after reviews all set.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
vquach2404: It produced an error without the cast I think because it is declared globally on line 65 as an optional.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
vquach2404: Availability is determined during its `load` but this is just an extra precaution as also seen in other SDK implementations. It can be removed if we deem it redundant.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
vquach2404: I originally attempted to use a dictionary but it later caused me more trouble to obtain its keys/values as strings. Please let me know if you have an alternate solution!

philium:
@vquach2404 Is this ready for me to review? I saw you requested review, but then there were subsequent changes, so I've been holding off.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
vquach2404: I spoke with Jennifer who designed this sample and confirmed that the user can add duplicates.

yo1995: OK!

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,389 @@
...
+                } else {
+                    // Prompt the user to create a custom value if none are available to select.
+                    addCustomValue()
+                }
+            } else {
+                // If no attribute is selected, present an alert.
+                presentAlert(title: "No attribute selected", message: "Please select an attribute before selecting a value.")
+            }
+        } else if cell == addConditionButton {
+            addCondition()
```
philium, line 186: The design of the samples UI is rather interesting. Once a condition is added, there's no way to update and/or remove it?

One idea is to turn the Define New Condition section into just a Conditions section. Tapping "Add Condition" would present a new view for configuring the condition to create. Upon creation of the new condition, a new row is added to the Conditions section. That row could then be deleted (via swipe-to-delete) to remove the condition. That is the more common way of doing this sort of thing and it would simplify the sample's main view controller significantly. Something to think about.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,389 @@
...
+        // Add the text field.
+        alert.addTextField { (textField) in
+            textField.placeholder = "Add a custom value"
+            textField.keyboardType = .numberPad
+        }
+        // Obtain user value.
+        alert.addAction(UIAlertAction(title: "Done", style: .default) { [weak self] _ in
+            let textfield = alert.textFields![0]
+            // Check for invalid input characters.
+            if !(CharacterSet(charactersIn: "0123456789").isSuperset(of: CharacterSet(charactersIn: textfield.text!))) {
```
philium, line 206: Whenever possible, text field input validation should be done on-demand rather than when the user taps the "Done" button. That is normally done through observing changes to the text field's text via its delegate, particularly by implementing `UITextFieldDelegate.textField(_:shouldChangeCharactersIn:replacementString:)`. Does that work with a text field in an alert? I don't know that I've ever tried to validate text in an alert's text field, but I don't see why that wouldn't work. For what it is worth, the best way I've found to validate numeric input is using a number formatter to try to create a number from a given string. There's just a couple edge cases when dealing with decimal numbers.

philium, line 206: Do negative and/or non-whole numbers need to be supported?

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,389 @@
...
+            }
+        })
+        // Add cancel button.
+        alert.addAction(UIAlertAction(title: "Cancel", style: .cancel))
+        
+        // Present the alert.
+        self.present(alert, animated: true, completion: nil)
+    }
+    
+    func addCondition() {
```
philium, line 223: I got a crash when I didn't actually select anything:

image

It would be best to only enable the "Add Condition" button when there's enough information to create a condition.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,389 @@
...
+            let parameters = AGSUtilityTraceParameters(traceType: .subnetwork, startingLocations: startingLocations as! [AGSUtilityElement])
+            parameters.traceConfiguration = configuration
+            // Trace the utility network.
+            utilityNetwork?.trace(with: parameters) { [weak self] (traceResults, error) in
+                guard let self = self else { return }
+                if let error = error {
+                    self.presentAlert(error: error)
+                } else {
+                    // Get the first result.
+                    let elementResult = traceResults?.first as! AGSUtilityElementTraceResult
```
philium, line 282: Is it possible for there to be no trace results?

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,389 @@
...
+            // Enable the buttons once all selections have been made
+            addConditionLabel.isEnabled = true
+            resetLabel.isEnabled = true
+            traceLabel.isEnabled = true
+        }
+    }
+    
+    // Convert the expression into a readable string.
+    func expressionToString(expression: AGSUtilityTraceConditionalExpression) -> String? {
+        if let categoryComparison = expression as? AGSUtilityCategoryComparison {
```
philium, line 307: I recommend switching over `expression`:

```swift
switch expression {
    case let categoryComparison as AGSUtilityCategoryComparison:
        ...
    ...
}
```

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,389 @@
...
+                if let nameOrValue = attributeComparison.otherNetworkAttribute?.name {
+                    let comparisonOperatorString = comparisonsStrings[attributeComparison.comparisonOperator.rawValue]
+                    return "`\(attributeComparison.networkAttribute.name)` \(comparisonOperatorString) `\(nameOrValue)`"
+                } else if let nameOrValue = attributeComparison.value {
+                    let comparisonOperatorString = comparisonsStrings[attributeComparison.comparisonOperator.rawValue]
+                    return "`\(attributeComparison.networkAttribute.name)` \(comparisonOperatorString) `\(nameOrValue)`"
+                }
+            }
+        } else if let andCondition = expression as? AGSUtilityTraceAndCondition {
+            return "(\(expressionToString(expression: andCondition.leftExpression)!)) AND\n(\(expressionToString(expression: andCondition.rightExpression)!))"
```
philium, line 329: Consider using a multiline string literal rather than an embedded `\n`.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,389 @@
...
+        return nil
+    }
+    
+    // Convert the values to matching data types.
+    func convertToDataType(otherValue: Any, dataType: AGSUtilityNetworkAttributeDataType) -> Any? {
+        switch dataType {
+        case .boolean:
+            return otherValue as! Bool
+        case .double:
+            if let valueString = otherValue as? String {
```
philium, line 342: When would this not be a `Double`? Same with the other cases that check for multiple types.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,389 @@
...
+        default:
+            return nil
+        }
+    }
+    
+    // Compare two values.
+    func compare(dataType: AGSUtilityNetworkAttributeDataType, comparee1: Any, comparee2: Any) -> Bool {
+        switch dataType {
+        case .boolean:
+            return isEqual(type: Bool.self, value1: comparee1, value2: comparee2)
```
philium, line 365: I think it might be simpler just to do the cast and comparison inline inside each cases rather than delegating to a separate method:

```suggestion
            return comparee1 as? Bool == comparee2 as? Bool
```

Similar with the other cases.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,389 @@
...
+    // Utility element to start the trace from.
+    var startingLocation: AGSUtilityElement?
+    // Holding the initial conditional expression.
+    var initialExpression: AGSUtilityTraceConditionalExpression?
+    // The trace configuration.
+    var configuration: AGSUtilityTraceConfiguration?
+    // The source tier of the utility network.
+    var sourceTier: AGSUtilityTier?
+    // Arrays of attributes, values, and their respective labels.
+    var values: [AGSCodedValue]?
```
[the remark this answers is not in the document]
vquach2404: This property is no longer needed so I removed it.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,389 @@
...
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let cell = tableView.cellForRow(at: indexPath)
+        tableView.deselectRow(at: indexPath, animated: true)
+        if cell == attributesCell {
+            // Create the attribute labels.
+            if attributeLabels.isEmpty {
+                attributes?.forEach { (attribute) in
+                    attributeLabels.append(attribute.name)
+                }
+            }
```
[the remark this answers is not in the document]
vquach2404: Interesting! I also decided to move some things around and make the `attributes` property local as well.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
vquach2404: I spoke with Jennifer who believes it's possible but not sure which cases will result to warnings. I've adjusted it slightly to possibly solve for this case. Let me know what you think!

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
vquach2404: I believe it depends on the type of attribute or value that was chosen. I adjusted the sample to allow negatives and decimals in which case could result in `Float`.

yo1995:
Leave some notes here. Will discuss later.

1. Dynamic table instead of static?
Is the mutating conditions rows the only reason to make it dynamic? If so, I suggest use a single row to display all conditions and make it static, to make the code look cleaner.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
yo1995: I agree. Maybe consider refactoring this sample a bit. I'll discuss with Viv to see the best approach.

yo1995:
Refer to #936

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/README.md
```diff
@@ -0,0 +1,65 @@
...
+* 'Phases Current' doesNotIncludeTheValues 'A'
+* 'Generation KW' lessThan '50'
+
+## How it works
+
+1. Create and load an `AGSUtilityNetwork` with a feature service URL, then get an asset type and a tier by their names.
+2. Populate the choice list for the comparison source with the non-system defined `AGSUtilityNetworkDefinition.networkAttributes`.  Populate the choice list for the comparison operator with the enum values from `AGSUtilityAttributeComparisonOperator`.
+3. Create an `AGSUtilityElement` from this asset type to use as the starting location for the trace.
+4. Update the selected barrier expression and the checked options in the UI using this tier's `AGSTraceConfiguration`.
+5. When 'Network Attribute' is selected, if its `AGSDomain` is an `AGSCodedValueDomain`, populate the choice list for the comparison value with its `AGSCodedValues`.  Otherwise, display a `UITextField` for entering an attribute value.
```
yo1995, line 27:
```suggestion
5. When "Network Attribute" is selected, if its `AGSDomain` is an `AGSCodedValueDomain`, populate the choice list for the comparison value with its `AGSCodedValues`. Otherwise, display a `UITextField` for entering an attribute value.
```

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/README.md
```diff
@@ -0,0 +1,65 @@
...
+6. When "Add condition" is tapped, create a new `AGSUtilityNetworkAttributeComparison` using the selected comparison source, operator, and selected or typed value. Use the selected source's `dataType` to convert the comparison value to the correct data type.
```
yo1995, line 28:
```suggestion
6. When "Add" is tapped, create a new `AGSUtilityNetworkAttributeComparison` using the selected comparison source, operator, and selected or typed value. Use the selected source's `dataType` to convert the comparison value to the correct data type.
```

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/README.md
```diff
@@ -0,0 +1,65 @@
...
+Example barrier conditions for the default dataset:
+
+* 'Transformer Load' equal '15'
+* 'Phases Current' doesNotIncludeTheValues 'A'
+* 'Generation KW' lessThan '50'
+
+## How it works
+
+1. Create and load an `AGSUtilityNetwork` with a feature service URL, then get an asset type and a tier by their names.
+2. Populate the choice list for the comparison source with the non-system defined `AGSUtilityNetworkDefinition.networkAttributes`.  Populate the choice list for the comparison operator with the enum values from `AGSUtilityAttributeComparisonOperator`.
```
yo1995, line 24:
```suggestion
2. Populate the choice list for the comparison source with the non-system defined `AGSUtilityNetworkDefinition.networkAttributes`. Populate the choice list for the comparison operator with the enum values from `AGSUtilityAttributeComparisonOperator`.
```

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/README.md
```diff
@@ -0,0 +1,65 @@
...
+3. Create an `AGSUtilityElement` from this asset type to use as the starting location for the trace.
+4. Update the selected barrier expression and the checked options in the UI using this tier's `AGSTraceConfiguration`.
+5. When 'Network Attribute' is selected, if its `AGSDomain` is an `AGSCodedValueDomain`, populate the choice list for the comparison value with its `AGSCodedValues`.  Otherwise, display a `UITextField` for entering an attribute value.
+6. When "Add condition" is tapped, create a new `AGSUtilityNetworkAttributeComparison` using the selected comparison source, operator, and selected or typed value. Use the selected source's `dataType` to convert the comparison value to the correct data type.
+7. If the traversability's list of `barriers` is not empty, create an `AGSUtilityTraceOrCondition` with the existing `barriers` and the new comparison from step 6.
+8. When "Trace" is tapped, create `AGSUtilityTraceParameters` passing in `subnetwork` and the default starting location.  Set its `traceConfiguration` with the modified options, selections, and expression; then trace the utility network with `AGSUtilityNetwork.trace(with:completion:)`.
```
yo1995, line 30:
```suggestion
8. When "Trace" is tapped, create `AGSUtilityTraceParameters` passing in `subnetwork` and the default starting location. Set its `traceConfiguration` with the modified options, selections, and expression; then trace the utility network with `AGSUtilityNetwork.trace(with:completion:)`.
```

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/README.metadata.json
```diff
@@ -0,0 +1,70 @@
...
+        "AGSUtilityTier",
+        "AGSUtilityTraceAndCondition",
+        "AGSUtilityTraceConfiguration",
+        "AGSUtilityTraceOrCondition",
+        "AGSUtilityTraceParameters",
+        "AGSUtilityTraceResult",
+        "AGSUtilityTraceType",
+        "AGSUtilityTraversability"
+    ],
+    "snippets": [
```
yo1995, line 66: Please populate metadata json again from the latest version.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/README.metadata.json
```diff
@@ -0,0 +1,70 @@
+{
+    "category": "Utility Network",
```
yo1995, line 2: It is fine for now, but if #919 merge before this one, then it will need to be updated. I'll double check once it is approved and ready to merge.

vquach2404:
@philium Ting and I have redesigned this sample. Please give it another review, thanks!

vquach2404:
@philium Ting and I have redesigned this sample. It's ready for your review, thanks!

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -1714,6 +1723,8 @@
 		F1124F0B24CFA97000E672EC /* EditFeaturesWithFeatureLinkedAnnotationViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = EditFeaturesWithFeatureLinkedAnnotationViewController.swift; sourceTree = "<group>"; };
 		F140E1EA23F35E92008AC44E /* DisplayAnnotationViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DisplayAnnotationViewController.swift; sourceTree = "<group>"; };
 		F140E1EC23F35EA6008AC44E /* DisplayAnnotation.storyboard */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; path = DisplayAnnotation.storyboard; sourceTree = "<group>"; };
+		F14A8F002474AD9D00242D96 /* ConfigureSubnetworkTrace.storyboard */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; path = ConfigureSubnetworkTrace.storyboard; sourceTree = "<group>"; };
+		F151AB8B22D7B0CA00055E45 /* MontereyElevation.dt2 */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text; name = MontereyElevation.dt2; path = "Portal Data/Rasters/MontereyElevation.dt2"; sourceTree = SOURCE_ROOT; };
```
yo1995, line 1727: Hmm I think this line is redundant. Did you pick both when merging from `v.next`?
It should be the line below.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/README.metadata.json
```diff
@@ -0,0 +1,71 @@
...
+        "AGSUtilityTraceConfiguration",
+        "AGSUtilityTraceOrCondition",
+        "AGSUtilityTraceParameters",
+        "AGSUtilityTraceResult",
+        "AGSUtilityTraceType",
+        "AGSUtilityTraversability"
+    ],
+    "snippets": [
+        "ConfigureSubnetworkTraceViewController.swift",
+        "ConfigureSubnetworkTraceOptionsViewController"
```
yo1995, line 68: Still missing the swift extension 😅 
In the future I'll make a drag and drop tool to generate metadata from folder, to replace our current script.

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
vquach2404: Yeah, I wasn't sure since the two are slightly different.

yo1995: The one with `SOURCE_ROOT` should be removed, as fixed in #942

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/README.md
```diff
@@ -0,0 +1,65 @@
...
+* 'Phases Current' doesNotIncludeTheValues 'A'
+* 'Generation KW' lessThan '50'
+
+## How it works
+
+1. Create and load an `AGSUtilityNetwork` with a feature service URL, then get an asset type and a tier by their names.
+2. Populate the choice list for the comparison source with the non-system defined `AGSUtilityNetworkDefinition.networkAttributes`. Populate the choice list for the comparison operator with the enum values from `AGSUtilityAttributeComparisonOperator`.
+3. Create an `AGSUtilityElement` from this asset type to use as the starting location for the trace.
+4. Update the selected barrier expression and the checked options in the UI using this tier's `AGSTraceConfiguration`.
+5. When an attribute has been selected, if its `AGSDomain` is an `AGSCodedValueDomain`, populate the choice list for the comparison value with its `AGSCodedValues`.  Otherwise, display a `UITextField` for entering an attribute value.
```
yo1995, line 27:
```suggestion
5. When an attribute has been selected, if its `AGSDomain` is an `AGSCodedValueDomain`, populate the choice list for the comparison value with its `AGSCodedValues`. Otherwise, display a `UITextField` for entering an attribute value.
```

philium:
It looks like the toolkit submodule has been updated:
image
That should not be part of this PR.

yo1995:
It's accidentally introduced by commit https://github.com/Esri/arcgis-runtime-samples-ios/pull/897/commits/c6e8706bfff7aad9a6d83532da19a60cdf846f92 . Let's revert that.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTrace.storyboard
```diff
@@ -0,0 +1,277 @@
...
+    <scenes>
+        <!--Configure Subnetwork Trace View Controller-->
+        <scene sceneID="Zqr-hk-gVd">
+            <objects>
+                <viewController id="eeV-QV-qKJ" customClass="ConfigureSubnetworkTraceViewController" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <view key="view" contentMode="scaleToFill" id="bNM-9x-Os8">
+                        <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <subviews>
+                            <tableView clipsSubviews="YES" contentMode="scaleToFill" alwaysBounceVertical="YES" dataMode="prototypes" style="plain" separatorStyle="default" rowHeight="-1" estimatedRowHeight="-1" sectionHeaderHeight="28" sectionFooterHeight="28" translatesAutoresizingMaskIntoConstraints="NO" id="AON-wr-WJ5">
```
philium, line 19: I recommend using a grouped style for both table views.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTrace.storyboard
```diff
@@ -0,0 +1,277 @@
...
+        <!--Add condition-->
+        <scene sceneID="cZu-hF-J2e">
+            <objects>
+                <tableViewController id="YFg-XB-smY" customClass="ConfigureSubnetworkTraceOptionsViewController" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <tableView key="view" clipsSubviews="YES" contentMode="scaleToFill" alwaysBounceVertical="YES" dataMode="static" style="plain" separatorStyle="default" rowHeight="-1" estimatedRowHeight="-1" sectionHeaderHeight="28" sectionFooterHeight="28" id="xZ5-pv-0lh">
+                        <rect key="frame" x="0.0" y="0.0" width="414" height="842"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
+                        <sections>
+                            <tableViewSection headerTitle="Condition" id="WQe-NV-8Gg">
```
philium, line 152: Since the title of the view is "Add Condition", I don't think this section needs a title.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTrace.storyboard
```diff
@@ -0,0 +1,277 @@
...
+                                    </tableViewCell>
+                                </cells>
+                            </tableViewSection>
+                        </sections>
+                        <connections>
+                            <outlet property="dataSource" destination="YFg-XB-smY" id="Vi7-bt-cbh"/>
+                            <outlet property="delegate" destination="YFg-XB-smY" id="ntn-Ai-qDU"/>
+                        </connections>
+                    </tableView>
+                    <navigationItem key="navigationItem" title="Add condition" id="OJ4-l4-Y7N">
```
philium, line 234:
```suggestion
                    <navigationItem key="navigationItem" title="Add Condition" id="OJ4-l4-Y7N">
```

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTrace.storyboard
```diff
@@ -0,0 +1,277 @@
...
+                                        <rect key="frame" x="0.0" y="115" width="414" height="59"/>
+                                        <autoresizingMask key="autoresizingMask"/>
+                                        <tableViewCellContentView key="contentView" opaque="NO" clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="center" preservesSuperviewLayoutMargins="YES" insetsLayoutMarginsFromSafeArea="NO" tableViewCell="SDQ-D2-eRK" id="4oS-Mf-8HG">
+                                            <rect key="frame" x="0.0" y="0.0" width="414" height="59"/>
+                                            <autoresizingMask key="autoresizingMask"/>
+                                            <subviews>
+                                                <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" verticalCompressionResistancePriority="751" text="Chained conditions shown here." textAlignment="center" lineBreakMode="tailTruncation" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="Vfb-sE-cuk">
```
philium, line 64: It just struck me that having the chained conditions in a separate section seems a little odd. Have you thought of putting them in a footer of the "Conditions" section?

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTrace.storyboard
```diff
@@ -0,0 +1,277 @@
...
+                            <toolbar opaque="NO" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="cNs-C6-zew">
+                                <rect key="frame" x="0.0" y="813" width="414" height="49"/>
+                                <items>
+                                    <barButtonItem title="Reset" id="Eec-Mb-rgh">
+                                        <connections>
+                                            <action selector="resetBarButtonItemTapped:" destination="eeV-QV-qKJ" id="kqe-1v-txa"/>
+                                        </connections>
+                                    </barButtonItem>
+                                    <barButtonItem style="plain" systemItem="flexibleSpace" id="RQv-69-HIS"/>
+                                    <barButtonItem systemItem="add" id="ISN-Q1-lte">
```
philium, line 97: Methinks this should be a system Done button.
image

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTrace.storyboard
yo1995: We tried to output the chained conditions to a label below the table and it doesn't look good. We'll try with the footer!

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTrace.storyboard
yo1995: Thanks for pointing that out!

 I once wondered why some Apple native tableview has rounded corner cells 😂 . We'll try that.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
```diff
@@ -0,0 +1,274 @@
...
+    
+    // MARK: UITableViewController
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        selectedAttribute = nil
+        selectedComparison = nil
+        selectedValue = nil
+    }
+
```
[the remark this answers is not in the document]
yo1995: I added these lines to trigger the didSet observer of these properties to initialize related cell UI states. 
Viv then made some updates to the code and I'm not sure if they are still needed. 🤔  We'll take another look.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/README.metadata.json
yo1995: This need to be fixed.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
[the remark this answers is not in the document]
vquach2404: These are still needed to ensure that the user's previous values aren't saved.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,387 @@
...
+        .notEqual: "NotEqual",
+        .greaterThan: "GreaterThan",
+        .greaterThanEqual: "GreaterThanEqual",
+        .lessThan: "LessThan",
+        .lessThanEqual: "LessThanEqual",
+        .includesTheValues: "IncludesTheValues",
+        .doesNotIncludeTheValues: "DoesNotIncludeTheValues",
+        .includesAny: "IncludesAny",
+        .doesNotIncludeAny: "DoesNotIncludeAny"
+    ]
```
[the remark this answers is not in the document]
vquach2404: I was able to do this for `ConfigureSubnetworkTraceViewController` but had to put another key value pair in `ConfigureSubnetworkTraceOptionsViewController` specifically for [line 157](https://github.com/Esri/arcgis-runtime-samples-ios/blob/11ed7d2407a3ac47aa6f5c68fead6a9b358b1f8c/arcgis-ios-sdk-samples/Utility%20Network/Configure%20subnetwork%20trace/ConfigureSubnetworkTraceOptionsViewController.swift#L157).

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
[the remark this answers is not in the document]
philium: How would that be? `viewDidLoad()` is only ever called once per instance of a view controller and only after the view has, well, loaded. If a new instance is being used every time, then these values would be `nil` every time `viewDidLoad()` is called. And if the same instance is used every time (which it isn't), then `viewDidLoad()` would only be called the first time the view is loaded and presented, so the previous values wouldn't be reset when the view controller is presented subsequent times. Does that make sense?

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTrace.storyboard
yo1995: A ref to auto height footer: https://medium.com/poka-techblog/uitableview-auto-sizing-header-footer-views-with-autolayout-d4a9dd0729e

yo1995: It seems the section footer's text label has pretty limited height and width; I'll instead put a custom label under the tableview to show the chained conditions.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// MARK: - UITextFieldDelegate
+
+extension ConfigureSubnetworkTraceOptionsViewController: UITextFieldDelegate {
+    // Ensure that the text field will only accept numbers.
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        let text = (textField.text! as NSString).replacingCharacters(in: range, with: string)
+        let validCharacters = ".-0123456789"
+        return CharacterSet(charactersIn: validCharacters).isSuperset(of: CharacterSet(charactersIn: text))
```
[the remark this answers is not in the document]
yo1995: Replaced by observer rule.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
```diff
@@ -0,0 +1,257 @@
...
+        }
+        // Add an observer to ensure the user does not input an empty string.
+        textFieldObserver = NotificationCenter.default.addObserver(
+            forName: UITextField.textDidChangeNotification,
+            object: nil,
+            queue: .main
+        ) { _ in
+            if let text = alertController.textFields!.first!.text {
+                // Enable the done button if the textfield is not empty and is a valid number.
+                doneAction.isEnabled = !text.isEmpty && NumberFormatter().number(from: text) != nil
```
yo1995, line 213: Actually after more thoughts, I think it doesn't need to check string isEmpty - empty string cannot convert to a number.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
```diff
@@ -0,0 +1,258 @@
...
+        }
+        optionsViewController.title = "Attributes"
+        show(optionsViewController, sender: self)
+    }
+    
+    // Transition to the comparison options view controller.
+    func showComparisonPicker() {
+        let selectedIndex = selectedComparison?.rawValue ?? -1
+        // An array of `AGSUtilityAttributeComparisonOperator` init from their raw value.
+        let attributeComparisonOperators = (0...9).map { AGSUtilityAttributeComparisonOperator(rawValue: $0)! }
```
yo1995, line 158: We discussed it earlier and I decided to make this change. Now I'm not satisfied about it. 😩 
I'll make some tweaks tomorrow.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+}
+
+/// An extension of `AGSUtilityAttributeComparisonOperator` that returns a human readable description.
+extension AGSUtilityAttributeComparisonOperator: CaseIterable {
+    // Manually set the maximum raw value of cases to facilitate protocol conformation.
+    private static let maxCaseValue = Self.doesNotIncludeAny.rawValue
+    
+    public static let allCases = {
+        return (0...maxCaseValue).map { AGSUtilityAttributeComparisonOperator(rawValue: $0)! }
+    }()
```
yo1995, line 387: To solve the comment https://github.com/Esri/arcgis-runtime-samples-ios/pull/897#discussion_r480242508, I decided to make this change, but still don't feel good about it. While it works, it has a few drawbacks

- It cannot detect if the `AGSUtilityAttributeComparisonOperator ` enum changes in the future, due to manually assigned
- By not being private, it would affect/pollute future usage of the enum

Wait for Phil's thoughts.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+    }
+}
+
+/// An extension of `AGSUtilityAttributeComparisonOperator` that returns a human readable description.
+extension AGSUtilityAttributeComparisonOperator: CaseIterable {
+    // Manually set the maximum raw value of cases to facilitate protocol conformation.
+    private static let maxCaseValue = Self.doesNotIncludeAny.rawValue
+    
+    public static let allCases = {
+        return (0...maxCaseValue).map { AGSUtilityAttributeComparisonOperator(rawValue: $0)! }
```
philium, line 386: This is a clever idea, but there's more resilient ways of getting all the cases, assuming the raw values begin at `0` are are sequential. Here's one option:

```swift
    static let allCases = Self(rawValue: 0).map { (firstCase) in
        Array(sequence(first: firstCase, next: { Self(rawValue: $0.rawValue + 1) }))
    } ?? []
```

philium (commented):
Could you please make all of the extension of `AGS` types private? Samples are meant to be, as much as possible, self-contained. Having the extension be private prevents the added members from leaking over into other samples.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
yo1995: Thanks! I actually tried a similar approach (less fancy, of course 😏), but found out that NS Enum does not give nil when init called with an invalid raw value. See the screenshot for rawValue >= 0x9

image

Therefore, the array will keep growing until crashing the app, when you tap on the comparison row in the options.

Further research told me NS Enum are similar to C enum that deal with unsafe pointers. I don't feel confident about unsafe, so I just used this uglier solution instead.

philium: That is fascinating. I thought for sure that used to work, but maybe I'm remembering incorrectly.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
```diff
@@ -256,3 +256,24 @@ class ConfigureSubnetworkTraceOptionsViewController: UITableViewController {
         }
     }
 }
+
+/// An extension of `AGSUtilityAttributeComparisonOperator` that returns a human readable description.
+private extension AGSUtilityAttributeComparisonOperator {
+    static let allCases = (Self.equal.rawValue...Self.doesNotIncludeAny.rawValue).map { AGSUtilityAttributeComparisonOperator(rawValue: $0)! }
```
philium, line 262: While this saves some typing, I'm wondering if it would be more straightforward to just use an array literal with each of the cases listed. I think that's what we've done elsewhere. An approach like this would be more useful if it was future-proof (ie. automatically included any new cases added in subsequent release), but it isn't.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
```diff
@@ -0,0 +1,279 @@
...
+        } else {
+            // Enable the value cell when an attribute is specified.
+            valueCell.textLabel?.isEnabled = false
+            valueCell.isUserInteractionEnabled = false
+        }
+    }
+    
+    // Transition to the attribute options view controller.
+    func showAttributePicker() {
+        let selectedIndex = possibleAttributes.firstIndex { $0 == selectedAttribute } ?? -1
```
philium, line 145: I didn't realize that `-1` is used as the selected index in some cases, which currently isn't supported by `OptionsTableViewController`. Due to recent changes to that class, passing `-1` as the selected index leads to a crash when a row is selected. I'll submit a PR to address that, but it will mean that this branch will need to be updated to account for those changes.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
```diff
@@ -0,0 +1,279 @@
...
+        case .notEqual: return "NotEqual"
+        case .greaterThan: return "GreaterThan"
+        case .greaterThanEqual: return "GreaterThanEqual"
+        case .lessThan: return "LessThan"
+        case .lessThanEqual: return "LessThanEqual"
+        case .includesTheValues: return "IncludesTheValues"
+        case .doesNotIncludeTheValues: return "DoesNotIncludeTheValues"
+        case .includesAny: return "IncludesAny"
+        case .doesNotIncludeAny: return "DoesNotIncludeAny"
+        @unknown default: return "Unknown"
```
philium, line 276: Should these titles be more natural? For example, "Not Equal" instead of "NotEqual"?

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
yo1995: Nice catch. I did a quick search in the project with `selectedIndex:` and it seems the existing samples doesn't use -1 to nilify the selection. After #962 is merged, we can fix it with a nil here.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceViewController.swift
```diff
@@ -0,0 +1,397 @@
...
+        case .lessThan: return "LessThan"
+        case .lessThanEqual: return "LessThanEqual"
+        case .includesTheValues: return "IncludesTheValues"
+        case .doesNotIncludeTheValues: return "DoesNotIncludeTheValues"
+        case .includesAny: return "IncludesAny"
+        case .doesNotIncludeAny: return "DoesNotIncludeAny"
+        @unknown default: return "Unknown"
+        }
+    }
+}
```
yo1995, line 397: I think it is fine to keep the title string like this, as these strings serve as the comparison operators in the trace condition expressions. They are not intended to read naturally.

yo1995 (dismissed):
I made minor changes to comment strings. Other than that, it looks good to me now. 👍  Thanks for the efforts!

yo1995 (dismissed):

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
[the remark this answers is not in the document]
yo1995: It seems like they should be replaced by a single call to `updateCellStates()`, and make sure this function handles all the cell UI states correctly.

yo1995 (dismissed):
👍

vquach2404:
@philium This is ready for another review!

yo1995:
se

While I'm playing it on smallest SE, it seems the text get clipped. I believe other previous samples might also have this problem, but should we use dynamic size for the cells?

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -881,6 +884,8 @@
 			dstPath = "";
 			dstSubfolderSpec = 7;
 			files = (
+				0094D34724E769D5007671A3 /* ConfigureSubnetworkTraceViewController.swift in CopyFiles */,
+				0094D34524E76907007671A3 /* ConfigureSubnetworkTraceOptionsViewController.swift in CopyFiles */,
```
philium, line 888: These two files have the wrong id, so when the project file is saved, Xcode removes them. You'll need to re-add these files to the Copy Files build phase.

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
```diff
@@ -0,0 +1,280 @@
...
+            if let selectedComparisonString = selectedComparison?.title {
+                comparisonCell.detailTextLabel?.text = selectedComparisonString
+            } else {
+                comparisonCell.detailTextLabel?.text = nil
+            }
+            doneBarButtonItem.isEnabled = selectedComparison != nil && selectedValue != nil
+        }
+    }
+    /// The value selected by the user.
+    var selectedValue: Any? {
```
philium, line 65: Should the type of this be `AGSCodedValue`?

philium (changes requested):
Regarding the text getting clipped, simply set the number of lines on the label to `0` so that the row's height adjusts to fit all the text:

image

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
vquach2404: No, `selectedValue` can also be a number since the user is able to input a custom value depending on which attribute is chosen.

yo1995 (approved):

philium (approved):

yo1995:
@saratchandrakarumuri Please take a look at this new sample. 😃

saratchandrakarumuri (approved):
Sample is good

PR #883, opened 2020-05-20
New: Navigate AR

yo1995:
## Description

Implementation of the `Navigate in AR` example. A brief demo below. 

A link to the branch folder: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/New-NavigateAR-vNext/arcgis-ios-sdk-samples/Augmented%20reality/Navigate%20in%20AR

**It will require a real device to test this sample.**

## Screenshots

|Navigation|
|------ |
||

Please refer to `/runtime/common-samples/issues/1965` for more screenshots.

## References and notes

- The `NavigateARCalibrationViewController.swift` is adopted from `Collect data in AR` sample.
- The original functionalities are proposed by Nathan. Big thanks for the extensive implementation details!
- Code is line wrapped at around 100 columns width.
- Segue is used to facilitate VC navigation.

PR #898, opened 2020-06-12
Fix: Navigate route

yo1995:
Refer to #837 and #883 . This PR fix coding style in the `Navigate route` sample.

Also as an additional effort to #890 , to make this sample better gradually.

### DIFF
--- a/arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
+++ b/arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
@@ -13,8 +13,8 @@
 // limitations under the License.
 
 import UIKit
-import ArcGIS
 import AVFoundation
+import ArcGIS
 
 // MARK: - Navigate route View Controller
 
@@ -78,7 +78,7 @@ class NavigateRouteViewController: UIViewController {
             mapView.locationDisplay.dataSource = makeDataSource(from: firstRoute)
             routeTracker = makeRouteTracker(result: routeResult)
             updateRouteGraphics(remaining: firstRoute.routeGeometry)
-            updateViewpoint(routeResult)
+            updateViewpoint(from: routeResult)
             // Enable bar button item.
             navigateButtonItem.isEnabled = true
         case .failure(let error):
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -13,10 +13,14 @@
 // limitations under the License.
 
 import UIKit
-import ArcGIS
 import AVFoundation
```
philium, line 16: FYI: It looks like this import isn't needed.

```suggestion
```

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -93,9 +102,9 @@ class NavigateRouteViewController: UIViewController {
     
     /// Make the simulated data source for this demo.
     ///
-    /// - Parameter result: The solved route from the route task.
+    /// - Parameter route: The first route from `AGSRouteResult` from the route task.
     /// - Returns: An `AGSSimulatedLocationDataSource` object.
-    func makeDataSource(_ route: AGSRoute) -> AGSSimulatedLocationDataSource {
+    func makeDataSource(from route: AGSRoute) -> AGSSimulatedLocationDataSource {
```
philium, line 107: This method should actually be named `makeDataSource(route: AGSRoute)`.

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -93,9 +102,9 @@ class NavigateRouteViewController: UIViewController {
     
     /// Make the simulated data source for this demo.
     ///
-    /// - Parameter result: The solved route from the route task.
+    /// - Parameter route: The first route from `AGSRouteResult` from the route task.
```
philium, line 105: This description explains where the value came from, not how the value is used in this method. A description should help the user understand the purpose of the thing being described, so something like this might be better:

```suggestion
    /// - Parameter route: An `AGSRoute` object whose geometry is used to
    /// configure the data source.
```

arcgis-ios-sdk-samples/Route & Directions/Navigate route/NavigateRouteViewController.swift
yo1995: Interesting. In the sample I used `AVSpeechSynthesizer` from [`AVFoundation`](https://developer.apple.com/documentation/avfoundation/avspeechsynthesizer), but after removing the import it still compiles with no error. 

When should I explicitly import a framework?

philium: Never mind. Even though it compiles without it (I'm not sure why), the file does rely on it, so leave it.

PR #877, opened 2020-05-14
Update: App Icon

yo1995:
**Description**: Update to the new purple icon from the creative lab.  
Should also consider if the viewer need to match the color scheme.

**Ref**: /common-samples/issues/1723

Note: This PR can be reviewed at this time, and merged later if we decide to adopt the new color scheme.

PR #899, opened 2020-06-18
[Dark Mode] Update: Content Display Logic

yo1995:
## Description

This PR adapts or updates every color in Content Display Logic folder to Dark Mode. It includes

- Change colors in `Main.storyboard` for each custom color in view controllers
- Add a color to `AppDelegate.swift`
- Change color for color picker view controller
- Change colors for the search bar, which is used both in content collection view controller (the main screen card view) and in `SearchForWebmapByKeywordViewController.swift`
- Change CSS styles for README and source code panes, so that they are responsive to Dark Mode
- Update colors in download progress view - for on-demand resource dependent samples

You can use <kbd>command</kbd> + <kbd>shift</kbd> + <kbd>A</kbd> to toggle Dark Mode in Simulator.

Also, I removed the force Light Mode flag, so there is no need to edit anything when testing! Reasons [below](https://github.com/Esri/arcgis-runtime-samples-ios/pull/899#discussion_r503593806).

## Does not change

- DemoTouchManager to show touches, no color is changed in this class.
- SVProgressHUD, it is a 3rd party dependency, and the author hasn't adapted it to iOS 13 yet. We consider replacing it with our own Swift implementation in total. It is used in many samples, including Trace utility network and Perform valve isolation trace.

## References

/common-samples/issues/1507
/common-samples/issues/2037

## Demo

201012-demo

### DIFF
--- a/arcgis-ios-sdk-samples/AppDelegate.swift
+++ b/arcgis-ios-sdk-samples/AppDelegate.swift
@@ -177,4 +177,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
 extension UIColor {
     // Also used as global tint/accent color.
     class var accentColor: UIColor { return UIColor(named: "AccentColor")! }
+    // The translucent background color for status labels.
+    class var statusLabelBackgroundColor: UIColor { return UIColor(named: "statusLabelBackgroundColor")! }
 }
--- a/arcgis-ios-sdk-samples/Content Display Logic/Controllers/ListViewController.swift
+++ b/arcgis-ios-sdk-samples/Content Display Logic/Controllers/ListViewController.swift
@@ -39,7 +39,6 @@ class ListViewController: UITableViewController {
         let cell = tableView.dequeueReusableCell(withIdentifier: "ListCell", for: indexPath)
         
         cell.textLabel?.text = list[indexPath.row]
-        cell.backgroundColor = .clear
         return cell
     }
     
--- a/arcgis-ios-sdk-samples/Content Display Logic/Controllers/SampleInfoViewController.swift
+++ b/arcgis-ios-sdk-samples/Content Display Logic/Controllers/SampleInfoViewController.swift
@@ -17,21 +17,16 @@ import WebKit
 
 class SampleInfoViewController: UIViewController {
     /// The web view that displays the readme.
-    @IBOutlet private weak var webView: WKWebView!
+    @IBOutlet var webView: WKWebView! {
+        didSet {
+            webView.navigationDelegate = self
+        }
+    }
     
     var readmeURL: URL?
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        
-        // We must construct the web view in code as long as we support iOS 10.
-        // Prior to iOS 11, there was a bug in WKWebView.init(coder:) that
-        // caused a crash.
-        let webView = WKWebView(frame: view.bounds)
-        webView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
-        webView.navigationDelegate = self
-        view.addSubview(webView)
-        self.webView = webView
         
         if let readmeURL = readmeURL,
             let html = markdownTextFromFile(at: readmeURL) {
@@ -53,7 +48,8 @@ class SampleInfoViewController: UIViewController {
     }
     
     func displayHTML(_ readmeContent: String) {
-        let cssPath = Bundle.main.path(forResource: "style", ofType: "css") ?? ""
+        let cssPath = Bundle.main.path(forResource: "style", ofType: "css")!
+        let darkmodePath = Bundle.main.path(forResource: "darkmode", ofType: "css")!
         let string = """
             <!doctype html>
             <html>
@@ -61,6 +57,7 @@ class SampleInfoViewController: UIViewController {
                 <link rel="stylesheet" href="\(cssPath)">
                 <link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/foundation/5.5.2/css/foundation.min.css">
                 <link rel="stylesheet" type="text/css" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css">
+                <link rel="stylesheet" href="\(darkmodePath)">
                 <meta name="viewport" content="initial-scale=1, width=device-width, height=device-height">
             </head>
             <body>
--- a/arcgis-ios-sdk-samples/Content Display Logic/Controllers/SourceCodeViewController.swift
+++ b/arcgis-ios-sdk-samples/Content Display Logic/Controllers/SourceCodeViewController.swift
@@ -102,7 +102,12 @@ class SourceCodeViewController: UIViewController, UIAdaptivePresentationControll
         if filenames.count > 1 {
             titleString = String(format: "%@ %@", (arrowPointingDown ? "▶︎" : " \u{25B4}"), filename)
         } else {
-            toolbarTitleButton.setTitleTextAttributes([.foregroundColor: UIColor.black], for: .normal)
+            if #available(iOS 13.0, *) {
+                toolbarTitleButton.setTitleTextAttributes([.foregroundColor: UIColor.label], for: .normal)
+            } else {
+                // Fallback on earlier versions
+                toolbarTitleButton.setTitleTextAttributes([.foregroundColor: UIColor.black], for: .normal)
+            }
         }
         toolbarTitleButton.title = titleString
     }
--- a/arcgis-ios-sdk-samples/Content Display Logic/Controllers/SourceCodeViewController.swift
+++ b/arcgis-ios-sdk-samples/Content Display Logic/Controllers/SourceCodeViewController.swift
@@ -17,10 +17,21 @@ import WebKit
 
 class SourceCodeViewController: UIViewController, UIAdaptivePresentationControllerDelegate {
     /// The view to which the web view is added.
-    @IBOutlet private weak var contentView: UIView!
+    @IBOutlet var contentView: UIView!
     /// The web view that displays the source code.
-    @IBOutlet private weak var webView: WKWebView!
-    @IBOutlet private weak var toolbarTitleButton: UIBarButtonItem!
+    @IBOutlet var webView: WKWebView!
+    @IBOutlet var toolbarTitleButton: UIBarButtonItem! {
+        didSet {
+            if filenames.count <= 1 {
+                if #available(iOS 13.0, *) {
+                    toolbarTitleButton.tintColor = UIColor.label
+                } else {
+                    toolbarTitleButton.tintColor = UIColor.black
+                }
+            }
+            toolbarTitleButton.possibleTitles = Set(filenames)
+        }
+    }
     
     private var listViewController: ListViewController!
     private var selectedFilenameIndex = 0
@@ -29,15 +40,6 @@ class SourceCodeViewController: UIViewController, UIAdaptivePresentationControll
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        
-        // We must construct the web view in code as long as we support iOS 10.
-        // Prior to iOS 11, there was a bug in WKWebView.init(coder:) that
-        // caused a crash.
-        let webView = WKWebView(frame: contentView.bounds)
-        webView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
-        webView.scrollView.alwaysBounceHorizontal = true
-        contentView.addSubview(webView)
-        self.webView = webView
         
         if let filename = currentFilename {
             loadHTMLPage(filename: filename)
@@ -80,11 +82,10 @@ class SourceCodeViewController: UIViewController, UIAdaptivePresentationControll
             cssPath = Bundle.main.path(forResource: "xcode", ofType: "css")!
         }
         let jsPath = Bundle.main.path(forResource: "highlight.pack", ofType: "js")!
-        let scale = UIDevice.current.userInterfaceIdiom == .phone ? "0.5" : "1.0"
         let stringForHTML = """
             <html>
             <head>
-                <meta name='viewport' content='width=device-width, initial-scale='\(scale)'/>
+                <meta name="viewport" content="initial-scale=1, width=device-width, height=device-height">
                 <link rel="stylesheet" href="\(cssPath)">
                 <script src="\(jsPath)"></script>
                 <script>hljs.initHighlightingOnLoad();</script>
@@ -98,16 +99,11 @@ class SourceCodeViewController: UIViewController, UIAdaptivePresentationControll
     }
     
     func setupToolbarTitle(_ filename: String, arrowPointingDown: Bool) {
-        var titleString = filename
+        let titleString: String
         if filenames.count > 1 {
             titleString = String(format: "%@ %@", (arrowPointingDown ? "▶︎" : " \u{25B4}"), filename)
         } else {
-            if #available(iOS 13.0, *) {
-                toolbarTitleButton.setTitleTextAttributes([.foregroundColor: UIColor.label], for: .normal)
-            } else {
-                // Fallback on earlier versions
-                toolbarTitleButton.setTitleTextAttributes([.foregroundColor: UIColor.black], for: .normal)
-            }
+            titleString = filename
         }
         toolbarTitleButton.title = titleString
     }
@@ -138,6 +134,6 @@ class SourceCodeViewController: UIViewController, UIAdaptivePresentationControll
     // MARK: - UIAdaptivePresentationControllerDelegate
     
     func adaptivePresentationStyle(for controller: UIPresentationController, traitCollection: UITraitCollection) -> UIModalPresentationStyle {
-        return UIModalPresentationStyle.none
+        return .none
     }
 }
--- a/arcgis-ios-sdk-samples/Content Display Logic/Controllers/SourceCodeViewController.swift
+++ b/arcgis-ios-sdk-samples/Content Display Logic/Controllers/SourceCodeViewController.swift
@@ -25,6 +25,7 @@ class SourceCodeViewController: UIViewController, UIAdaptivePresentationControll
     private var listViewController: ListViewController!
     private var selectedFilenameIndex = 0
     var filenames = [String]()
+    private var currentFilename: String!
     
     override func viewDidLoad() {
         super.viewDidLoad()
@@ -38,16 +39,25 @@ class SourceCodeViewController: UIViewController, UIAdaptivePresentationControll
         contentView.addSubview(webView)
         self.webView = webView
         
-        if let filename = filenames.first {
+        if let filename = currentFilename {
+            loadHTMLPage(filename: filename)
+        } else if let filename = filenames.first {
+            currentFilename = filename
             loadHTMLPage(filename: filename)
         }
     }
     
+    // Change the highlight.js rendered HTML when switch between modes.
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        loadHTMLPage(filename: currentFilename)
+    }
+    
     func loadHTMLPage(filename: String) {
-        if let content = self.contentOfFile(filename) {
-            self.setupToolbarTitle(filename, arrowPointingDown: true)
-            let htmlString = self.htmlStringForContent(content)
-            self.webView.loadHTMLString(htmlString, baseURL: URL(fileURLWithPath: Bundle.main.bundlePath))
+        if let content = contentOfFile(filename) {
+            setupToolbarTitle(filename, arrowPointingDown: true)
+            let htmlString = htmlStringForContent(content)
+            webView.loadHTMLString(htmlString, baseURL: URL(fileURLWithPath: Bundle.main.bundlePath))
         }
     }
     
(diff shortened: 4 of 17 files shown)


yo1995 (commented):
~~The first round of changes I made to the main display logic. After the meeting, I feel it would more approachable with smaller PRs that involves less changed files, so I'll discuss with Sarat and try to figure out the best approach.~~

outdated

arcgis-ios-sdk-samples/Content Display Logic/Controllers/SourceCodeViewController.swift
```diff
@@ -38,16 +39,25 @@ class SourceCodeViewController: UIViewController, UIAdaptivePresentationControll
...
             loadHTMLPage(filename: filename)
         }
     }
     
+    /// Change the highlight.js rendered HTML when switch  between modes.
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        loadHTMLPage(filename: currentFilename)
+    }
+
```
yo1995, line 55: This ensures the css get loaded every time we swap the source files, so that the source code pane always be in correct color mode.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/SampleInfoViewController.swift
```diff
@@ -54,13 +54,15 @@ class SampleInfoViewController: UIViewController {
...
         let cssPath = Bundle.main.path(forResource: "style", ofType: "css") ?? ""
+        let darkmodePath = Bundle.main.path(forResource: "darkmode", ofType: "css") ?? ""
         let string = """
             <!doctype html>
             <html>
             <head>
                 <link rel="stylesheet" href="\(cssPath)">
                 <link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/foundation/5.5.2/css/foundation.min.css">
                 <link rel="stylesheet" type="text/css" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css">
+                <link rel="stylesheet" href="\(darkmodePath)">
```
yo1995, line 65: Add an additional css to only adapt a few colors to dark mode automatically, to avoid changing the original css style.

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/Main.storyboard
```diff
@@ -267,17 +260,17 @@
...
                         <subviews>
                             <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Select a sample from the list" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="99s-uK-gYc">
-                                <rect key="frame" x="79.666666666666686" y="395" width="217" height="21"/>
+                                <rect key="frame" x="79.666666666666686" y="368" width="217" height="21"/>
                                 <fontDescription key="fontDescription" type="system" pointSize="17"/>
-                                <color key="textColor" red="0.66666666666666663" green="0.66666666666666663" blue="0.66666666666666663" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                <color key="textColor" systemColor="secondaryLabelColor" red="0.23529411759999999" green="0.23529411759999999" blue="0.26274509800000001" alpha="0.59999999999999998" colorSpace="custom" customColorSpace="sRGB"/>
```
yo1995, line 269: This is the detail label under a certain sample's cell in the list.

image

I decide to change it to a slightly different gray color, as it better fits the definition of `secondaryLabelColor`.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ListViewController.swift
```diff
@@ -39,7 +39,6 @@ class ListViewController: UITableViewController {
         let cell = tableView.dequeueReusableCell(withIdentifier: "ListCell", for: indexPath)
         
         cell.textLabel?.text = list[indexPath.row]
-        cell.backgroundColor = .clear
```
yo1995, line 42: No need to explicitly set this background color - bgc for cell is default to clear color.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/SampleInfoViewController.swift
```diff
@@ -24,14 +24,8 @@ class SampleInfoViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         
-        // We must construct the web view in code as long as we support iOS 10.
-        // Prior to iOS 11, there was a bug in WKWebView.init(coder:) that
-        // caused a crash.
-        let webView = WKWebView(frame: view.bounds)
         webView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
         webView.navigationDelegate = self
```
yo1995, line 28: Since we already dropped support for below iOS 12, there's no need to do this. I've added a WKWebView to main.storyboard and connected with an IBOutlet.

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/Main.storyboard
```diff
@@ -357,13 +350,31 @@
                     <view key="view" contentMode="scaleToFill" id="p2z-Do-b04">
                         <rect key="frame" x="0.0" y="0.0" width="375" height="724"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
-                        <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                        <subviews>
```
yo1995, line 353: The IBOutlet of WKWebView added to storyboard.

yo1995 (commented):
~~This will be quite an effort to review and verify. Please take your time to build, look around and scrutinize these changes. 🙏 Later will check with Sarat to ensure they do not break anything.~~

outdated

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -169,9 +169,38 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
...
+    static let secondaryBlue: UIColor = { return #colorLiteral(red: 0, green: 0.368, blue: 0.584, alpha: 1) }()
+    
+    static let systemBackgroundWithOpacity: UIColor = {
+        if #available(iOS 13, *) {
+            return UIColor.systemBackground.withAlphaComponent(0.6)
+        } else {
+            return #colorLiteral(red: 1, green: 1, blue: 1, alpha: 0.6)
+        }
+    }()
+
```
yo1995, line 182: I'm not pretty sure about the changes here. I would like to make them `IBInspectable`, but using Color assets makes sense as well: https://nshipster.com/dark-mode/#manage-colors-with-an-asset-catalog.

Should I instead create these colors as asset catalog to avoid future re-visits?
Using asset catalog would also be beneficial for changing theme colors.

yo1995: Update: this change will get reverted and created as a color asset catalog.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/SourceCodeViewController.swift
```diff
@@ -63,7 +73,12 @@ class SourceCodeViewController: UIViewController, UIAdaptivePresentationControll
     }
     
     func htmlStringForContent(_ content: String) -> String {
-        let cssPath = Bundle.main.path(forResource: "xcode", ofType: "css") ?? ""
+        let cssPath: String
+        if traitCollection.userInterfaceStyle == .dark {
+            cssPath = Bundle.main.path(forResource: "solarized-dark", ofType: "css") ?? ""
+        } else {
+            cssPath = Bundle.main.path(forResource: "xcode", ofType: "css") ?? ""
```
philium, line 80: Instead of specifying a default value in both places, you could make `cssPath` optional and then specify a default value at the point of use.

arcgis-ios-sdk-samples/Content Display Logic/Views/DownloadProgressView.swift
```diff
@@ -102,7 +102,7 @@ class DownloadProgressView: UIView {
         self.transparentShapeLayer.frame = CGRect(x: 0, y: 0, width: 2 * self.radius, height: 2 * self.radius)
         self.transparentShapeLayer.path = UIBezierPath(arcCenter: CGPoint(x: self.radius, y: self.radius), radius: self.radius, startAngle: 0, endAngle: 2 * CGFloat.pi, clockwise: true).cgPath
         self.transparentShapeLayer.fillColor = UIColor.clear.cgColor
-        self.transparentShapeLayer.strokeColor = UIColor(red: 0, green: 0, blue: 0, alpha: 0.3).cgColor
+        self.transparentShapeLayer.strokeColor = UIColor.white.cgColor
```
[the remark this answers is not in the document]
philium: Does this mean the layer is no longer transparent? If so, should it be renamed to reflect that?

yo1995:
> It seems a little odd that the navigation bar (other than the search text field, which isn't always visible) doesn't change in dark mode. I haven't looked at it yet, but I'm guessing we'll have separate branding colors in light and dark mode? Either way, I was thinking about the title text color. White in dark mode looks odd to me. Right now at least, because of the bar color, it is basically the inverse of `titleTextColor`. We could actually make it the inverse in dark mode as well by using a dynamic color:

My thoughts keep growing after discussions about UI color choices. 🤔 💭 

Esri does have a universal palette for different colors, but not for Dark Mode. I took an approach to pair up Dark Mode colors described here: https://github.com/Esri/arcgis-runtime-samples-ios/pull/915#discussion_r458290518

For the nav bar
- its background color should be different in light and dark. I've come up with a `primaryThemeColor` same as the logo color in #915 that both adapt light and dark mode
- title text color should always be white/light gray, given that the background color will be purple. I tried with black text in dark mode and it doesn't look good. I also tried other apps with theme color on my phone, such as Slack that uses a violet red theme - it keeps the title as light color both in light and dark mode.

Since Dark Mode is not as prioritized as App Icon and theme color update, we might consider push back this update and first go with #877 and #915. After setting up a universal color assets catalog, we can directly use named colors for convenience.

I'll convert this PR back to draft and revisit afterwards.

arcgis-ios-sdk-samples/Content Display Logic/Views/DownloadProgressView.xib
```diff
@@ -33,21 +32,21 @@
...
+                            <rect key="frame" x="91.5" y="152" width="57" height="33"/>
                             <fontDescription key="fontDescription" type="boldSystem" pointSize="17"/>
                             <state key="normal" title="Cancel"/>
                             <connections>
                                 <action selector="cancelAction" destination="-1" eventType="touchUpInside" id="0Uh-4A-vKB"/>
                             </connections>
                         </button>
                     </subviews>
-                    <color key="backgroundColor" white="1" alpha="1" colorSpace="calibratedWhite"/>
+                    <color key="backgroundColor" systemColor="secondarySystemBackgroundColor" red="0.94901960780000005" green="0.94901960780000005" blue="0.96862745100000003" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
```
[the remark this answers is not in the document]
yo1995: Ahh it gonna be hard to resolve these merge conflicts! 😩

yo1995 (commented):
@vquach2404 @philium I've revamped this PR and make the content display logic Dark Mode compatible. Please give it a try! 🦉 

Thanks.

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/Main.storyboard
```diff
@@ -348,41 +344,69 @@
...
                                 <constraints>
-                                    <constraint firstAttribute="height" constant="44" id="Kff-Ig-9pK"/>
+                                    <constraint firstAttribute="bottom" secondItem="KE4-Ic-Gt2" secondAttribute="bottom" id="ApL-oy-iHl"/>
+                                    <constraint firstItem="KE4-Ic-Gt2" firstAttribute="leading" secondItem="rA3-JQ-Mdx" secondAttribute="leading" id="a3B-K1-Iw3"/>
+                                    <constraint firstAttribute="trailing" secondItem="KE4-Ic-Gt2" secondAttribute="trailing" id="nxz-eX-LGE"/>
+                                    <constraint firstItem="KE4-Ic-Gt2" firstAttribute="top" secondItem="rA3-JQ-Mdx" secondAttribute="top" id="sqb-Ou-quV"/>
                                 </constraints>
+                            </view>
+                            <toolbar opaque="NO" clipsSubviews="YES" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="cZS-cc-Di0">
+                                <rect key="frame" x="0.0" y="764" width="414" height="44"/>
```
yo1995, line 409: There is an issue with the toolbar bottom constraints right now. Since the Source Code View Controller is embedded in a container view, storyboard fails to snap the safe area with its superview.

image

I've asked for help from our instructor in last training, but feel free to give me some help/thoughts!

vquach2404 (commented):
I tested it and it everything looks good besides the toolbar issue you mentioned. You may have more luck getting a response from our course leader if you direct message him or mention him in your message.

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/Main.storyboard
philium: I'm not seeing that with an iOS 14 simulator. Which simulator are you using?

image

philium: Never mind. I didn't see that some changes had been force pushed. After resetting my local branch, I now see the issue. I'll take a look.

yo1995: Sorry for the force push. There were quite a few conflicts and I decided to ditch the outdated changes and redo from scratch. 

The previous version set the container view's background to the same as the toolbar's bar tint to hide this problem, but that won't work under Dark Mode.

The workaround that I came up with is to 

1. Uncheck the translucent checkbox of toolbar, so that it doesn't have the UIVisualEffect view for translucency
2. Set its bar tint to the same as the base/root view's background color, such as `systemBackgroundColor`. 

However, toolbars in samples or in source code pane shouldn't be different in colors, so I didn't make this change.

yo1995: Solved by making a small tweak in c27369e 😏

yo1995:
@philium I tweaked the toolbar for source code pane and now it looks correct. 

In terms of changing the info button segue into present modally, I incline to stick to the current practice, i.e. tapping the info button push onto nav stack. 
<details>

There might be some samples that depends on the `viewDidAppear(_:)` method, and [iOS 13's interactive dismissal](https://developer.apple.com/documentation/uikit/uiviewcontroller/3229894-modalinpresentation) does not call it. Unless we make it a fullscreen modal presentation - but that doesn't make a huge difference to our current "show/push" approach.
<details>

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/Main.storyboard
yo1995: |Translucent|Opaque|
|-|-|
|image|image|

One more thing to mention: while the purple tint looks beautiful on an opaque bar, it might be a bit bright on a translucent bar. Should I tweak the accentColor for Dark Appearance?

philium: Yeah, looks like that might be a good idea.

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -177,4 +177,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
 extension UIColor {
     // Also used as global tint/accent color.
     class var accentColor: UIColor { return UIColor(named: "AccentColor")! }
+    // The translucent background color for status labels.
+    class var statusLabelBackgroundColor: UIColor { return UIColor(named: "statusLabelBackgroundColor")! }
```
philium, line 181: It looks like this property isn't being used. Am I missing something?

yo1995: I will need this color to replace (unify) many status label's background color in each category. I add it now so that we can branch from `v.next` to update Dark Mode for multiple categories in parallel, after this PR is merged.

If I add this color in a category's PR where it is first needed, it means we have to wait until that category's PR is merged before we can use this color in other categories.

What do you think?

philium: I'm a big fan of not introducing code that isn't actually used yet. Plans change all the time and if the code ends up not getting used, it becomes a liability rather than an asset. However, I understand why you added this property now and I think that makes sense.

philium (approved):

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/Main.storyboard
yo1995: Slightly adjusted the purple color in Dark Mode
|Before|After|
|-|-|
||image|

arcgis-ios-sdk-samples/AppDelegate.swift
yo1995: Thanks!

yo1995:
@saratchandrakarumuri We will continuously make progress to adapt to Dark Mode in the following couple of sprints! 

Please take a look at this PR to ensure the main screens (except sample-specific screens) are adapted correctly. Thanks! 🦉

saratchandrakarumuri:
* Download progress bar is not clearly visible in dark mode
IMG_0C054881D576-1

* In full screen mode, current dark mode looks good, but on an iPad in split mode, colors look little clumsy with multiple shades of dark

IMG_0119
IMG_0120

yo1995:
@saratchandrakarumuri 

* Download progress bar is not clearly visible in dark mode

progressView

I tried multiple pairs of grayscale colors with no luck. 🤕 None of them look perfect in both Light and Dark Mode. 
So this time I replaced the progressing stroke with green. Does using green to represent progress make sense? @philium I thought to use the purple accent color for progress, but I remember that color should only used by interact-able elements on the screen.

* In full screen mode, current dark mode looks good, but on an iPad in split mode, colors look little clumsy with multiple shades of dark

Fixed.

philium:
> Does using green to represent progress make sense?

I think the green looks fine.

philium (dismissed):

yo1995:
@saratchandrakarumuri This is ready for another look. 👀

saratchandrakarumuri:
@yo1995 Sorry forgot to submit the comment 🙈 ...
* Download progress bar looks good
* I don't know the design decisions and effort behind it, but even though now the while viewer is same color I feel grey color is easier on eyes than solid black... have you tried using grey everywhere?

yo1995:
> I don't know the design decisions and effort behind it, but even though now the while viewer is same color I feel grey color is easier on eyes than solid black... have you tried using grey everywhere?

This makes sense. In fact some apps does use [`secondarySystemBackground`](https://developer.apple.com/documentation/uikit/uicolor/3173137-secondarysystembackground) (dark gray) as the major dark bgc. Currently for most places I just used the default color recommended by Apple. @philium What's your suggestion?
I'll make another commit to switch gray and give it a try, and we can go back if it doesn't look well.

yo1995:
@saratchandrakarumuri I've changed some of the solid black colors into dark gray, and it does look softer. 

@vquach2404 @philium How do you like it comparing to the previous full black `#000000`? 
I used the `tertiarySystemBackgroundColor` (`#FFFFFF`,
`#2C2C2E`) for the main collection view, version info VC as well as source code and README background color.

Simulator Screen Shot - iPhone 11 - 2020-10-28 at 14 55 13

If we don't want this change, it is easy to just revert 4ad2c34 to go back.

vquach2404:
I think the new gray color looks great! I noticed that the list of samples is still black. Will that be changing to gray as well?

yo1995:
> I think the new gray color looks great! I noticed that the list of samples is still black. Will that be changing to gray as well?

I left that as solid black `#000000` on purpose. 
We uses lots of tableviews in our app, and I tend to set all of them with a single color/style, so I didn't change its default settings (which is black by default and dark gray when row selected).

But it is very easy to set it to gray. Wait for more input from Phil and Sarat.

saratchandrakarumuri:
@yo1995 If we change that one to grey like the rest of the viewer, what would be the selection color?

vquach2404 (commented):
Simulator Screen Shot - iPhone 12 Pro Max - 2020-10-29 at 16 53 59
This looks better! My concern is about the color of buttons. Although the button should be a different color than the background, the black against the gray seems slightly out of place. Could we make buttons just a darker gray?

yo1995:
> This looks better! My concern is about the color of buttons. Although the button should be a different color than the background, the black against the gray seems slightly out of place. Could we make buttons just a darker gray?

@vquach2404  Do you mean making the toolbar's background a gray color, instead of solid black? That is quite hard to achieve. 🤔 

view hierarchy

The toolbar by default does not have a background color; instead, it is "translucent" like a frosted glass. So the color is actually a Gaussian blur of the layers beneath. See the view hierarchy above.

The reason why it is solid black is that currently the Apple default background color for the root views for each sample (as well as the README and source code panes) is `systemBackground` - which is solid black and solid white.

gray-toolbar

After changing the root view's bgc to gray it looks like above. It is undesirable to have toolbars with inconsistent colors. We can go with 1 of the 3 options below.

- Keep it as is. 
- Override the toolbar background color in AppDelegate. 
- Changing each sample's root view's bgc to `tertiarySystemBackground` so that toolbar blurs to dark gray.

I tend to keep it as is, so that we can always use the default color whenever we add new samples or edit existing ones.

vquach2404:
> * Override the toolbar background color in AppDelegate.
I personally find the drastic contrast of the toolbar against the background odd so this option seems best to me. Does anyone else feel the same @saratchandrakarumuri  @philium ? If not, we'll keep the toolbars the way it is.

philium:
I too think the toolbar looks odd showing up as black. If we could make it gray somehow, that would probably be better.

yo1995:
Ready for another review. If we are happy with the new colors, we'll proceed to update all categories in the following days before release.

- Fix the colors in samples for Dark Mode
- Update the screenshots to reflect new colors
- Fix any wrong constraints in the storyboard that affects appearance

Details at: https://github.com/Esri/arcgis-runtime-samples-ios/pull/915#issuecomment-692327353

saratchandrakarumuri (approved):
New look is good and consistent

vquach2404 (approved):
Thanks for making that change, it looks good!

yo1995:
I'm very sorry to find that I got the background color name wrong in a3bf026b 🤦‍♂️🤦‍♂️

```diff
- statusLabelBackgroudColor
+ statusLabelBackgroundColor
```

Will log an issue for it...

PR #900, opened 2020-06-19
Fix: Update highlight.js to fix code renderer

yo1995:
**Description**: This PR updates the `highlight.js` script in the app, which is used to highlight swift source code in the code pane for each sample.

**Bug**: The older version has trouble rendering multiline string literals, which leaves the whole file un-rendered.

**Fix**: Fixed by updating the version of the JavaScript source file to latest release version of v10.1.1 .

- A few samples suffer from this issue before:
    - Navigate route
    - Load WFS with XML query
    - Raster function (service)

**Ref**: /runtime/common-samples/issues/1310

Configuration of the new `highlight.js`. Download from https://highlightjs.org/download/

Note: It is now a Swift only version.

conf

|`Before fix`|`After fix`|
|:------:|:------:|
|before|after|


philium:
We don't need Objective-C support, do we?

yo1995:
> We don't need Objective-C support, do we?

It seems we don't, but the original script does include highlighting for ObjC. Below is a screenshot of the previous version script that also register for OC.

image

 Do I create a Swift-only version of `highlight.js`?

philium:
I don't see any reason why we would need Objective-C support. I'm also not sure that we need JSON support. What were you thinking with regards to that? I'd say leave it the way it was (without JSON support) and we can always add it in if/when needed.

philium:
I'm guessing the script was used back when the samples were written in Objective-C, Swift support was added during the transition from Objective-C to Swift, and Objective-C was just never taken out after it was no longer needed.

yo1995:
> I'm guessing the script was used back when the samples were written in Objective-C, Swift support was added during the transition from Objective-C to Swift, and Objective-C was just never taken out after it was no longer needed.

OK then I'll update with Swift only.

PR #869, opened 2020-05-08
Update READMEs

philium:
This PR updates `v.next` with the changes from the `readme-updates` branch.

PR #901, opened 2020-06-19
Fix: Issues in readme updates

yo1995:
This PR addressed the remaining issues in #869 which are found during the review of preview developer site build.


arcgis-ios-sdk-samples/Features/Time based query/README.md
```diff
@@ -18,8 +18,8 @@ Run the sample, and a subset of records will be displayed on the map. Pan and zo
 2. The `featureRequestMode` of the `AGSServiceFeatureTable` is set to `manualCache`, so that the developer can control how and when the feature table is populated with data.
 3. An `AGSFeatureLayer` is created by passing in the instance of the `AGSServiceFeatureTable`.
 4. An `AGSTimeExtent` object is created by specifying start and end date/time objects.
-5. Apply the `AGSTimeExtent` to an `AGSQueryParmaters` object.
-6. Pass the `AGSQueryParmaters` into `AGSServiceFeatureTable.populateFromService(with:clearCache:outFields:completion:)`.
+5. 5. Apply the `AGSTimeExtent` to an `AGSQueryParameters` object.
```
philium, line 21: Is there supposed to be an extra `5.` here?

yo1995: 🤮 Copy-paste fat-finger slipped in. Fixed.

yo1995:
> In the new ArcGIS Vector Tiled Layer (Custom Style) screenshots, it looks like the toolbar at the bottom doesn't extend into the safe area. Do you know if there's an issue for that? If not, could you please create one?

Oops, thanks for pointing that out. I replaced the screenshots with correct ones.

It is because I snapped the images on Dark Mode branch, where I changed the palette and make the bar tint light gray instead of pure white. And the `ArcGIS Vector Tiled Layer (Custom Style)` sample assigned a custom white background color for the parent view.

This will be fixed as part of the huge walkthrough of adapting every sample to correct semantic colors.

PR #902, opened 2020-06-19
Show a popup implementation

vquach2404:
I just merged the design so I'm adding it to iOS as well! Looking forward to feedback!


arcgis-ios-sdk-samples/Display information/Show a popup/ShowPopup.storyboard
```diff
@@ -0,0 +1,42 @@
...
+                                <rect key="frame" x="-20" y="-20" width="454" height="936"/>
+                                <color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
+                            </view>
+                        </subviews>
+                        <color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
+                        <constraints>
+                            <constraint firstItem="sst-pw-ZgP" firstAttribute="trailing" secondItem="AnO-nA-r5S" secondAttribute="trailing" constant="20" symbolic="YES" id="Nel-Mr-3ig"/>
+                            <constraint firstAttribute="top" secondItem="sst-pw-ZgP" secondAttribute="top" constant="20" symbolic="YES" id="T5U-LL-uju"/>
+                            <constraint firstAttribute="leading" secondItem="sst-pw-ZgP" secondAttribute="leading" constant="20" symbolic="YES" id="awz-ex-Mus"/>
+                            <constraint firstItem="sst-pw-ZgP" firstAttribute="bottom" secondItem="AnO-nA-r5S" secondAttribute="bottom" constant="20" symbolic="YES" id="t4J-8C-rEW"/>
```
yo1995, line 29: It seems to have included some not needed settings.

image

Please clear all the constraints and set again.

arcgis-ios-sdk-samples/Display information/Show a popup/ShowPopup.storyboard
```diff
@@ -0,0 +1,42 @@
...
+        <!--Show Popup View Controller-->
+        <scene sceneID="RbY-vT-1T7">
+            <objects>
+                <viewController id="Tu7-xN-hEQ" customClass="ShowPopupViewController" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <view key="view" contentMode="scaleToFill" id="AnO-nA-r5S">
+                        <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <subviews>
+                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="sst-pw-ZgP" customClass="AGSMapView">
+                                <rect key="frame" x="-20" y="-20" width="454" height="936"/>
```
yo1995, line 20: Same here. Reset and set again would solve this.

arcgis-ios-sdk-samples/Display information/Show a popup/ShowPopupViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+
+class ShowPopupViewController: UIViewController, AGSGeoViewTouchDelegate, AGSPopupsViewControllerDelegate {
+    @IBOutlet private weak var mapView: AGSMapView! {
+        didSet {
+            mapView.map = makeMap()
+        }
+    }
+    
+    var featureLayer: AGSFeatureLayer?
+    var popupsViewController: AGSPopupsViewController?
```
yo1995, line 26: No need to keep a reference to the VC. It can be created on-the-fly.
```suggestion
```

arcgis-ios-sdk-samples/Display information/Show a popup/ShowPopupViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+            if let error = result.error {
+                self.presentAlert(error: error)
+            } else if !result.popups.isEmpty {
+                // Display a popup only if it exists.
+                self.popupsViewController = AGSPopupsViewController(popups: result.popups)
+                // Display the popup as a formsheet -- specified for iPads.
+                self.popupsViewController?.modalPresentationStyle = .formSheet
+                // Present the popup.
+                self.present(self.popupsViewController!, animated: true)
+                self.popupsViewController?.delegate = self
```
yo1995, line 61:
```suggestion
                // Display a popup only if it exists.
                let popupsViewController = AGSPopupsViewController(popups: result.popups)
                // Display the popup as a formsheet -- specified for iPads.
                popupsViewController.modalPresentationStyle = .formSheet
                // Present the popup.
                self.present(popupsViewController, animated: true)
                popupsViewController.delegate = self
```

Remove the reference to popupVC and create on-the-fly.

arcgis-ios-sdk-samples/Display information/Show a popup/ShowPopupViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+            }
+        }
+    }
+    
+    // MARK: - AGSPopupsViewControllerDelegate methods
+    func popupsViewControllerDidFinishViewingPopups(_ popupsViewController: AGSPopupsViewController) {
+        // Dismiss the popups view controller.
+        self.dismiss(animated: true)
+        // Reset the popups view controller
+        self.popupsViewController = nil
```
yo1995, line 71:
```suggestion
```

arcgis-ios-sdk-samples/Display information/Show a popup/ShowPopupViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+                self.featureLayer = map.operationalLayers[0] as? AGSFeatureLayer
+            }
+        }
+        return map
+    }
+    
+    // MARK: - AGSGeoViewTouchDelegate
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        // Identify the specified feature layer.
+        self.mapView.identifyLayer(featureLayer!, screenPoint: screenPoint, tolerance: 12, returnPopupsOnly: false) { [weak self] (result: AGSIdentifyLayerResult) in
```
yo1995, line 50: Force unwrap the featureLayer here seems a bit strange. I'd either create `var featureLayer: AGSFeatureLayer!` or guard let the feature layer if it is nil. But I doubt the second case would ever happen. 🤔 

Will leave it to Phil.

arcgis-ios-sdk-samples/Display information/Show a popup/ShowPopup.storyboard
yo1995: The map view.

arcgis-ios-sdk-samples/Display information/Show a popup/ShowPopupViewController.swift
yo1995: Also Phil would suggest remove the leading `self.` 😏

vquach2404:
@yo1995 @philium I made a slight enhancement. Added highlighting to indicate which feature was selected. Please give this another review!

arcgis-ios-sdk-samples/Display information/Show a popup/ShowPopupViewController.swift
```diff
@@ -51,6 +51,11 @@ class ShowPopupViewController: UIViewController, AGSGeoViewTouchDelegate, AGSPop
             if let error = result.error {
                 self.presentAlert(error: error)
             } else if !result.popups.isEmpty {
+                // Unselect the previous feature and select the newest one.
+                self.featureLayer?.clearSelection()
+                let features = result.geoElements as? [AGSFeature]
+                let selectedFeature = features?.first
+                self.featureLayer?.select(selectedFeature!)
```
yo1995, line 58: If it has to be unwrapped, maybe unwrap in the first place to avoid the overhead of 2 optionals chained together.

arcgis-ios-sdk-samples/Display information/Show a popup/ShowPopupViewController.swift
```diff
@@ -0,0 +1,82 @@
...
+    // MARK: - AGSGeoViewTouchDelegate
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        // Identify the specified feature layer.
+        mapView.identifyLayer(featureLayer!, screenPoint: screenPoint, tolerance: 12, returnPopupsOnly: false) { [weak self] (result: AGSIdentifyLayerResult) in
+            guard let self = self else { return }
+            if let error = result.error {
+                self.presentAlert(error: error)
+            } else if !result.popups.isEmpty {
+                // Unselect the previous feature and select the newest one.
+                self.featureLayer?.clearSelection()
```
philium, line 55: Rather than clearing the selection before selecting the next one, I would suggest clearing the selection when the popup is dismissed.

vquach2404: Hm, since the popup covers the whole or most of the screen, I think it may be better to leave it as is so that the user can identify which item was selected after the popup was displayed. What do you think?

philium: It seems to me that the primary benefit of a selection is for the user to know for which feature the popup is being shown. I don't see as much benefit in leaving the feature selected indefinitely after the popup is dismissed, but maybe that's just me.

yo1995: I see reasons in both ways, so probably stick with the original design to keep the consistency among platforms.

PR #907, opened 2020-07-06
Update: Surface placements

yo1995:
**Description**: This PR updates `Surface placements` sample. It adds a slider and a switch toggle for different surface placement modes, as well as refactor the code to adapt these changes.

Branch: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/Update-SurfacePlacement-vNext/arcgis-ios-sdk-samples/Scenes/Surface%20placements

Note: The warning below can be ignored. It is an Apple issue. ref: https://developer.apple.com/forums/thread/132035

```
ArcGIS Runtime SDK Samples[82431:3910227] invalid mode 'kCFRunLoopCommonModes' provided to CFRunLoopRunSpecific - break on _CFRunLoopError_RunCalledWithInvalidMode to debug. This message will only appear once per execution.
```

**Ref**: 

- /common-samples/issues/1846
- /common-samples/issues/2088

|`Demo`|
|:------:|
|demo|


arcgis-ios-sdk-samples/Scenes/Surface placements/README.md
```diff
@@ -1,15 +1,46 @@
 # Surface placements
 
-This sample demonstrates how to position graphics using different surface placements
+Position graphics relative to a surface using different surface placement modes.
```
vquach2404, line 3: Please update the description text in the contentPlist file as well!

arcgis-ios-sdk-samples/Scenes/Surface placements/README.md
```diff
@@ -1,15 +1,46 @@
...
+
+* AGSGraphic
+* AGSGraphicsOverlay
+* AGSLayerSceneProperties
+* AGSSurfacePlacement
+* AGSSurface
+
+## About the data
+
+The scene shows a view of Brest, France. Four points are shown hovering with positions defined by each of the different surface placement modes.
```
philium, line 36: There are five placement modes, but only four points. ~Am I missing something?~ I see. I can't help but wonder if there readme could be clearer about that, particularly because it says "Create an `AGSGraphicsOverlay` for _each placement mode_" (emphasis mine).

arcgis-ios-sdk-samples/Scenes/Surface placements/README.md
```diff
@@ -1,15 +1,46 @@
...
+
+## Additional information
+
+This sample uses an elevation service to add elevation/terrain to the scene. Graphics are positioned relative to that surface for the `drapedBillboarded` and `relative` surface placement modes.
+
+This sample uses an elevation service to add elevation/terrain to the scene. Graphics are positioned relative to that surface for the `drapedBillboarded`, `drapedFlat`, and `relative` surface placement modes. It also uses a scene layer containing 3D models of buildings. Graphics are positioned relative to that scene layer for the `relativeToScene` surface placement mode.
```
philium, line 42: This paragraph looks like it is meant to replace the previous one.

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacementsViewController.swift
```diff
@@ -17,65 +17,128 @@
 import UIKit
 import ArcGIS
 
-/// A view controller that manages the interface of the Surface Placements
-/// sample.
+/// A view controller that manages the interface of the Surface placements sample.
```
philium, line 20: The word "sample" was on a separate line so that the comment wrapped at the 80 column page guide:

image

Also, I used title case in these documentation comments because there's a word (sample) that appears after the title, making it really confusing as to where the title ends.

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacementsViewController.swift
```diff
@@ -17,65 +17,128 @@
...
+        formatter.numberStyle = .none
+        return formatter
+    }()
+    
+    // MARK: - Actions
+    
+    @IBAction func switchValueChanged(_ sender: UISwitch) {
+        // Toggle the visibility of two draped mode graphics overlays respectively.
+        graphicsOverlays.forEach { graphicOverlay in
+            switch graphicOverlay.sceneProperties!.surfacePlacement {
```
philium, line 62: Instead of looking for the draped graphic overlays each time, it would be cool to store them in a dictionary keyed by surface placement. Then updating their visibility would be much simpler:

```swift
overlaysBySurfacePlacement[.drapedFlat]?.isVisible = sender.isOn
overlaysBySurfacePlacement[.drapedBillboarded]?.isVisible = sender.isOn
```

Or better yet, just save off the draped overlays:

```swift
drapedOverlays.forEach { $0.isVisible = sender.isOn }
```

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacementsViewController.swift
```diff
@@ -17,65 +17,128 @@
...
+    
+    @IBAction func sliderValueChanged(_ sender: UISlider) {
+        let zValue = Double(sender.value)
+        zValueLabel.text = zValueFormatter.string(from: zValue as NSNumber)!
+        // Set the z-value of each geometry of surface placement graphics.
+        graphicsOverlays.forEach { graphicOverlay in
+            graphicOverlay.graphics.forEach { graphic in
+                let currentGraphic = graphic as! AGSGraphic
+                let originalPoint = currentGraphic.geometry as! AGSPoint
+                let newPoint = AGSPoint(
```
philium, line 81: Another way of doing this is with `AGSGeometryEngine`:

```swift
let newPoint = AGSGeometryEngine.geometry(bySettingZ: zValue, in: originalPoint) as! AGSPoint
```

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacementsViewController.swift
```diff
@@ -17,65 +17,128 @@
...
     }
+    
+    // MARK: UIViewController
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Add the source code button item to the right of navigation bar.
+        (self.navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["SurfacePlacementsViewController"]
+        // Initialize the draped mode visibility.
+        drapedModeSwitch.isOn = false
```
philium, line 139: Could this be set in the storyboard?

arcgis-ios-sdk-samples/Scenes/Surface placements/README.md
yo1995: 😵  Exactly... forgot to remove the previous version...

arcgis-ios-sdk-samples/Scenes/Surface placements/README.md
yo1995: I've made some changes, but from my point of view it is even better to remove the second sentence, to avoid confusion. 🤔

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacementsViewController.swift
yo1995: Ahh I see. How can I config Xcode to let it display the 3-slash comment as monospace font? 🤔 

image

---

Personally I'm all for a better coding style and format. While 80 column line wrap is ideal for a nice looking codebase, it occasionally drives me nut to find ways to cut short a line 🌀  (e.g. add newlines to function params, flatten condition/callback pyramid by extracting closures into functions, etc.). 

And since `SwiftLint` linter has a 40-line limit for a single function, when refactoring previous samples, I run into cases that trigger linter rule if I add a few long lines of code. And I need to find ways to bypass it, such as 

https://github.com/Esri/arcgis-runtime-samples-ios/blob/02231f0af8e174676968d17d96547a2e6bc67905/arcgis-ios-sdk-samples/Utility%20Network/Display%20utility%20associations/DisplayUtilityAssociationsViewController.swift#L151

So my thought is to loosen line break to 100-column width, considering the sometimes quite long API names in Swift.
@vquach2404 what's your thought? Or how do we wrap longer lines in ObjC codebase? 😳

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacementsViewController.swift
yo1995: It would be really great if SDK enums can conform to `CaseIterable`! 😄

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacementsViewController.swift
philium: To be clear, I only wrap documentation comments at the page guide so that they stay at a nice readable width. I haven't tried nor do I see any reason to start trying to wrap code at a page guide. That would just be nuts! 🥜

yo1995: I see!

yo1995:
After the Friday review, I decide to make a few small changes to the sample

- text label offset
- color for text label should be other than blue, not to blend with the sky
- change switch to segmented control

yo1995:
Ready for another review.

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacements.storyboard
```diff
@@ -1,35 +1,106 @@
...
+                            <toolbar opaque="NO" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="eWj-fZ-V3J">
+                                <rect key="frame" x="0.0" y="769" width="414" height="44"/>
+                                <items>
+                                    <barButtonItem style="plain" systemItem="flexibleSpace" id="PVB-DB-c7R"/>
+                                    <barButtonItem style="plain" id="4bj-UX-pB2">
+                                        <segmentedControl key="customView" opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="left" contentVerticalAlignment="top" segmentControlStyle="bar" selectedSegmentIndex="0" id="yTQ-cO-QvB">
+                                            <rect key="frame" x="20" y="5.5" width="374" height="33"/>
+                                            <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
+                                            <segments>
+                                                <segment title="Draped Billboarded"/>
```
philium, line 75: What you have is fine, but since both segments have the word "Draped", you might consider moving that outside the control into a label for the control.

yo1995: Yes this makes sense. When I changed it I wasn't sure whether to keep it outside segmented control, so I copied dotnet's approach: https://github.com/Esri/arcgis-runtime-samples-dotnet/pull/806/files?short_path=d1ee392#diff-d1ee392254bb7519df0497bfbb792d78

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacementsViewController.swift
```diff
@@ -20,62 +20,116 @@ import ArcGIS
...
+        }
+    }
+    
+    /// A dictionary for graphics overlays of different surface placement modes.
+    var overlaysBySurfacePlacement = [AGSSurfacePlacement: AGSGraphicsOverlay]()
+    /// A formatter to format z-value strings.
+    let zValueFormatter: MeasurementFormatter = {
+        let formatter = MeasurementFormatter()
+        formatter.unitStyle = .short
+        formatter.unitOptions = .providedUnit
```
philium, line 57: In other samples I've not specified this option so that the users sees the measurement in the unit that makes the most sense for their locale. It is up to you, but I would recommend doing that.

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacementsViewController.swift
```diff
@@ -20,62 +20,116 @@ import ArcGIS
...
-        
-        (self.navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["SurfacePlacementsViewController"]
-    }
-    
-    /// Creates a graphics overlay for the given surface placement.
+    /// Create a graphics overlay for the given surface placement.
     ///
-    /// - Parameter surfacePlacement: The surface placement for which to create
-    /// a graphics overlay.
+    /// - Parameters:
```
philium, line 106: I recommend using the format generated by Xcode, which is what was already there:

image

The rule seems to be `- Parameter` for a single parameter and `- Parameters:` for multiple parameters.

yo1995: Ha I see... The comment formats for single and multiple params are different! Will fix.

arcgis-ios-sdk-samples/Scenes/Surface placements/SurfacePlacementsViewController.swift
yo1995: Hmm I tend to leave it as is. If not specified, system locale would use miles as unit, which is not very readable (number always read around 0.1 miles). 😳 
There are a few other samples with it specified, such as Viewshed, LineOfSight, Animate 3D, etc.

philium: There is [`naturalScale`](https://developer.apple.com/documentation/foundation/measurementformatter/unitoptions/1642065-naturalscale), which would fix the problem of the measurement not being very readable. With that option, I'm pretty sure you'd get measurements in feet.

yo1995: Nice to earn a new thing everyday 😄

yo1995:
@saratchandrakarumuri Please take a look at this update. Thanks! 🙂

yo1995:
OK, let me give a try...

yo1995:
Aha it seems changing back to blue dismissed approvals...

yo1995:
Sorry my bad... always forget this! 😅

PR #909, opened 2020-07-08
[Dark Mode] Update: Utility network

yo1995:
This PR fixes the hard coded colors in `Utility network` category to make them semantic and Dark Mode adaptive.


vquach2404:
Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-04 at 10 48 55
Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-04 at 10 49 14
These two buttons are still blue but purple when pressed.

yo1995:
> These two buttons are still blue but purple when pressed.

This is a know issue, but I'm not sure about the cause. 🤔  If you `touchDown` on the buttons without `touchUp`, they will turn purple.
See comment here: https://github.com/Esri/arcgis-runtime-samples-ios/pull/924#issuecomment-708720094

We'll wait for Phil's thoughts.

vquach2404:
It seems like the toolbar for Configure subnetwork trace differs from the toolbars in the other samples. Is there a reason for this? 
Configure subnetwork trace:
Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-04 at 11 07 23
Perform valve isolation:
Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-04 at 11 07 19
I also assumed that we would keep the toolbar and background colors consistent with the main sample viewer. Using gray for the sample viewer but black for the samples seemed a bit odd to me. What are your thoughts on making the backgrounds and toolbars gray to match the sample viewer colors?

yo1995:
> It seems like the toolbar for Configure subnetwork trace differs from the toolbars in the other samples. (The color difference is more noticeable on an iPad device) Is there a reason for this?

Hmm I don't see a difference. I picked the colors from Simulator with the color picker tool in Photoshop and both of the colors are the same. 😳 

> I also assumed that we would keep the toolbar and background colors consistent with the main sample viewer. Using gray for the sample viewer but black for the samples seemed a bit odd to me. What are your thoughts on making the backgrounds and toolbars gray to match the sample viewer colors?

|black|gray|
|-|-|
|toolbar-black|toolbar-gray|

Both are the same color in Light mode. The only difference is in Dark Mode.
It is easy to change, but I'm a bit hesitant on this choice.

### Black

- Pros
1. follows Apple default color - no need to introduce the `tertiarySystemBackground` color
2. looks more consistent with the default table view colors
3. less changes to existing samples
- Cons
1. is not consistent with source code toolbars

### Gray

- Pros
1. consistent with source code toolbars
2. nicer to eyes being not solid black
- Cons
1. less in contrast
2. need to configure the root view's background color to `tertiarySystemBackground`. samples without toolbar may either need to do the same, or keep the default but make the root bgc inconsistent among samples.

I built and tested on a real device, and I prefer the darker black color. But I'm open to make the change, as that was the conclusion of our last discussion: https://github.com/Esri/arcgis-runtime-samples-ios/pull/899#issuecomment-719824730

Maybe we can open a vote somewhere in the Slack channels 😀 .

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTrace.storyboard
```diff
@@ -18,7 +19,7 @@
                         <subviews>
                             <tableView clipsSubviews="YES" contentMode="scaleToFill" alwaysBounceVertical="YES" dataMode="prototypes" style="grouped" separatorStyle="default" rowHeight="-1" estimatedRowHeight="-1" sectionHeaderHeight="18" sectionFooterHeight="18" translatesAutoresizingMaskIntoConstraints="NO" id="AON-wr-WJ5">
                                 <rect key="frame" x="0.0" y="0.0" width="414" height="813"/>
-                                <color key="backgroundColor" cocoaTouchSystemColor="groupTableViewBackgroundColor"/>
+                                <color key="backgroundColor" systemColor="systemGroupedBackgroundColor"/>
```
yo1995, line 22: This is a deprecated color in storyboard, and even though it doesn't give warnings, it is good to replace it. Read more at: https://developer.apple.com/documentation/uikit/uicolor/1623404-grouptableviewbackground

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTrace.storyboard
```diff
@@ -85,7 +86,8 @@
                                 </items>
                             </toolbar>
                         </subviews>
-                        <color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
+                        <viewLayoutGuide key="safeArea" id="Qfl-lK-vPg"/>
+                        <color key="backgroundColor" systemColor="tertiarySystemBackgroundColor"/>
```
yo1995, line 90: To keep consistent among all samples, the root view will use `tertiarySystemBackground` so that the toolbar are displayed as a lighter gray color in Dark Mode.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTrace.storyboard
```diff
@@ -52,7 +54,6 @@
                                                     <nil key="highlightedColor"/>
                                                 </label>
                                             </subviews>
-                                            <color key="backgroundColor" white="0.0" alpha="0.0" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
```
yo1995, line 55: No need to add clear color to a view explicitly, as the default color for a `UIView` is clearColor.

arcgis-ios-sdk-samples/Utility network/Trace utility network/TraceUtilityNetwork.storyboard
```diff
@@ -16,22 +18,28 @@
                         <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
+                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="pzi-IO-kfE" customClass="AGSMapView">
+                                <rect key="frame" x="0.0" y="44" width="414" height="769"/>
+                            </view>
```
yo1995, line 23: Adjusted the order of map view and statusLabel so that the label is on top of the map.

yo1995 (commented):
@vquach2404 Ready for another review, and I've changed the color for toolbar!

vquach2404:
> I also assumed that we would keep the toolbar and background colors consistent with the main sample viewer. 

The changes look good! I should've been clearer here. I was also talking about the table view background in Configure subnetwork trace. I think the making the table view background gray would be nice for consistency's sake. What do you think?

yo1995:
> > I also assumed that we would keep the toolbar and background colors consistent with the main sample viewer.
> 
> The changes look good! I should've been clearer here. I was also talking about the table view background in Configure subnetwork trace. I think the making the table view background gray would be nice for consistency's sake. What do you think?

See the review comments above. Apple does have recommended colors for tableview (both plain style and grouped style), and I tend to stick to the default in this case: https://developer.apple.com/documentation/uikit/uicolor/1623404-grouptableviewbackground

vquach2404:
> However, we may consider using `secondarySystemGroupedBackground` or `tertiarySystemGroupedBackground` instead, but we need to discuss on that.

Hm, we could leave it up for discussion. But I agree -- in this case, it seems best to stick to the default color.

vquach2404 (dismissed):

saratchandrakarumuri:
- In `Trace utility network`, Banner background is removed and it is now hard to read text in light mode

    IMG_0121


- Progress HUD is still not adopted, I guess it is expected for now since it is a 3rd party tool. Just want to confirm

- Black color table view in `Configure subnetwork trace` is not consistent, but doesn't look that odd with this sample so I am okay with that

yo1995:
> * In `Trace utility network`, Banner background is removed and it is now hard to read text in light mode

Interesting... Did you pull from the latest commit? This is what I see right now. (white text, dark background)
traceUN

> * Progress HUD is still not adopted, I guess it is expected for now since it is a 3rd party tool. Just want to confirm

Right. #983 and I'm still investigating what is the best way...

> * Black color table view in `Configure subnetwork trace` is not consistent, but doesn't look that odd with this sample so I am okay with that

😃  🙌

saratchandrakarumuri (approved):
@yo1995 With latest commit banner is now good on both iOS 13 and 14 versions

saratchandrakarumuri (approved):
Screenshot update is good

philium (approved):

PR #888, opened 2020-05-27
Update metadata

vquach2404:
@yo1995 and I have updated and edited all metadata files including alphabetizing, renaming, and formatting issues. Now it's time to merge this with the README updates! Looking forward to feedback!

PR #914, opened 2020-07-20
Hotfix: wrong category and duplicate entry on developer site

yo1995:
ref: /runtime/common-samples/issues/2006

These 2 issues have been fixed in #888 on `philium/readme-updates` branch. However, in today's mtg, Mike suggested that they should be fixed ASAP as a hotfix on master branch, to avoid keeping looking bad until we merge the whole update.

Therefore, I made this hotfix to mitigate the issues in metadata, so that they can be fixed on developer site during the weekly build.

### DIFF
--- a/arcgis-ios-sdk-samples/Route & Directions/Find closest facility to an incident (interactive)/README.metadata.json
+++ b/arcgis-ios-sdk-samples/Route & Directions/Find closest facility to an incident (interactive)/README.metadata.json
@@ -1,5 +1,5 @@
 {
-    "category": "Routing & Directions",
+    "category": "Routing and Directions",
     "description": "Find a route to the closest facility from a location.",
     "ignore": false,
     "images": [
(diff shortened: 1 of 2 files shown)


arcgis-ios-sdk-samples/Route & Directions/Find closest facility to an incident (interactive)/README.metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-    "category": "Routing & Directions",
+    "category": "Routing and Directions",
```
jf990, line 2: @yo1995 This category should be Route and Directions, to match the other samples in this category.

yo1995: Sorry for the miscue - a hotfix is at #917

PR #915, opened 2020-07-20
Update: app color scheme for U9

yo1995:
This PR updates both App icon and global tint color to purple, as well as created a Color assets catalog.

- Change the global tint color to `primaryThemeColor` in Main.storyboard
- Fix most hardcoded tint color in code
- Fix customized tint color in a few storyboards
- Override most common `UIControl`s' tint color in AppDelegate
- Devise a set of semantic named colors as `Assets.xcassets`

---

There are a few issues yet to be fixed:

- The `SFSafariViewController` for authentication - tint is blue 
Auth VC

- ~~After changing some hardcoded colors into named colors, wrong storyboard designs emerge~~
[wrong-toolbar image](https://user-images.githubusercontent.com/9660181/88833034-88d98100-d186-11ea-89eb-a0d9b161f745.png)

Edit: this does not apply anymore. Will be fixed in Dark Mode adaption

- The white tint color for category icon in content collection view is not in grayscale - will be fixed as part of Dark Mode adaption.


Colors.xcassets/EsriBlue.colorset/Contents.json
```diff
@@ -0,0 +1,20 @@
...
+  "colors" : [
+    {
+      "color" : {
+        "color-space" : "srgb",
+        "components" : {
+          "alpha" : "1.000",
+          "blue" : "0.763",
+          "green" : "0.476",
+          "red" : "0.000"
+        }
```
yo1995, line 11: The colors are picked from Esri Brand Guidelines https://webapps-cdn.esri.com/brand/guidelines.html . For Dark Mode counterparts, I made small adjustments based on the original color.

arcgis-ios-sdk-samples/Content Display Logic/Views/DownloadProgressView.xib
```diff
@@ -33,14 +33,15 @@
...
+                            <rect key="frame" x="8" y="101" width="224" height="41"/>
                             <fontDescription key="fontDescription" type="system" pointSize="17"/>
                             <nil key="textColor"/>
                             <nil key="highlightedColor"/>
                         </label>
                         <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="roundedRect" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="eOX-4U-fqJ">
-                            <rect key="frame" x="91.5" y="132" width="57" height="33"/>
+                            <rect key="frame" x="91.5" y="152" width="57" height="33"/>
                             <fontDescription key="fontDescription" type="boldSystem" pointSize="17"/>
+                            <color key="tintColor" name="primaryThemeColor"/>
```
philium, line 44: Could this be set on the xib's root view so that it would apply to any control in the view, not just this one?

arcgis-ios-sdk-samples/Maps/Download preplanned map/DownloadPreplannedMapViewController.swift
```diff
@@ -20,7 +20,7 @@ import ArcGIS
 // MARK: - Constants
 
 private enum Constants {
-    static let activityIndicatorColor = UIColor.primaryBlue
+    static let activityIndicatorColor = UIColor.primaryThemeColor
```
philium, line 23: If we want the color of activity indicator's to match the global accent color, then we should do it through the appearance proxy, not in individual samples like this. That being said, I'm not sure that's what we want. The accent color is usually an indication of an interactive element, which an activity indicator is not. Because of that, I would suggest just removing this so that the default color is used.

arcgis-ios-sdk-samples/Maps/Download preplanned map/PreplannedMapAreaTableViewCell.swift
```diff
@@ -19,7 +19,7 @@ import UIKit
 // MARK: - Constants
 
 private enum Constants {
-    static let progressTintColor = UIColor.primaryBlue
+    static let progressTintColor = UIColor.primaryThemeColor
```
philium, line 22: This shouldn't be needed anymore since the `tintColor` of `UIProgressView`'s appearance proxy is now set in the app delegate.

arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/GenerateOfflineMapOverrides.storyboard
```diff
@@ -52,11 +51,11 @@
                                     </label>
                                     <progressView opaque="NO" contentMode="scaleToFill" verticalHuggingPriority="750" progress="0.5" translatesAutoresizingMaskIntoConstraints="NO" id="GNP-kU-lcZ">
                                         <rect key="frame" x="8" y="33.5" width="303" height="2"/>
-                                        <color key="tintColor" red="0.0" green="0.36862745099999999" blue="0.58431372550000005" alpha="1" colorSpace="calibratedRGB"/>
+                                        <color key="tintColor" name="primaryThemeColor"/>
```
philium, line 54: Could you please set the global tint color of the storyboard rather than the tint color on individual views?

yo1995:
> * Could you please address this issue?
>   image
>   Just select the misplaced view and use Editor > Update Frames.

Sure. I didn't fix it because I was afraid to introduce > 20 autogenerated unrelated changes to `main.storyboard`. Every time we touch a xib, Xcode want to make it up-to-date... 🤥 

> * Why are there two different app icons (`AppIcon` and `AppIcon2`)?

AppIcon2 is used in the app version info page. Do you think it's OK to remove it? It might need to be there because AppIcon has different dimension requirements to a normal image asset.

arcgis-ios-sdk-samples/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -86,7 +86,7 @@ class DistanceMeasurementAnalysisViewController: UIViewController, AGSGeoViewTou
         sceneView.analysisOverlays.add(analysisOverlay)
         
         distanceView.backgroundColor = .backgroundGray
-        distanceView.tintColor = .primaryBlue
+        distanceView.tintColor = .primaryThemeColor
```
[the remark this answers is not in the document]
yo1995: This is the question I raised last week in our group chat with Mike. 

My observation is that:

- The `main.storyboard` file has a global tint color, which affects all the content display logic views in it. But this global tint is not inherited by any sample's separate storyboard.
image
- Our current setup makes every sample have a separate storyboard. If we do it by setting everything in storyboard, we'll need change every storyboard's (>180 samples 😓 ) global tint to our custom named theme color. I don't know if Xcode 12 can help with overriding them in one place. I'll read more about accent color to see if it helps.
- If we keep our current approach, i.e. 

    - overriding some `UIControl`s tint color in AppDelegate, e.g. tool bar, progress view, bar button, slider, etc.
    - set global tint for `main.storyboard`

  then we'll need to remove all other hardcoded tint colors like this from code, and make sure they get overriden. Still, this would require a few changes in some storyboards.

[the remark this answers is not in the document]
philium: Not every sample has a view that needs to be tinted. And for many that do, that is taken care of with the appearance changes made in the app delegate. For the remaining samples, setting the global tint color of the storyboard is a better solution than manually setting the tint color of certain views since it applies to every view in the sample. I'm not suggesting updating the global tint color of every storyboard. But for cases like this where we're setting it in code, we should switch to setting it in the storyboard itself. Does that make sense?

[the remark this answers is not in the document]
yo1995: Yes. This sounds like a more doable way to manage these colors. I'll make these changes later.

Colors.xcassets/Contents.json
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
yo1995: Will first be fixed in #927 .

Colors.xcassets/primaryThemeColor.colorset/Contents.json
```diff
@@ -0,0 +1,38 @@
+{
```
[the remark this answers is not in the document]
yo1995: Is it safe to make this change to project file in Xcode 12 Beta, while still using Xcode 11 for development? 😳 

I'm not familiar with the Xcode beta version compatibility. Thanks!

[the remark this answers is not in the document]
philium: Yes, it is safe. The unrecognized setting will just be ignored (it appears at the bottom of the settings list).

arcgis-ios-sdk-samples/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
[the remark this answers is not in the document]
yo1995: Actually, after a closer examine to this particular hardcoded tint color, I find it is an invalid setting, since the `distanceView` itself is a stack view, so no color would ever take effect. I'll remove it altogether.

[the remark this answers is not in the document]
philium: Remember that `tintColor` is inherited from ancestors, so setting it on the stack view affects all subviews of it. That includes the segmented control. Since segmented controls are tinted on iOS 12 and we still support iOS 12, removing it altogether would make a difference.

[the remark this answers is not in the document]
yo1995: > Remember that `tintColor` is inherited from ancestors, so setting it on the stack view affects all subviews of it. That includes the segmented control. Since segmented controls are tinted on iOS 12 and we still support iOS 12, removing it altogether would make a difference.

OK I see. Then I'll still remove this view's hardcoded color, while set the global tint of its storyboard to accent color.

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -169,9 +172,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
...
 
 extension UIColor {
-    class var primaryBlue: UIColor { return #colorLiteral(red: 0, green: 0.475, blue: 0.757, alpha: 1) }
-    class var secondaryBlue: UIColor { return #colorLiteral(red: 0, green: 0.368, blue: 0.584, alpha: 1) }
-    class var backgroundGray: UIColor { return #colorLiteral(red: 0.973, green: 0.973, blue: 0.973, alpha: 1) }
-    class var primaryTextColor: UIColor { return #colorLiteral(red: 0.196, green: 0.196, blue: 0.196, alpha: 1) }
-    class var secondaryTextColor: UIColor { return #colorLiteral(red: 0.349, green: 0.349, blue: 0.349, alpha: 1) }
+    // Also used for global tint color.
+    class var primaryThemeColor: UIColor { return UIColor(named: "primaryThemeColor")! }
+    class var secondaryThemeColor: UIColor { return UIColor(named: "secondaryThemeColor")! }
```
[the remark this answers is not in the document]
yo1995: I will remove not used colors from code.

However, I'd like to keep them in asset catalog. They are copied from Esri's branding guide, and I think it would be good to have a few "safe" colors to use in case we need them in the future, rather than using arbitrary custom or system colors. 😳 

If need to remove, I'll still keep 1 translucent background gray for future dark mode adaption, even if it is not in use right now.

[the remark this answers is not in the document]
philium: It is generally a best practice to not include unused code or assets in a repo as it unnecessarily as it can have negative consequences, such as increasing build time and the size of the app. While leaving these colors in would have a negligible impact on build time and size, there are other consequences. The colors in the branding guide could change between now and when the unused colors may be used, potentially leading to using the wrong color. And sometimes plans change, meaning that the colors could end up not being used at all.

It is easy enough to add a color asset to the asset catalog. I would strongly suggest including only the color assets that are currently being used and adding additional colors only when they become needed.

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -106,23 +106,26 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
...
             navigationBarAppearanceProxy.titleTextAttributes = [.foregroundColor: UIColor.white]
-            navigationBarAppearanceProxy.barTintColor = .primaryBlue
+            navigationBarAppearanceProxy.barTintColor = .primaryThemeColor
         }
         navigationBarAppearanceProxy.tintColor = .white
-        
+        // Override tint color for `UIControl`s.
         UIToolbar.appearance().barTintColor = .backgroundGray
-        UIToolbar.appearance().tintColor = .primaryBlue
+        UIToolbar.appearance().tintColor = .primaryThemeColor
```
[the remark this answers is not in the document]
yo1995: I've made some change to this! Do you think `UISwitch` should be tinted? 
I see reasons for both
- yes: it is a control, user interactive, keep the same behavior as prior versions
- no: intuitive default green/gray color for on/off, Apple don't tint it with accent color.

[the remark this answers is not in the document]
philium: Actually, Apple does tint switches in some cases. I found an example in Pages. So I would say go ahead and keep it tinted.

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -169,9 +177,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
...
 
 extension UIColor {
-    class var primaryBlue: UIColor { return #colorLiteral(red: 0, green: 0.475, blue: 0.757, alpha: 1) }
-    class var secondaryBlue: UIColor { return #colorLiteral(red: 0, green: 0.368, blue: 0.584, alpha: 1) }
-    class var backgroundGray: UIColor { return #colorLiteral(red: 0.973, green: 0.973, blue: 0.973, alpha: 1) }
-    class var primaryTextColor: UIColor { return #colorLiteral(red: 0.196, green: 0.196, blue: 0.196, alpha: 1) }
-    class var secondaryTextColor: UIColor { return #colorLiteral(red: 0.349, green: 0.349, blue: 0.349, alpha: 1) }
+    // Also used as global tint/accent color.
+    class var accentColor: UIColor { return UIColor(named: "AccentColor")! }
+    class var backgroundGray: UIColor { return UIColor(named: "backgroundGray")! }
```
philium, line 182: A couple comments about this color:

- It is defined using a system color that was introduced in iOS 13. What is the behavior on iOS 12? I'm a bit surprised there's not a warning about that.
- It is only used in two places. Could we changes those places to use a system color and then get rid of this color?

yo1995:
> It looks like there are some app icon sizes missing. Do we not have those or do they not show up in Xcode 11?
> 
> image

They are absent in current version on `v.next`, so I didn't fill them. Do you want them filled?

philium:
> Do you want them filled?

Yes please.

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -169,9 +177,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
...
+    class var backgroundGrayWithOpacity: UIColor { return UIColor(named: "backgroundGrayWithOpacity")! }
```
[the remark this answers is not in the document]
yo1995: Yes. I'll add this color with another PR later. 
This color will be used in future Dark Mode adaption to replace the custom translucent background colors for status labels.

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -169,9 +177,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
...
+    class var primaryTextColor: UIColor { return UIColor(named: "primaryTextColor")! }
```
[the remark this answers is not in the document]
yo1995: Yes. Since it is a black color in light mode, and doesn't make sense to change under dark mode as it is part of the map graphics, I change it to `UIColor.black` - `#000000`.

arcgis-ios-sdk-samples/AppDelegate.swift
yo1995: > * It is defined using a system color that was introduced in iOS 13. What is the behavior on iOS 12? I'm a bit surprised there's not a warning about that.

My guess is it uses a similar strategy as storyboard. When a system color is assigned, storyboard will provide a fallback color for iOS versions prior to 13 (e.g. white color is backgroundColor in light mode, in iOS 12).
```xml
<color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
```
Though I cannot tell from its content json...

> * It is only used in two places. Could we changes those places to use a system color and then get rid of this color?

Yes. Let me figure it out...

yo1995: It can be replaced by either `secondarySystemBackgroundColor` or `systemGray6`, but both colors are iOS 13+ only. Since we want to make this color Dark Mode adaptive, we can either keep this custom named color in asset catalog, or add it via code with `if #available iOS 13`. I tend to keep it in the assets. What do you think?

yo1995:
> * The launch screen still has the old color.

Will fix it. Also, launchScreen on iOS 13 is not displaying the nav bar, and that will be fixed in another PR later.

> * On iOS 12, the color of buttons in the navigation bar is not correct.

Setting the global tint in storyboard override nav bar's tint in iOS 12. I'm trying to figure out how to solve it.

> * Table view rows that act as buttons are not tinted property. Here's an example from the Map Reference Scale sample:

Nice catch. I only run a script to check all customized tint color in storyboard and code, but didn't check these which are not set to the tintColor property. For this particular case for example, it set `cell.textLabel?.textColor = view.tintColor`, which is both a bad practice and hard to find. I'll try with multiple methods to snatch these issues.

> I only looked at a handful of samples. Have you looked through all of the samples on each supported OS?

No I didn't. As I said above, I used shell script to find those custom tint colors, but apparently there are more than that... 😓 

Additionally I used these 2 keyword to search and fixed a few more issues.
- `tintColor`
- `textColor`

arcgis-ios-sdk-samples/Maps/Take screenshot/MapViewScreenshot.storyboard
```diff
@@ -1,41 +1,39 @@
...
                                     </imageView>
                                     <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="X" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="cmO-fw-Kfb">
-                                        <rect key="frame" x="341" y="20" width="14" height="24"/>
+                                        <rect key="frame" x="380" y="20" width="14" height="24"/>
                                         <constraints>
                                             <constraint firstAttribute="width" constant="14" id="Cze-aC-aK3"/>
                                             <constraint firstAttribute="height" constant="24" id="M9S-QK-EWl"/>
                                         </constraints>
                                         <fontDescription key="fontDescription" type="system" weight="heavy" pointSize="20"/>
-                                        <color key="textColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
```
yo1995, line 38: The white "X" at upper left corner should be white color.

arcgis-ios-sdk-samples/Maps/Open map (URL)/OpenMapURL.storyboard
```diff
@@ -77,29 +75,26 @@
...
                                     <autoresizingMask key="autoresizingMask"/>
                                     <subviews>
                                         <label opaque="NO" multipleTouchEnabled="YES" contentMode="left" text="Title" lineBreakMode="tailTruncation" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" id="RqL-Bf-Ac8">
-                                            <rect key="frame" x="20" y="0.0" width="374" height="43.666666666666664"/>
+                                            <rect key="frame" x="20" y="0.0" width="374" height="43.5"/>
                                             <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
-                                            <color key="backgroundColor" red="0.0" green="0.0" blue="0.0" alpha="0.0" colorSpace="custom" customColorSpace="sRGB"/>
                                             <fontDescription key="fontDescription" type="system" pointSize="16"/>
-                                            <color key="textColor" red="0.0" green="0.0" blue="0.0" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
```
yo1995, line 96: I fixed a few wrong custom colors in this storyboard, but they are not really related to this PR. Should I revert them?

arcgis-ios-sdk-samples/Layers/Display subtype feature layer/DisplaySubtypeFeatureLayer.storyboard
```diff
@@ -159,7 +160,7 @@
                                                     <rect key="frame" x="20" y="0.0" width="374" height="43.5"/>
                                                     <autoresizingMask key="autoresizingMask"/>
                                                     <fontDescription key="fontDescription" type="system" pointSize="17"/>
-                                                    <color key="textColor" systemColor="systemBlueColor" red="0.0" green="0.47843137250000001" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
```
yo1995, line 162: This is a cell-as-button cell with blue color.

philium:
@yo1995 In the Map Reference Scale sample, the Done button's text color is now incorrect. On iOS 12 and iOS 13, it is the default system tint color:

image

On iOS 14, the Done button's text color is the accent color, so it blends in with the navigation bar:

image

arcgis-ios-sdk-samples/Maps/Open map (URL)/OpenMapURL.storyboard
philium: No, leaving those changes in this PR is fine.

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -106,23 +106,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
...
+            // Nothing to do! iOS 14 handles global tint with accent color.
+        } else {
+            // Override tint color for `UIControl`s.
+            UIToolbar.appearance().tintColor = .accentColor
+            UISlider.appearance().tintColor = .accentColor
+            UITableViewCell.appearance().tintColor = .accentColor
+            UIProgressView.appearance().tintColor = .accentColor
+            UIView.appearance(whenContainedInInstancesOf: [UIAlertController.self]).tintColor = .accentColor
         }
-        navigationBarAppearanceProxy.tintColor = .white
```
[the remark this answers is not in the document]
yo1995: Actually this is still needed to override sample's nav tint.

yo1995:
> In the Map Reference Scale sample, the Done button's text color is now incorrect. 

I see. I'll revert the change to nav bar's tint in AppDelegate.

yo1995:
One more thing to examine: see if there is any "link color" in storyboard and code, and see if it is accent color.

yo1995:
Hi Phil @philium , this is ready for review again.

Should we roll out another TestFlight version with the latest updates in `v.next` and this PR, and ask our Cocoa team members to test, so that we can spot potential issues?

philium:
@yo1995 Thanks! I'll try to get to this later today. Once it is merged, it would be good to put up another TestFlight build.

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -19,6 +19,7 @@
 		00686FD02480889900EDA705 /* NavigateARNavigatorViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 00686FCF2480889900EDA705 /* NavigateARNavigatorViewController.swift */; };
 		00686FD52481CF4D00EDA705 /* NavigateARRoutePlannerViewController.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = 00C024E42475E4AB00E1DA8D /* NavigateARRoutePlannerViewController.swift */; };
 		00686FD62481CF4D00EDA705 /* NavigateARNavigatorViewController.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = 00686FCF2480889900EDA705 /* NavigateARNavigatorViewController.swift */; };
+		009D8BEF24F9AC8D00FD7E76 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 009D8BEE24F9AC8D00FD7E76 /* LaunchScreen.storyboard */; };
```
philium, line 22: It looks like Xcode doesn't like where this item is. Please open the project and then do something that would cause Xcode to re-save the project (such as rearranging any item) and then undo. Xcode will re-save the project with this item in the proper place.

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/LaunchScreen.storyboard
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
philium, line 1: Thank you for converting the launch screen nib into a launch storyboard. With a storyboard, we can more closely model the main view of the app on app launch. I spent quite a bit of time playing with the colors to get them right, but there just doesn't seem to be a way. Very frustrating. In any case, please make the following changes:

- Mark the storyboard as a launch storyboard.
  Screen Shot 2020-09-02 at 7 12 48 AM
- Use a split view controller. I know we can't get the colors right, but we can at least get the other aspects of the launch image correct. The goal is for the launch image to resemble the initial view of the app as closely as possible.

yo1995:
Spotted another issue: on iOS 12 iPad, the detail split view has wrong bar tint. Might be due to global storyboard tint.

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/LaunchScreen.storyboard
yo1995: The 2 issues you mentioned are fixed. Now the LaunchScreen uses a `UISplitViewController`. There are a few comments and issues:

demo

- On iOS 12, the LaunchScreen is free of issue 🎉  for
    - Landscape: iPad, iPhone
    - Portrait: iPhone
- For iOS 12 iPad Portrait mode, only the detail VC is displayed in LaunchScreen. A hack to solve it is [here](https://stackoverflow.com/questions/38564870/uisplitviewcontroller-allvisible-in-launchscreen-storyboard).
- On iOS 13, due to the new "prefer large title" attribute for navigation bar, a blank screen will show for both iPhone or iPad. 
    - I tried several Apple native apps with navigation bar on their first screen (Apple Research, TestFlight, Health) and none of them show a "placeholder frame" of the nav bar with large title. So I assume showing a blank screen is actually the expected behavior for iOS 13 and above.
- On iOS 13 and iOS 14 the searchBar is displayed, while on iOS 12 it is hidden. It is probably due to [this](https://stackoverflow.com/questions/57581557/how-to-initally-hide-searchbar-in-navigation-controller-on-ios-13). I wonder if we should make it consistent as iOS 12's behavior on iOS 13 and above?

philium: Thanks for doing all that research! What you have is fine.

philium (approved):

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/LaunchScreen.storyboard
yo1995: Thanks. I'll add Sarat to review this and also roll out another TestFlight version. After that I'll post in cocoa channel to let  our team give it a try.

yo1995:
@saratchandrakarumuri This is an update to app-wide colors, and it might break some tests. Please take some time to test if it break something.

Major changes are
- LaunchScreen color
- Main collection view nav bar color
- tint colors for UIControls, e.g. bar button, slider, switches, toolbar components
- AppIcon
- Misc color adaptions in samples

saratchandrakarumuri (commented):
I have did a quick sweep through all the samples and found following

Opening  `Change viewpoint` after `Display layer view state` is taking lot of time to render map. This is pretty inconsistent but ran into it a couple of times

Remnants of old color scheme in some places

- Old icon displayed in recents apps screen on phone device. Works well on iPad
IMG_C34A748CFB0B-1

- `Query` button in `Query a map image layer` sample
IMG_CFBE00F1C5E2-1

- `Zoom` button in `Feature layer rendering mode (map)` sample
IMG_58FBE5B567A8-1

- `Cancel` button in `Feature layer query`
- Accessory buttons on feature callouts in edit data samples
IMG_E9187FDC46C2-1
IMG_0097

- Directions deletion button in `Route around barriers` , `Offline routing`
IMG_250343474AC5-1

- Banner in `Viewshed (geoprocessing)`
IMG_B342A8159192-1

yo1995:
@saratchandrakarumuri Thanks for your efforts. I'll try to fix them tomorrow. 
I took a quick look and most of these colors escaped my last search with various reasons. 😩  
I'll fix and provide more details.

yo1995:
@saratchandrakarumuri 

- > Opening  `Change viewpoint` after `Display layer view state` is taking lot of time to render map. This is pretty inconsistent but ran into it a couple of times

I cannot reproduce this. Double checked and the `Display layer view state` sample does not leak memory.

- > Old icon displayed in recents apps screen on phone device. Works well on iPad

Could you try to delete the app from Simulator and build again? I double checked the assets and there are no old icons remaining, so I guess it's a caching problem.

icons

- > `Query` button in `Query a map image layer` sample

Fixed. Sorry for missing that. I double checked all `UIButton`s in the code and logged a few issues. No other buttons have wrong tint.

- > `Zoom` button in `Feature layer rendering mode (map)` sample

I also checked all buttons in storyboard with

```sh
find . \( -name '*.xib' -or -name '*.storyboard' \) \
    -exec echo {} \;        \
    -exec xmlstarlet sel -t \
            -m "//button" -c . -n  {} \; > ~/Desktop/buttons.md
```

and fortunately, `Feature layer rendering mode (map)` is the only one that has a custom color that has to be changed. (although there are a few other buttons with custom colors, which we should fix in Dark Mode process.)

- > `Cancel` button in `Feature layer query`

This one was really tricky! There is no explicit declaration of a button in this sample. Instead, the button comes along with the search bar. Fixed by overriding the global tint color of this sample. And again, fortunately, this is the only sample that uses a search bar with cancel button! 😥 

- > Accessory buttons on feature callouts in edit data samples

@philium I'm not pretty sure about this. I think the UI for an `AGSCallout` is coming from the SDK. Is it possible/Should we override its color in the sample viewer? Also @dg0yal as this could be another color adaption we can make for both Dark Mode and global accent tint.

- > Directions deletion button in `Route around barriers` , `Offline routing`

These 2 are an image of a blue trash can... @philium what's your thought? Should I replace the icon with a black trash can icon and tint it with accent color? Or probably we just leave it as-is.

- > Banner in `Viewshed (geoprocessing)`

This one is intended to left unchanged (at least for this PR, might be fixed in Dark Mode later). There are a few similar samples that uses the strange blue color, such as `Line of sight (location)`

Since the status label/banner is not interact-able, it is fine not to use the purple tint.

saratchandrakarumuri:
> ❓ Unchanged. Could you try to delete the app from Simulator and build again? I double checked the assets and there are no old icons remaining, so I guess it's a caching problem.  

Weirdly restarting the device after reinstalling worked 🤷‍♂️ 

@yo1995 Rest of the fixes are also looking good... I will wait for Phil's comments on the other points you tagged him on

philium:
I've heard that simulator builds are frakenbuilds and I've definitely seen changes not show up until the app was deleted and reinstalled.

philium:
@yo1995 Regarding this:

image

Just mark the image as template image in the asset catalog and that should fix the problem.

philium:
@yo1995 Regarding this:

image

Could the text color be changed to the default? I think that's what we use for most of these status bar views. It doesn't really make sense for them to be tinted purple, blue, or any other color.

yo1995:
I see. Let me fix those.

saratchandrakarumuri:
@yo1995 One important piece I missed last time is, What is plan on updating images for all the sample readmes? Is it tracked in a different issue/PR?

yo1995:
> One important piece I missed last time is, What is plan on updating images for all the sample readmes? Is it tracked in a different issue/PR?

According to my understanding, after this PR get merged and before it's uploaded to AppStore, we'll update the 2 images on developer site and in this repo's README. See more in `/common-samples/issues/2159`. Mike will also help with this.

For other images for each sample, since there are already some screenshots that are outdated, we plan to update them along with the Dark Mode PRs, where we'll 

- Check each sample for Dark Mode colors
- Remove/Change wrong usage of UI designs: constraints, UIButton, custom colors, etc.
- Recapture the screenshot with iPhone 11 dimensions
- Unify styling: status label appearance, button and switches, etc.

And each category will be updated in a separate PR, so that it is easier to review.

yo1995 (commented):
@philium I've applied your suggestions. Please take another look. Thanks! 😸

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -20,6 +20,7 @@
 		00686FD52481CF4D00EDA705 /* NavigateARRoutePlannerViewController.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = 00C024E42475E4AB00E1DA8D /* NavigateARRoutePlannerViewController.swift */; };
 		00686FD62481CF4D00EDA705 /* NavigateARNavigatorViewController.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = 00686FCF2480889900EDA705 /* NavigateARNavigatorViewController.swift */; };
 		008D175224EEEEBD0001BB8F /* loudoun_anno.geodatabase in Resources */ = {isa = PBXBuildFile; fileRef = 008D175124EEEEBD0001BB8F /* loudoun_anno.geodatabase */; settings = {ASSET_TAGS = (loudoun_anno, ); }; };
+		009D8BEF24F9AC8D00FD7E76 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 009D8BEE24F9AC8D00FD7E76 /* LaunchScreen.storyboard */; };
```
philium, line 23: This is not sorted in the project file correctly. Please have Xcode re-save the project so that it gets moved to the right spot.

philium (changes requested):

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
yo1995: Fixed.

philium (dismissed):

saratchandrakarumuri:
> Accessory buttons on feature callouts in edit data samples

@philium @yo1995 Any updates on what to with this?

philium:
@saratchandrakarumuri Which samples still have an issue with that? I am seeing the proper tint color used in some Edit Data samples, such as Update Attributes (Feature Service).

yo1995:
@philium No I didn't change the color for it, and it still remains blue.

|||
|-|-|
|callout-accessory-button|callout-accessory-button-2|

philium:
@yo1995 Which version of iOS are you using? I'm seeing the purple color.

image

yo1995:
> Which version of iOS are you using? I'm seeing the purple color.

I'm still seeing blue for iOS 12 and iOS 13. 
In iOS 14 it might be overriden by the Accent Color setting in info.plist...? 🤔

philium:
@yo1995 Setting the global tint color in the storyboard for those samples should fix the problem on both iOS 12 and iOS 13.

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -106,23 +106,29 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
...
+        if #available(iOS 14.0, *) {
+            // Nothing to do! iOS 14 handles global tint with accent color.
+        } else {
+            // Override tint color for `UIControl`s.
+            UIToolbar.appearance().tintColor = .accentColor
+            UISlider.appearance().tintColor = .accentColor
+            UITableViewCell.appearance().tintColor = .accentColor
+            UIProgressView.appearance().tintColor = .accentColor
+            UIView.appearance(whenContainedInInstancesOf: [UIAlertController.self]).tintColor = .accentColor
+            UIButton.appearance(whenContainedInInstancesOf: [AGSCallout.self]).tintColor = .accentColor
```
yo1995, line 130: @philium Instead of changing each storyboard which mapView that has a callout, I decide to override it in `AppDelegate`, as it involves less changes. What do you think?

philium: That works, although it probably doesn't need to be restricted to just `AGSCallout`. Either way is fine; we could always change it later if needed.

philium (approved):

arcgis-ios-sdk-samples/AppDelegate.swift
yo1995: @saratchandrakarumuri All issues you [mentioned](https://github.com/Esri/arcgis-runtime-samples-ios/pull/915#issuecomment-691335967) are resolved now! Feel free to take another look.

saratchandrakarumuri (approved):
@yo1995 I already verified other changes and the callout color changes are also looking good

yo1995:
Thanks your helpful feedback! 🎉 

Now the color update is merged, we'll head toward Dark Mode adaption and look forward to get it done before U10 release.

PR #917, opened 2020-07-23
Hotfix: wrong category in metadata

yo1995:
An issue causing the site not to render correctly. See image below

image


arcgis-ios-sdk-samples/Route & Directions/Find closest facility to an incident (interactive)/README.metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-    "category": "Routing and Directions",
+    "category": "Route and Directions",
```
maryharvey, line 2: Route and directions (lower case d)

maryharvey: Agreed to leave as Route and Directions. The casing will be addresses as a cross-platform issue.

PR #919, opened 2020-07-27
Fix: category casing in metadata and project file

yo1995:
This PR fixes the issue where category casing is not consistent on developer site. This is due to the case in their respective metadata files are not title case.

The PR changes several places:
1. In project file: 
    - change `Utility Network` to `Utility network`
    - change `Route & Directions` to `Route and directions`
    - change `Cloud & Portal` to `Cloud and portal`
2. In ContentPlist.plist
    - change all category names to title case to align with Apple's standard.
3. In 3 categories, fix all samples' metadata category: 
    - Cloud and portal
    - Route and directions
    - Utility network

Inconsistent casing


arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -2179,7 +2179,7 @@
 				3E0B58451E36BCCD00951767 /* Analysis */,
 				3EEE1D671E4935FC003263FC /* Cloud & Portal */,
 				3EEA06001D21C3E800E03774 /* Scenes */,
-				D98CC08F22E25D8200618682 /* Utility Network */,
+				D98CC08F22E25D8200618682 /* Utility network */,
```
maryharvey, line 2182: I am wondering if this change is necessary. You are updating this line but not updating Route & Directions and Cloud & Portal. Perhaps it is not necessary for the TOC listing?

yo1995: This change does not have effect on dev site TOC, but should be done as part of casing update.

In terms of `Route & Directions` and `Cloud & Portal` in project file, I didn't receive comments on whether we should make them consistent or keep them as-is, so I didn't change them. I tend to keep category names consistent everywhere, but I don't know who can decide on if we should make this move. Maybe wait for Phil's feedback? 🤔

philium:
This seems to be outside of the scope of the README update issue and the `readme-updates` branch is already very long-lived. If we're happy with the way that branch is, I say we go ahead and merge it and then tackle the category name changes as part of a separate issue. What do you think?

yo1995:
> This seems to be outside of the scope of the README update issue and the `readme-updates` branch is already very long-lived. If we're happy with the way that branch is, I say we go ahead and merge it and then tackle the category name changes as part of a separate issue. What do you think?

Sounds good to me. 

One reason I targeted this PR to `philium/readme-updates` is that these category renaming in metadata would have effect on dev site, and Mary and John indicated these categories should be fixed sooner than later.

I'll temporarily put this PR on hold and wait for Mary's feedback next week. Meanwhile, since the `philium/readme-updates` branch should already be free of issues, I'll render another preview with it and do a final check, before we merge #869 . 

If it is good, can we merge #869 next week? 😳 

---

@maryharvey do you think it is OK to push back this fix a little bit? I assume it would be merged by U9 release, but in case it doesn't, is it acceptable to leave those category names as-is a little longer?

yo1995:
As #869 is successfully merged 🎉 , we can start to review these changes and decide if they are needed.

vquach2404:
I think it'd be good to go ahead with these changes since it's a relatively quick and simple fix that would keep our sample names consistent. But I do agree that if we're going to update Utility networks in the project file, we should do the same for Route & directions and Cloud & portal.

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
yo1995: @maryharvey @vquach2404 I've addressed your feedback and make these category names unified everywhere. What do you think?

vquach2404: I like that they're uniform but you might be right about the labels. Maybe we should consider keeping the '&' and only change the capitalization. It keeps the labels short and much more readable compared to having a long line of all letters for a label.

yo1995:
Nudge to the thread - how do you think of the change to `Route & Directions` and `Cloud & Portal`

vquach2404:
> Nudge to the thread - how do you think of the change to `Route & Directions` and `Cloud & Portal`

Not sure if you saw my [comment](https://github.com/Esri/arcgis-runtime-samples-ios/pull/919/files#r470244708), but I think it'd be better as "Route & directions" and "Cloud & portal".

yo1995:
> Not sure if you saw my comment but I think it'd be better as "Route & directions" and "Cloud & portal".

Yep I saw it. I'm still waiting for inputs from @maryharvey and @philium . Pros and cons for making this change

- Pros: now category names are consistent everywhere: contentplist, folder name, metadata
- Cons: it might not be good looking

Personally I'm in favor of making this change, so that we won't need to be careful of mapping category names anymore, as now they are all the same everywhere. Besides, it makes sense to keep everything in sentence case, which is our new standard.

philium:
I personally would love to see us follow platform conventions inside the app, such as using title case for category and sample names. The names could be sentence case in the metadata files so as to follow the website convention. Same with the ampersand. The platform convention is to use it in titles. For examples, just open the Settings app: Display & Brightness, Home Screen & Clock, etc.

yo1995:
> I personally would love to see us follow platform conventions inside the app, such as using title case for category and sample names. The names could be sentence case in the metadata files so as to follow the website convention. Same with the ampersand. The platform convention is to use it in titles. For examples, just open the Settings app: Display & Brightness, Home Screen & Clock, etc.

So do you think instead of making changes this PR, we should make all the category names in content plist to title case? e.g. Edit Data, Display Information, etc.?

Either way, we should settle on a consistent casing. 😁

philium:
It is probably something to talk over with the team, but if we could follow iOS conventions inside the app while following Esri conventions on the developer site, that would be great.

maryharvey (approved):

yo1995:
@saratchandrakarumuri A bit of context: we want to make the category names casing consistent both for the app and for the developer site. 

Not sure if it breaks your tests... please give a try.

PR #922, opened 2020-07-29
New: Realistic lighting and shadows

yo1995:
## Description

Implementation of the `Realistic lighting and shadows` sample. A brief demo below. 

Link to the branch folder: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/New-RealisticLightingAndShadows-vNext/arcgis-ios-sdk-samples/Scenes/Realistic%20lighting%20and%20shadows

|Demo|
|------ |
|sunlight-demo|

~~Looking forward to dropping iOS 12 and start to take advantage of `SFSymbols`!~~

Please refer to `/runtime/common-samples/issues/175` for details.


philium:
@yo1995 Is this ready for review? I noticed that you pushed some changes after requesting review, so I wasn't sure. If it is, @vquach2404 could you please review it first? Then I'll take a look at it.

yo1995:
> @yo1995 Is this ready for review? I noticed that you pushed some changes after requesting review, so I wasn't sure. If it is, @vquach2404 could you please review it first? Then I'll take a look at it.

Yes it is ready for review. I updated according to samples team feedback last Friday. I'll first wait for @vquach2404 's review and solve the conflicts before yours.

For future PRs, probably Vivian and I can first finish reviewing each other before adding you as a reviewer, so that you get less notifications. 😁

arcgis-ios-sdk-samples/Scenes/Realistic lighting and shadows/RealisticLightingAndShadowsViewController.swift
```diff
@@ -0,0 +1,119 @@
...
+        present(alertController, animated: true)
+    }
+    
+    // MARK: UIViewController
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Add the source code button item to the right of navigation bar.
+        (self.navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["RealisticLightingAndShadowsViewController"]
+        sceneView.scene?.load { [weak self] (error) in
```
vquach2404, line 108: I would suggest creating a separate `loadScene` function and call it from the `sceneView` observer if possible to simply `viewDidLoad`

arcgis-ios-sdk-samples/Scenes/Realistic lighting and shadows/RealisticLightingAndShadowsViewController.swift
```diff
@@ -0,0 +1,120 @@
...
+        scene.operationalLayers.add(buildingsLayer)
+        // Load the scene.
+        scene.load { [weak self] (error) in
+            if let error = error {
+                self?.presentAlert(error: error)
+            } else {
+                let camera = AGSCamera(latitude: 45.54605, longitude: -122.69033, altitude: 941.00021, heading: 162.58544, pitch: 60.0, roll: 0)
+                self?.sceneView.setViewpointCamera(camera)
+            }
+        }
```
yo1995, line 75: I'd like to know your opinions on putting load function in this method, or even using the loadable pattern.

I feel like `makeMap()`/`makeScene()` functions should only create an object, and not manipulate UI even in escaping closures. But it also makes sense to load the object on creation, to avoid calling load again somewhere else.

yo1995:
@philium this is ready for review.

arcgis-ios-sdk-samples/Scenes/Realistic lighting and shadows/RealisticLightingAndShadowsViewController.swift
philium: It really depends. In this case, there's no reason to explicitly call load. The viewpoint camera can be set on the scene view as soon as it is available. There's no need to wait until the scene is loaded to set the viewpoint camera on the scene view.

arcgis-ios-sdk-samples/Scenes/Realistic lighting and shadows/RealisticLightingAndShadowsViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        let scene = AGSScene(basemap: .topographic())
+        scene.baseSurface = surface
+        scene.operationalLayers.add(buildingsLayer)
+        return scene
+    }
+    
+    // MARK: - Actions
+    
+    @IBAction func sliderValueChanged(_ sender: UISlider) {
+        dateComponents.minute = Int(sender.value)
```
philium, line 74: `dateComponents` is only used here in this method, so as far as I can tell, there's no reason to make it an instance property. I would just create it here and then get rid of the property.

```suggestion
        let dateComponents = DateComponents(minute: Int(sender.value))
```

arcgis-ios-sdk-samples/Scenes/Realistic lighting and shadows/RealisticLightingAndShadowsViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        let date = Calendar.current.date(byAdding: dateComponents, to: today)!
```
philium, line 75: `today` is only used here as well, right? In that case, unless there's a perform concern (which I don't think there would be), I would just move the calculation of it into this method and name it `startOfToday`. That way it always uses the current, which could change while the sample is open for a variety of reasons.

arcgis-ios-sdk-samples/Scenes/Realistic lighting and shadows/RealisticLightingAndShadowsViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        let alertController = UIAlertController(
+            title: "Choose a lighting mode for the scene view.",
+            message: nil,
+            preferredStyle: .actionSheet
+        )
+        let modes: [(name: String, mode: AGSLightingMode)] = [
+            ("Light and shadows", .lightAndShadows),
+            ("Light only", .light),
+            ("No light", .noLight)
+        ]
```
philium, line 90: Consider using `KeyValuePairs`:
```suggestion
        let modes: KeyValuePairs<String, AGSLightingMode> = [
            "Light and shadows": .lightAndShadows,
            "Light only": .light,
            "No light": .noLight
        ]
```

arcgis-ios-sdk-samples/Scenes/Realistic lighting and shadows/RealisticLightingAndShadowsViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        modes.forEach { (name, mode) in
+            let action = UIAlertAction(title: name, style: .default) { _ in
+                self.sceneView.sunLighting = mode
+            }
+            alertController.addAction(action)
+        }
+        let cancelAction = UIAlertAction(title: "Cancel", style: .cancel, handler: nil)
```
philium, line 97:
```suggestion
        let cancelAction = UIAlertAction(title: "Cancel", style: .cancel)
```

arcgis-ios-sdk-samples/Scenes/Realistic lighting and shadows/RealisticLightingAndShadows.storyboard
```diff
@@ -0,0 +1,104 @@
...
+                                        </connections>
+                                    </barButtonItem>
+                                    <barButtonItem width="42" style="plain" systemItem="fixedSpace" id="PVk-Ok-Zb2"/>
+                                </items>
+                            </toolbar>
+                            <toolbar opaque="NO" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="umr-Q2-zU9">
+                                <rect key="frame" x="0.0" y="774" width="414" height="44"/>
+                                <items>
+                                    <barButtonItem style="plain" systemItem="flexibleSpace" id="LMf-qC-QSu"/>
+                                    <barButtonItem title="AM" id="SKA-Sg-ke6">
```
philium, line 59: A titled bar button item should be used for something like this since the items still act like buttons. Tap on and hold on either AM or PM to see what I mean. Instead, could you use bar button items labels instead?

arcgis-ios-sdk-samples/Scenes/Realistic lighting and shadows/RealisticLightingAndShadowsViewController.swift
yo1995: Nice! New stuff learned. 🎉  [ref](https://developer.apple.com/documentation/swift/keyvaluepairs)

yo1995:
@saratchandrakarumuri Please take a look at this new sample! Thanks.

PR #923, opened 2020-07-30
Fix: Feature layer selection

yo1995:
This PR refactors the identify workflow for `Feature layer selection` sample using `AGSGeoView.identifyLayer(_:screenPoint:tolerance:returnPopupsOnly:maximumResults:completion:)` method.

ref: /runtime/common-samples/issues/2032


philium:
> I think we may want to rename the storyboard and view controller to spell out the entire sample name "FeatureLayerSelection__".

@vquach2404 Could you please create an issue for that? I agree that would be good to do, but I think doing it in a separate PR might be best.

arcgis-ios-sdk-samples/Features/Feature layer selection/FLSelectionViewController.swift
```diff
@@ -15,78 +15,83 @@
...
+        mapView.identifyLayer(featureLayer, screenPoint: screenPoint, tolerance: 12.0, returnPopupsOnly: false, maximumResults: 10) { [weak self] result in
+            guard let self = self else { return }
+            if let error = result.error {
+                self.presentAlert(error: error)
+                return
             }
+            // Un-select previous results.
+            self.featureLayer.unselectFeatures(self.selectedFeatures)
+            // Select current results.
+            self.selectedFeatures = result.geoElements.map { $0 as! AGSFeature }
```
philium, line 91: No need to use `map(_:)` here. Just cast the array, not each feature individually.

```suggestion
            self.selectedFeatures = result.geoElements as! [AGSFeature]
```

arcgis-ios-sdk-samples/Features/Feature layer selection/README.md
```diff
@@ -1,15 +1,35 @@
...
+Select features in a feature layer.
 
-## How to use the sample
+![Map with features](feature-layer-selection-1.png)
+![Selected features](feature-layer-selection-2.png)
+
+## Use case
 
-Tap on a feature on the map to select it
+Selecting features, whether by query or identify, can be an important step both in editing data and visualizing results. One possible use case would be to query a feature layer containing street furniture. A query might look for type "bench" and return a list of bench features contained in the features with an attribute of type bench. These might be selected for further editing or may just be highlighted visually.
```
philium, line 10: I think this should be either:
```suggestion
Selecting features, whether by query or identify, can be an important step in both editing data and visualizing results. One possible use case would be to query a feature layer containing street furniture. A query might look for type "bench" and return a list of bench features contained in the features with an attribute of type bench. These might be selected for further editing or may just be highlighted visually.
```
or:
```suggestion
Selecting features, whether by query or identify, can be an important step both in editing data and in visualizing results. One possible use case would be to query a feature layer containing street furniture. A query might look for type "bench" and return a list of bench features contained in the features with an attribute of type bench. These might be selected for further editing or may just be highlighted visually.
```

arcgis-ios-sdk-samples/Features/Feature layer selection/FLSelection.storyboard
```diff
@@ -1,35 +1,54 @@
...
+                                    <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Selected feature count shown here." textAlignment="center" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="MSB-XK-96k">
+                                        <rect key="frame" x="0.0" y="0.0" width="414" height="21"/>
+                                        <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                        <nil key="textColor"/>
+                                        <nil key="highlightedColor"/>
+                                    </label>
+                                </subviews>
+                                <color key="backgroundColor" white="1" alpha="0.59999999999999998" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                                <constraints>
+                                    <constraint firstItem="MSB-XK-96k" firstAttribute="top" secondItem="gTA-cM-d3f" secondAttribute="top" id="0NH-nf-VjW"/>
```
philium, line 34: Something seems odd about these constraints. The label's top and bottom should be constrained to the top and bottom of its superview, not the safe area. Right now there are no constraints to the bottom of the superview, meaning that its height is not well defined.

yo1995:
> @vquach2404 Could you please create an issue for that? I agree that would be good to do, but I think doing it in a separate PR might be best.

I can create another PR to change these names.

arcgis-ios-sdk-samples/Features/Feature layer selection/FLSelection.storyboard
yo1995: This is similar to one of previous discussions: https://github.com/Esri/arcgis-runtime-samples-ios/pull/837#discussion_r403031441

I want to create a "HUD" atop, without letting the label stretch into the safe area, while the translucent background should snap to superview's leading and trailing anchors. Even though currently status messages never overflow the width, I still want to play it safely.

image

So what I did is to embed the label in a view, snap the label to the safe area of the view, and set the view's leading and trailing to superview.

And for the top constraint of the view, it is to snap to top of map view. The height of this wrapping view is decided by the height of the label, so there is no bottom anchor for it. Do you think it's preferable to give an explicit height to it?

philium: > Do you think it's preferable to give an explicit height to it?

Yes and no. Not by means of an explicit height constraint, but there should be an unbroken line of constraints from one side of each view to the other that unambiguously determine both the size and position of the view. For the wrapping view, as far as I can tell, there are no constraint(s) that determines its height. To fix that, I would add a constraint from the top of the label to the top of the wrapper view and another constraint from the bottom of the label to the bottom of the wrapper view, replacing the current top and bottom constraints for the label. That way the wrapper view's height will be equal to the height of the label, which is determined by the label's intrinsic content size (which is based on the label's text).

yo1995: Addressed. Hope this time it looks right... 😄

yo1995:
Clicked merge too fast 🤥 . This PR didn't really change the expected result of the original sample, but it might be good to verify it works... @saratchandrakarumuri could you please take a look at this sample? Thanks.

saratchandrakarumuri:
@yo1995 Yup... sample works as expected without any issues

saratchandrakarumuri:
@yo1995 Yup... Sample working good without any issues

PR #924, opened 2020-07-31
New: Buffer list

yo1995:
## Description

Implementation of the `Buffer list` sample. A brief demo below. 

Link to the branch folder: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/New-BufferList-vNext/arcgis-ios-sdk-samples/Geometry/Buffer%20list

|Demo (design is updated now)|
|------ |
|buffer-list-demo|

Please refer to `/runtime/common-samples/issues/539` for details.


arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
```diff
@@ -0,0 +1,184 @@
...
+        (navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["BufferListViewController"]
+        // Load the map image layer.
+        mapImageLayer.load { [weak self] (error) in
+            if let error = error {
+                self?.presentAlert(error: error)
+                self?.setStatus(message: "Fail to load basemap.")
+            } else {
+                self?.setStatus(message: "Tap on the map to add buffers.")
+            }
+        }
```
yo1995, line 160: This loadable call can be moved into makeMap function.

arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
```diff
@@ -0,0 +1,184 @@
...
+// MARK: - AGSGeoViewTouchDelegate
+
+extension BufferListViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        // Only proceed when the tapped point is within the boundary.
+        guard AGSGeometryEngine.geometry(boundaryPolygon, contains: mapPoint) else {
+            setStatus(message: "Tap within the boundary to add buffer.")
+            return
+        }
+        // The spatial reference in this sample use US feet as unit.
```
vquach2404, line 173:
```suggestion
        // The spatial reference in this sample uses US feet as a unit.
```

arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
```diff
@@ -0,0 +1,184 @@
...
+        let radius = bufferRadius.converted(to: .feet).value
+        // Ensure that the buffer radius is a positive value.
+        guard radius > 0 else { return }
+        // Create and add graphic symbolizing the tap point.
+        let pointGraphic = AGSGraphic(geometry: mapPoint, symbol: nil)
+        tapLocationsGraphicsOverlay.graphics.add(pointGraphic)
+        // Keep track of tapped points and their radius.
```
vquach2404, line 180:
```suggestion
        // Keep track of tapped points and their radii.
```

arcgis-ios-sdk-samples/Geometry/Buffer list/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## How to use the sample
+
+Tap on the map to add points. Tap the create button to draw buffer(s) around the points (the radius of the buffer is determined by the value of the slider). Toggle on union if you want the result to union (combine) the buffers. Tap the trash button to start over. The red dashed envelope shows the area where you can expect reasonable results for planar buffer operations with the North Central Texas State Plane spatial reference.
```
vquach2404, line 13:
```suggestion
Tap on the map to add points. Tap the create button to draw buffer(s) around the points (the radius of the buffer is determined by the value of the slider). Toggle on union if you want the result to union (combine) the buffers. Tap the trash icon to start over. The red dashed envelope shows the area where you can expect reasonable results for planar buffer operations with the North Central Texas State Plane spatial reference.
```
Just a word suggestion.

arcgis-ios-sdk-samples/Geometry/Buffer list/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## How it works
+
+1. Use `class AGSGeometryEngine.bufferGeometries(_:distances:unionResults:)` to create a list of `AGSGeometry`.
```
vquach2404, line 17:
```suggestion
1. Use `class AGSGeometryEngine.bufferGeometries(_:distances:unionResults:)` to create a list of `AGSGeometry`s.
```
Consider making this plural.

arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
yo1995: Thanks! This is helpful. I don't know the plural form of radius before... 😅

arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
```diff
@@ -0,0 +1,184 @@
...
+        return overlay
+    }()
+    /// An overlay to display tapped locations with red circle symbols.
+    let tapLocationsGraphicsOverlay: AGSGraphicsOverlay = {
+        let overlay = AGSGraphicsOverlay()
+        let circleSymbol = AGSSimpleMarkerSymbol(style: .circle, color: .red, size: 10)
+        overlay.renderer = AGSSimpleRenderer(symbol: circleSymbol)
+        return overlay
+    }()
+    /// An array of tapped points and buffer radius (in US feet) tuple.
```
yo1995, line 82:
```suggestion
    /// An array of tapped points and buffer radii (in US feet) tuple.
```

vquach2404:
bufferList
After testing the sample some more, I realized that the radius doesn't update until after a new buffer has been created. Could you adjust it so that the value on the slider reflects the radius of the current buffer?

yo1995:
> After testing the sample some more, I realized that the radius doesn't update until after a new buffer has been created. Could you adjust it so that the value on the slider reflects the radius of the current buffer?

Yeah I was unsure about the UI design. 🤔  The existing samples from dotnet and Java are both desktop, and they use a text field for the radius. I found it intrusive to present an alert asking for radius value, so I used slider instead. 

But a slider cannot explicitly tell which buffer the value is associated with. Do you mean to read the slider value after the center point graphic is created, and before adding the next point?

The better design I could come up with is to add a "preview buffer" for each center point, to visualize the size of the buffer. But that would make implementation details diverge.

Maybe I'll present this sample this Friday and ask for feedback.

vquach2404:
> But a slider cannot explicitly tell which buffer the value is associated with. Do you mean to read the slider value after the center point graphic is created, and before adding the next point?

I just mean that the size of the current buffer that is being created should take the size that is reflected on the slider. So each time the value of the slider is changed, the radius should update to the new value.

yo1995:
> I just mean that the size of the current buffer that is being created should take the size that is reflected on the slider. So each time the value of the slider is changed, the radius should update to the new value.

I see. The original design doc is a bit misleading. The create button is actually "drawing" the graphics, while "creating" the buffers is done by tapping on the map. The sample is designed to be used as repeating "set a slider value -> tap on the map" multiple times, and tap "create" once, to create a list of buffers.

I understand you are trying to use the create button as an add button, so that each time it's tapped, it add a buffer with the slider radius and the center point. I also think this workflow make more sense, but we'll need to consult our team if we want to change the design.

arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
```diff
@@ -0,0 +1,184 @@
...
+            setStatus(message: "Buffers created.")
+        }
+    }
+    
+    @IBAction func clearButtonTapped(_ sender: UIBarButtonItem) {
+        bufferGraphicsOverlay.graphics.removeAllObjects()
+        tapLocationsGraphicsOverlay.graphics.removeAllObjects()
+        tappedPointsAndRadius.removeAll()
+        setStatus(message: "Buffers removed. Tap on the map to add buffers.")
+    }
```
yo1995, line 133: Clear button and create button should have some UI binding

- After clear is tapped, set both to disabled
- When more than 0 buffer exists, enable both buttons

yo1995:
I'm moving this PR back to draft, and will discuss with our team on Friday about some design issues mentioned above.

yo1995:
I've addressed some comments from Friday's review and decide to use the textField UI. However, since the deisgn is still being refactored by Trevor, I'll put it on hold for a while.

arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+    }()
+    
+    /// An overlay to display the boundary.
+    var boundaryGraphicsOverlay: AGSGraphicsOverlay {
+        let overlay = AGSGraphicsOverlay()
+        let lineSymbol = AGSSimpleLineSymbol(style: .dash, color: .red, width: 5)
+        let boundaryGraphic = AGSGraphic(geometry: boundaryPolygon, symbol: lineSymbol)
+        overlay.graphics.add(boundaryGraphic)
+        return overlay
+    }
```
yo1995, line 62: Since this overlay depends on the geometry of the `boundaryPolygon`, it is better to be instantiated by an instance method. However, I found it looking more "symmetric" to also create it as a computed variable. Wait for Phil's thought.

yo1995 (commented):
@vquach2404 Please take another look as we are moving forward with the current datasource! Thanks.

vquach2404 (commented):
simulator_screenshot_3A91E2B2-679A-46FA-9E0E-28F0194D1D6E
Please add some padding in between the Union label and the switch.

vquach2404:
What do you think about creating the buffers immediately after the user enters its size? The Union switch could also recreate the buffers if the user changes the switch after the buffers have been made. That way the user can visualize how large the buffer will be and you might even be able to include an "undo last point" button.

yo1995:
> What do you think about creating the buffers immediately after the user enters its size? The Union switch could also recreate the buffers if the user changes the switch after the buffers have been made. That way the user can visualize how large the buffer will be and you might even be able to include an "undo last point" button.

In this case each time the user tap on the map or toggle the switch, the whole buffer will get re-created, which might be not efficient. I'll make another commit later today to give it a try.

vquach2404 (commented):
Thanks for making those changes! I think it looks great now and I feel like this workflow is much more intuitive. 

Also, is the cancel button supposed to be blue?
Simulator Screen Shot - iPhone 11 Pro Max - 2020-10-14 at 16 44 34

yo1995:
> Also, is the cancel button supposed to be blue?

It should be purple. It seems only to be blue on iOS 14. 
Will need to consult Phil to see why it is blue when created... 🤔

arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
```diff
@@ -0,0 +1,238 @@
...
+    }()
+    /// An overlay to display tapped locations with red circle symbols.
+    let tappedLocationsGraphicsOverlay: AGSGraphicsOverlay = {
+        let overlay = AGSGraphicsOverlay()
+        let circleSymbol = AGSSimpleMarkerSymbol(style: .circle, color: .red, size: 10)
+        overlay.renderer = AGSSimpleRenderer(symbol: circleSymbol)
+        return overlay
+    }()
+    /// An array of tapped points and buffer radii (in US feet) tuple.
+    var tappedPointsAndRadius = [(point: AGSPoint, radius: Double)]() {
```
philium, line 79: The plural of radius is _radii_ or _radiuses_.

arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
```diff
@@ -0,0 +1,238 @@
...
+            setStatus(message: "Tap on the map to add buffers.")
+            return
+        }
+        
+        // Create the buffers.
+        // Notice: the radius distances has the same unit of the map's spatial reference's unit.
+        // In this case, the statePlaneNorthCentralTexas spatial reference uses US feet.
+        if let bufferPolygon = AGSGeometryEngine.bufferGeometries(tappedPointsAndRadius.map { $0.point }, distances: tappedPointsAndRadius.map { NSNumber(value: $0.radius) }, unionResults: isUnionSwitch.isOn) {
+            // Add graphics symbolizing the tap point.
+            tappedLocationsGraphicsOverlay.graphics.addObjects(from: tappedPointsAndRadius.map { AGSGraphic(geometry: $0.point, symbol: nil) })
```
philium, line 150: Instead of iterating over `tappedPointsAndRadius` multiple times, I would suggest iterating over it once to get the array of points and radii:

```suggestion
        let (points, radii) = tappedPointsAndRadius.reduce(into: ([AGSPoint](), [NSNumber]())) { (result, pointAndRadius) in
            result.0.append(pointAndRadius.point)
            result.1.append(pointAndRadius.radius as NSNumber)
        }
        if let bufferPolygon = AGSGeometryEngine.bufferGeometries(points, distances: radii, unionResults: isUnionSwitch.isOn) {
            // Add graphics symbolizing the tap point.
            tappedLocationsGraphicsOverlay.graphics.addObjects(from: points.map { AGSGraphic(geometry: $0, symbol: nil) })
```

FYI: Double can be unconditionally cast to `NSNumber` (`double as NSNumber`, `doubles as [NSNumber]`, etc.). I don't know if casting is more efficient that initializing, but it is the preferred approach.

arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
```diff
@@ -0,0 +1,238 @@
...
+
+extension BufferListViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        // Only proceed when the tapped point is within the boundary.
+        guard AGSGeometryEngine.geometry(boundaryPolygon, contains: mapPoint) else {
+            setStatus(message: "Tap within the boundary to add buffer.")
+            return
+        }
+        // Use an alert to get radius input from user.
+        let alert = UIAlertController(title: "Provide a buffer radius", message: "Between 0 and 300 \(distanceFormatter.string(from: bufferRadius.unit))", preferredStyle: .alert)
```
philium, line 187: I'm not sure it makes sense to split that sentence across the title and the message. It might be better to just combine them into a single sentence and use that as the title. Don't forget to add a period to the end!

image

arcgis-ios-sdk-samples/Geometry/Buffer list/BufferListViewController.swift
```diff
@@ -0,0 +1,238 @@
...
+        didSet {
+            undoBarButtonItem.isEnabled = !tappedPointsAndRadius.isEmpty
+            clearBarButtonItem.isEnabled = !tappedPointsAndRadius.isEmpty
+            // Redraw the buffers.
+            drawBuffers()
+        }
+    }
+    
+    /// The radius of the buffer.
+    var bufferRadius: Measurement<UnitLength> = Measurement(value: 0, unit: .miles)
```
philium, line 89: Correct me if I'm wrong, but it seems like this property isn't really necessary. The unit is used in one place, but the unit is always the same, so you could either declare a constant for it or just use `UnitLength.miles` directly. The value is only referenced in `doneAction`'s handler, so there really isn't a reason to store it in a property.

philium (approved):

yo1995:
@saratchandrakarumuri Please take a look at this new sample 😁 .
@philium Sorry for dismissing your review but I had to solve the merge conflict. Please give another approval! Thanks.

philium (changes requested):
The merged caused some items to become unordered, so Xcode automatically makes some changes to the project file when it is saved. Please make those changes in this PR.

FYI: Xcode keeps items sorted by identifier, so seeing items identifiers out-of-order is a clue that the project file needs to be re-saved.

image

yo1995:
> The merged caused some items to become unordered...

Ah my bad. 🤦  I finally find out it is GitHub online merge solving editor that makes Xcode unhappy. Next time I'll proceed with more caution.

cc @vquach2404 in the future when resolving merge conflict in GitHub online editor, be sure to let Xcode perform the auto adjustments before merging the PR to `v.next`.

philium (approved):

saratchandrakarumuri (approved):
Looks good

PR #928, opened 2020-08-05
Edit features with feature-linked annotation implementation

vquach2404:
Implementation of Edit features with feature-linked annotation implementation. Looking forward to feedback!


arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+    @IBOutlet weak var mapView: AGSMapView! {
+        didSet {
+            // Create the map with a light gray canvas basemap centered on Loudoun, Virginia.
+            let map = AGSMap(basemapType: .lightGrayCanvasVector, latitude: 39.0204, longitude: -77.4159, levelOfDetail: 18)
+            mapView.map = map
+            loadGeodatabase()
+        }
+    }
+    private var selectedFeature: AGSFeature?
+    private var selectedFeatureIsPolyline = false
```
yo1995, line 28: I think it's OK to keep them private, but Phil might ask to remove them. Let's wait for his response.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+                                }
+                            }
+                        }
+                        // Ensure the feature is not nil and select it.
+                        guard let selectedFeature = self.selectedFeature else { return }
+                        featureLayer.select(selectedFeature)
+                        // If the selected feature is a point, prompt the edit attributes alert.
+                        if selectedFeature.geometry?.geometryType.rawValue == 1 {
+                            self.showEditableAttributes(selectedFeature: selectedFeature)
+                        } else if selectedFeature.geometry?.geometryType.rawValue == 3 {
```
yo1995, line 104: Actually I think it's more readable if the whole block (from line 86) is a `switch case` control flow, so that you handle `selectedFeature` as polyline and point separately, rather than using `if else`.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+                                    // Set selected feature to nil.
+                                    self.selectedFeature = nil
+                                    // Show a message to select straight (single segment) polylines only.
+                                    self.presentAlert(title: "Make a different selection", message: "Select straight (single segment) polylines only.")
+                                    return
+                                }
+                            }
+                        }
+                        // Ensure the feature is not nil and select it.
+                        guard let selectedFeature = self.selectedFeature else { return }
```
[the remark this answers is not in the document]
vquach2404: I put it there because line 84 changes `self.selectedFeature` and was causing complications if I guarded before that. An alternative is guarding before line 86. What're your thoughts?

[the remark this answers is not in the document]
vquach2404: I can move it higher up after line 84 but can't move it higher than that because `self.selectedFeature` is nil prior to that line.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+        alert.addAction(UIAlertAction(title: "Cancel", style: .cancel) { _ in
+            self.clearSelection()
+            self.selectedFeature = nil
+        })
+        // Add a "Done" option to complete the editing process and close the alert.
+        alert.addAction(UIAlertAction(title: "Done", style: .default) { _ in
+            let addressTextField = alert.textFields?[0]
+            let streetTextField = alert.textFields?[1]
+            // Set the new values for the attributes and update the selected feature.
+            selectedFeature.attributes["AD_ADDRESS"] = Int((addressTextField?.text)!)
```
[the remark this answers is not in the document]
vquach2404: The value held in `AGSFeature.attributes` cannot be an Int, only NSNumber -- so it was the best way to display the string value without it being an optional. And it is later casted as an `Int` because there was no `NSNumber`  initializer that would accept the value from the text field.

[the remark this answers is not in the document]
yo1995: I see. The latter case String to Int can utilize `NumberFormatter`, but it might be an overkill. Using Int.init should be fine. 👍

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
vquach2404: I could include the block starting from line 86 because it complicated things with the alert but I implemented the switch case.

vquach2404:
@yo1995 Thanks for the suggestions! I made some changes and it is ready for another review.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+class EditFeaturesWithFeatureLinkedAnnotationViewController: UIViewController, AGSGeoViewTouchDelegate {
+    @IBOutlet weak var mapView: AGSMapView! {
+        didSet {
+            // Set the touch delegate.
+            self.mapView.touchDelegate = self
```
yo1995, line 22:
```suggestion
            mapView.touchDelegate = self
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+            // Create the map with a light gray canvas basemap centered on Loudoun, Virginia.
+            let map = AGSMap(basemapType: .lightGrayCanvasVector, latitude: 39.0204, longitude: -77.4159, levelOfDetail: 18)
+            mapView.map = map
+            loadGeodatabase()
```
yo1995, line 26: Phil might mention this call site... It might be OK to call the load here though, as load function is adding layers to the map.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+                            let polylineArray = polyline.parts.array()
+                            polylineArray.forEach { part in
+                                if part.pointCount > 2 {
+                                    // Set selected feature to nil.
+                                    self.selectedFeature = nil
+                                    // Show a message to select straight (single segment) polylines only.
+                                    self.presentAlert(title: "Make a different selection", message: "Select straight (single segment) polylines only.")
+                                }
+                            }
+                            return
```
yo1995:
```suggestion
```
returned too early here 😅

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+        // Identify across all layers
+        mapView.identifyLayers(atScreenPoint: screenPoint, tolerance: 10.0, returnPopupsOnly: false) { (results: [AGSIdentifyLayerResult]?, error: Error?) in
+            if let error = error {
+                self.presentAlert(error: error)
+            } else {
+                results?.forEach { (result) in
+                    if let featureLayer = result.layerContent as? AGSFeatureLayer {
+                        // Get a reference to the identified feature
+                        self.selectedFeature = result.geoElements[0] as? AGSFeature
+                        guard let selectedFeature = self.selectedFeature else { return }
```
yo1995:
```suggestion
                        // Get a reference to the identified feature
                        guard let selectedFeature = result.geoElements[0] as? AGSFeature else { return }
                        self.selectedFeature = selectedFeature
```
^^ What I was suggesting. It looks nicer this way, though it makes no difference in result.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+            } else {
+                movePoint(mapPoint: mapPoint)
+            }
+        }
+    }
+    
+    func selectFeature(screenPoint: CGPoint) {
+        // Clear any previously selected features.
+        clearSelection()
+        selectedFeature = nil
```
yo1995, line 76: It seems `func selectFeature(screenPoint: CGPoint)` is only called when `selectedFeature == nil`. Are these 2 lines needed? 🤔 (I think at lease nilify is not needed) If you cannot figure it out, we may leave it to Phil's judgement.

If you want to ensure when `selectedFeature` is nil, selection is guaranteed to be cleared, maybe consider using one of its observers (willSet, didSet).

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+            if let error = error {
+                self.presentAlert(error: error)
+            } else {
+                results?.forEach { (result) in
+                    if let featureLayer = result.layerContent as? AGSFeatureLayer {
+                        // Get a reference to the identified feature
+                        self.selectedFeature = result.geoElements[0] as? AGSFeature
+                        guard let selectedFeature = self.selectedFeature else { return }
+                        // If the selected feature is a polyline with any part containing more than one segment (i.e. a curve).
+                        if let polyline = selectedFeature.geometry as? AGSPolyline {
```
yo1995: This could also get moved into the switch case. Since you decide a geometry's type is a polyline, you can confidently (force) cast it to a polyline.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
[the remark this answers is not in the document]
vquach2404: Update: I had to leave the guard statement at its original place because of the possible value change of `self.selectedFeature`at line 93. Moving the guard statement prior to that line caused the program to continue with the curved polyline selection and made `selectedFeatureIsPolyline` true when it should be false.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+        // Clear any previously selected features.
+        clearSelection()
+        selectedFeature = nil
+        
+        // Identify across all layers
+        mapView.identifyLayers(atScreenPoint: screenPoint, tolerance: 10.0, returnPopupsOnly: false) { (results: [AGSIdentifyLayerResult]?, error: Error?) in
+            if let error = error {
+                self.presentAlert(error: error)
+            } else {
+                results?.forEach { (result) in
```
yo1995, line 83: One advantage to guard at top in this closure is to get rid of non-geometry results fast.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,229 @@
...
+            self.clearSelection()
+            self.selectedFeatureIsPolyline = false
+            self.selectedFeature = nil
+        })
+        alert.addAction(UIAlertAction(title: "Yes", style: .default) { _ in
+            // Get the selected feature's geometry as a polyline nearest to the map point.
+            guard let polyline = selectedFeature.geometry as? AGSPolyline else { return }
+            // Get the nearest vertex to the map point on the selected feature polyline.
+            guard let nearestVertex = AGSGeometryEngine.nearestVertex(in: polyline, to:
+                (AGSGeometryEngine.projectGeometry(mapPoint, to: (polyline.spatialReference)!) as? AGSPoint)!) else { return }
```
yo1995, line 192:
```suggestion
            guard let polyline = selectedFeature.geometry as? AGSPolyline,
                let selectedMapPoint = AGSGeometryEngine.projectGeometry(mapPoint, to: polyline.spatialReference!) as? AGSPoint,
                let nearestVertex = AGSGeometryEngine.nearestVertex(in: polyline, to: selectedMapPoint) else { return }
```
I never see `AGSGeometryEngine` geometry fail, so it might be safe to force unwrap everywhere. If to use guard, you can combine them into 1 guard statement.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+![Image of edit features with feature-linked annotation](feature-linked-annotation.png)
+
+## Use case
+
+Annotation is useful for displaying text that you don't want to move or resize when the map is panned or zoomed (unlike labels which will move and resize). Feature-linked annotation will update when a feature attribute referenced by the annotation expression is also updated. Additionally, the position of the annotation will transform to match any transformation to the linked feature's geometry.
+
+## How to use the sample
+
+Pan and zoom the map to see that the text on the map is annotation, not labels. Tap one of the address points to update the house number (AD\_ADDRESS) and street name (ST\_STR\_NAM). Tap one of the dashed parcel polylines and tap another location to change its geometry. NOTE: Selection is only enabled for points and straight (single segment) polylines. The feature-linked annotation will update accordingly.
```
yo1995, line 13: I'm a bit confused about the field names. In code they are `AD_ADDRESS` and `ST_STR_NAM`, but here they have backslash. Should they be unified?

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,229 @@
...
+            }
+        }
+    }
+    
+    func selectFeature(screenPoint: CGPoint) {
+        // Clear any previously selected features.
+        clearSelection()
+        selectedFeature = nil
+        
+        // Identify across all layers
```
yo1995, line 78:
```suggestion
        // Identify across all layers.
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,229 @@
...
+    }
+    
+    // Create an alert dialog with edit texts to allow editing of the given feature's 'AD_ADDRESS' and 'ST_STR_NAM' attributes.
+    func showEditableAttributes(selectedFeature: AGSFeature) {
+        // Create an alert controller and customize the title and message.
+        let alert = UIAlertController(title: "Edit feature attributes", message: "Edit the 'AD_ADDRESS' and 'ST_STR_NAM' attributes.", preferredStyle: .alert)
+        // Add text fields to prompt user input.
+        alert.addTextField()
+        alert.addTextField()
+        // Popuplate the text fields with the current values.
```
yo1995, line 124:
```suggestion
        // Populate the text fields with the current values.
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/README.md
vquach2404: The backslashes allow them to be rendered correctly. Without them, half of the phrases were italicized.

yo1995: Ha nice catch! I see. Thanks. 😄

vquach2404:
@yo1995 thanks for the review! @philium ready for your review!

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+            }
+        }
+    }
+    
+    // Create an alert dialog with edit texts to allow editing of the given feature's 'AD_ADDRESS' and 'ST_STR_NAM' attributes.
+    func showEditableAttributes(selectedFeature: AGSFeature) {
+        // Create an alert controller and customize the title and message.
+        let alert = UIAlertController(title: "Edit feature attributes", message: "Edit the 'AD_ADDRESS' and 'ST_STR_NAM' attributes.", preferredStyle: .alert)
+        // Add text fields to prompt user input.
+        alert.addTextField()
```
yo1995, line 122: It might be good to use the `configurationHandler` to set the `textField`'s properties, i.e. 
```swift
        alert.addTextField { textField in
            textField.keyboardType = .numberPad
            textField.text = (selectedFeature.attributes["AD_ADDRESS"] as! NSNumber).stringValue
        }
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+    private var selectedFeature: AGSFeature?
+    private var selectedFeatureIsPolyline = false
+    
+    func loadGeodatabase() {
+        // Load geodatabase from shared resources.
+        let geodatabaseURL = Bundle.main.url(forResource: "loudoun_anno", withExtension: ".geodatabase")!
+        let geodatabase = AGSGeodatabase(fileURL: geodatabaseURL)
+        geodatabase.load { (error: Error?) in
```
philium, line 36: There's several issues here:

- Nothing is holding on to the geodatabase. After the geodatabase is loaded, it will be deallocated when the completion handler is done executing. The geodatabase should be an instance property of the view controller.
- It appears that the completion handler is never called.
- The completion handler is strongly capturing `self`.
- The completion handler is strongly capturing the geodatabase. It should be an unowned reference or it can be accessed through `self` after making it an instance property.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+            if let error = error {
+                self.presentAlert(error: error)
+            } else {
+                guard let map = self.mapView.map else { return }
+                // Create feature layers from tables in the geodatabase.
+                let parcelLinesFeatureLayer = AGSFeatureLayer(featureTable: geodatabase.geodatabaseFeatureTable(withName: "ParcelLines_1")!)
```
philium, line 42: Consider consolidating this code. For example:

```swift
let featureTableNames = ["ParcelLines_1", "Loudoun_Address_Points_1"]
let featureTables = featureTableNames.compactMap { geodatabase.geodatabaseFeatureTable(withName: $0) }
let featureLayers = featureTables.map(AGSFeatureLayer.init)
map.operationalLayers.addObjects(from: featureLayers)
```

Similar for the annotation layers.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+                map.operationalLayers.add(parcelLinesFeatureLayer)
+                map.operationalLayers.add(addressPointFeatureLayer)
+                // Add the annotation layers to the map.
+                map.operationalLayers.add(parcelLinesAnnotationLayer)
+                map.operationalLayers.add(addressPointsAnnotationLayer)
+            }
+        }
+    }
+    
+    // MARK: - AGSGeoViewTouchDelegate
```
philium, line 57: Sample view controllers typically conform to `AGSGeoViewTouchDelegate` in an extension. Is there a reason why that couldn't be done with this sample?

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+    // Select the nearest feature or move the point or polyline vertex to the given screen point.
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        // If a feature hasn't been selected.
+        if selectedFeature == nil {
+            selectFeature(screenPoint: screenPoint)
+        } else {
+            // Move the feature.
+            if selectedFeatureIsPolyline {
```
philium, line 65: `selectedFeatureIsPolyline` is a stored property, but it is only read here. Rather than setting it in multiple places, the value is easily derived from the selected feature itself, so it would be better to derive it here, like so:

```swift
if let selectedFeature = selectedFeature {
    if selectedFeature.geometry?.geometryType == .polyline {
        ...
    } else {
        ...
    }
} else {
    ...
}
```

Then `selectedFeatureIsPolyline` wouldn't even be needed and could just be removed.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+                movePolylineVertex(mapPoint: mapPoint)
+            } else {
+                movePoint(mapPoint: mapPoint)
+            }
+        }
+    }
+    
+    func selectFeature(screenPoint: CGPoint) {
```
philium, line 73: The name of this method should be `selectFeature(at:)`.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+        
+        // Identify across all layers.
+        mapView.identifyLayers(atScreenPoint: screenPoint, tolerance: 10.0, returnPopupsOnly: false) { (results: [AGSIdentifyLayerResult]?, error: Error?) in
+            if let error = error {
+                self.presentAlert(error: error)
+                return
+            }
+            results?.forEach { (result) in
+                //Get a reference to the identified feature layer and feature.
+                guard let featureLayer = result.layerContent as? AGSFeatureLayer, let selectedFeature = result.geoElements[0] as? AGSFeature else { return }
```
philium, line 86:
```suggestion
                guard let featureLayer = result.layerContent as? AGSFeatureLayer, let selectedFeature = result.geoElements.first as? AGSFeature else { return }
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+        selectedFeature = nil
+        
+        // Identify across all layers.
+        mapView.identifyLayers(atScreenPoint: screenPoint, tolerance: 10.0, returnPopupsOnly: false) { (results: [AGSIdentifyLayerResult]?, error: Error?) in
+            if let error = error {
+                self.presentAlert(error: error)
+                return
+            }
+            results?.forEach { (result) in
+                //Get a reference to the identified feature layer and feature.
```
philium, line 85:
```suggestion
                // Get a reference to the identified feature layer and feature.
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+    
+    func selectFeature(screenPoint: CGPoint) {
+        // Clear any previously selected features.
+        clearSelection()
+        selectedFeature = nil
+        
+        // Identify across all layers.
+        mapView.identifyLayers(atScreenPoint: screenPoint, tolerance: 10.0, returnPopupsOnly: false) { (results: [AGSIdentifyLayerResult]?, error: Error?) in
```
philium, line 79: The completion handler is holding a strong reference to `self`.

philium, line 79: `identifyLayers(atScreenPoint:tolerance:returnPopupsOnly:completion:)` returns a cancellable. You might want to store it as an instance property and cancel the current identify operation before starting a new one. That could happen if the user taps multiple times in rapid succession.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+                        if part.pointCount > 2 {
+                            // Set the value to nil.
+                            self.selectedFeature = nil
+                            // Present the alert.
+                            self.presentAlert(title: "Make a different selection", message: "Select straight (single segment) polylines only.")
+                            return
+                        }
+                    }
+                    // If the selected feature is a valid polyline, set the value to true.
+                    self.selectedFeatureIsPolyline = true
```
philium, line 105: The selected features doesn't need to be set to `nil` because that as already done before the identify operation, correct? Also, this could be simplified by using `contains(where:)`:
```suggestion
                    // If the selected feature is a polyline with any part containing more than one segment (i.e. a curve), present an alert.
                    if polyline.parts.array().contains(where: { $0.pointCount > 2 }) {
                        self.presentAlert(title: "Make a different selection", message: "Select straight (single segment) polylines only.")
                        return
                    }
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+                default:
+                    return
+                }
+                // Select the feature.
+                featureLayer.select(selectedFeature)
+                // Set the feature globally.
+                self.selectedFeature = selectedFeature
```
philium, line 112: The results array is being iterated, so it seems like there's the potential for multiple features to be selected. Should `selectedFeature` actually be a set?

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+    
+    // Create an alert dialog with edit texts to allow editing of the given feature's 'AD_ADDRESS' and 'ST_STR_NAM' attributes.
+    func showEditableAttributes(selectedFeature: AGSFeature) {
+        // Create an alert controller and customize the title and message.
+        let alert = UIAlertController(title: "Edit feature attributes", message: "Edit the 'AD_ADDRESS' and 'ST_STR_NAM' attributes.", preferredStyle: .alert)
+        // Add text fields to prompt user input.
+        alert.addTextField()
+        alert.addTextField()
+        // Populate the text fields with the current values.
+        alert.textFields?[0].text = (selectedFeature.attributes["AD_ADDRESS"] as! NSNumber).stringValue
```
philium, line 125: Can `attributes["AD_ADDRESS"]` be `nil`? If so, the force cast would be cause a crash in that case.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+        alert.addAction(UIAlertAction(title: "Done", style: .default) { _ in
+            let addressTextField = alert.textFields?[0]
+            let streetTextField = alert.textFields?[1]
+            // Set the new values for the attributes and update the selected feature.
+            selectedFeature.attributes["AD_ADDRESS"] = Int((addressTextField?.text)!)
+            selectedFeature.attributes["ST_STR_NAM"] = streetTextField?.text
+            selectedFeature.featureTable?.update(selectedFeature)
+        })
+        // Present the alert.
+        self.present(alert, animated: true, completion: nil)
```
philium, line 145:
```suggestion
        present(alert, animated: true)
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+    }
+    
+    // Move the currently selected point feature to the given map point, by updating the selected feature's geometry and feature table.
+    func movePoint(mapPoint: AGSPoint) {
```
philium, line 149: The name of this function should be something like`moveSelectedFeature(to:)`.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+        guard let selectedFeature = self.selectedFeature else { return }
+        // Create an alert to confirm that the user wants to update the geometry.
+        let alert = UIAlertController(title: "Confirm update", message: "Would you like to update the geometry?", preferredStyle: .alert)
+        // Clear the selection and selected feature if "No" is selected.
+        alert.addAction(UIAlertAction(title: "No", style: .cancel) { _ in
+            self.clearSelection()
+            self.selectedFeature = nil
```
philium, line 156: I see this in multiple places. Could `clearSelection()` set `selectedFeature` to `nil`?

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+            selectedFeature.attributes["ST_STR_NAM"] = streetTextField?.text
+            selectedFeature.featureTable?.update(selectedFeature)
+        })
+        // Present the alert.
+        self.present(alert, animated: true, completion: nil)
+    }
+    
+    // Move the currently selected point feature to the given map point, by updating the selected feature's geometry and feature table.
+    func movePoint(mapPoint: AGSPoint) {
+        guard let selectedFeature = self.selectedFeature else { return }
```
philium, line 150: `self` isn't needed here.
```suggestion
        guard let selectedFeature = selectedFeature else { return }
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+        // Create an alert to confirm that the user wants to update the geometry.
+        let alert = UIAlertController(title: "Confirm update", message: "Would you like to update the geometry?", preferredStyle: .alert)
```
philium, line 152: I suggest you reread the section in the HIG regarding alerts. A better message would be something like "Are you sure you want to move the selected feature?" and the button titles should be "Move" and "Cancel", not "Yes" and "No".

Similar with the alert in `movePolylineVertex(mapPoint:)`.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+                    self.presentAlert(error: error)
+                } else {
+                    // Clear selection of polyline.
+                    self.clearSelection()
+                    self.selectedFeature = nil
+                }
+            }
+        })
+        // Present the alert.
+        self.present(alert, animated: true, completion: nil)
```
philium, line 173:
```suggestion
        present(alert, animated: true)
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+    }
+    
+    // Move the last of the vertex point of the currently selected polyline to the given map point, by updating the selected feature's geometry and feature table.
+    func movePolylineVertex(mapPoint: AGSPoint) {
```
philium, line 177: The name of this function should be something like `moveLastVertexOfSelectedFeature(to:)`.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+                    self.presentAlert(error: error)
+                } else {
+                    // Clear selection of polyline.
+                    self.clearSelection()
+                    self.selectedFeatureIsPolyline = false
+                    self.selectedFeature = nil
+                }
+            }
+        })
+        self.present(alert, animated: true, completion: nil)
```
philium, line 211:
```suggestion
        present(alert, animated: true)
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+        // Clear the selection and selected feature if "No" is selected.
+        alert.addAction(UIAlertAction(title: "No", style: .cancel) { _ in
+            self.clearSelection()
+            self.selectedFeature = nil
+        })
+        alert.addAction(UIAlertAction(title: "Yes", style: .default) { _ in
+            // Set the selected feature's geometry to the new map point.
+            selectedFeature.geometry = mapPoint
+            // Update the selected feature's feature table.
+            selectedFeature.featureTable?.update(selectedFeature) { (error: Error?) in
```
philium, line 162: Completion handler is strongly capturing `self`.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+                let nearestVertex = AGSGeometryEngine.nearestVertex(in: polyline, to: selectedMapPoint) else { return }
+            let polylineBuilder = AGSPolylineBuilder(polyline: polyline)
+            // Get the part of the polyline nearest to the map point.
+            polylineBuilder.parts[nearestVertex.partIndex].removePoint(at: nearestVertex.partIndex)
+            // Add the map point as the new point on the polyline.
+            polylineBuilder.parts[nearestVertex.partIndex].addPoint(AGSGeometryEngine.projectGeometry(mapPoint, to: polylineBuilder.parts[nearestVertex.partIndex].spatialReference!) as! AGSPoint)
+            // Set the selected feature's geometry to the new polyline.
+            selectedFeature.geometry = polylineBuilder.toGeometry()
+            // Update the selected feature's feature table.
+            selectedFeature.featureTable?.update(selectedFeature) { (error: Error?) in
```
philium, line 200: Completion handler is strongly capturing `self`.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotation.storyboard
```diff
@@ -0,0 +1,42 @@
...
+                                <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
+                                <color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
+                            </view>
+                        </subviews>
+                        <color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
+                        <constraints>
+                            <constraint firstAttribute="bottom" secondItem="ShL-2L-KLv" secondAttribute="bottom" id="S29-b1-app"/>
+                            <constraint firstAttribute="trailing" secondItem="ShL-2L-KLv" secondAttribute="trailing" id="TUw-FV-Ejk"/>
+                            <constraint firstItem="ShL-2L-KLv" firstAttribute="leading" secondItem="GJn-Lh-FXz" secondAttribute="leading" id="VLT-a8-Pd0"/>
+                            <constraint firstItem="ShL-2L-KLv" firstAttribute="top" secondItem="GJn-Lh-FXz" secondAttribute="top" id="hbe-fB-wc8"/>
```
philium, line 29: The top of the map view should be constrained to the top of the safe area so that it doesn't underlap the navigation bar, which is not translucent.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
vquach2404: Only one feature is to be selected but the results are narrowed down on line 73. Is there another method to select the desired feature that avoids iterating through the array? I've tried a few methods of getting the first result but they didn't work out.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
vquach2404: No, it can't be nil.

vquach2404:
@philium @yo1995 Thanks for the feedback! I've addressed your comments and also added a textfield observer to ensure that the user doesn't enter empty strings. Please let me know what you think!

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
philium: The thing with `forEach` is that there's no way to stop iterating. The provided closure will be called for each element in the sequence. I haven't tested it, but it seems like this would work:

```swift
if let error = error {
    ...
} else if let results = results,
          let firstResult = results.first(where: { $0.layerContent is AGSFeature && !$0.geoElements.isEmpty }) {
    let featureLayer = firstResult.layerContent as! AGSFeatureLayer
    let selectedFeature = firstResult.geoElements.first as! AGSFeature
    ...
}
```

yo1995: Nice catch! I noticed this for another forEach earlier but missed this one...

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,246 @@
...
+        }
+        alert.addAction(doneAction)
+        // Add a text field to prompt the user to input the street address.
+        alert.addTextField { (textField) in
+            // Prompt a number pad for the address.
+            textField.keyboardType = .numberPad
+            // Populate the text fields with the current value.
+            textField.text = (selectedFeature.attributes["AD_ADDRESS"] as! NSNumber).stringValue
+            // Add an observer to ensure the user does not input an empty string.
+            _ = NotificationCenter.default.addObserver(forName: UITextField.textDidChangeNotification, object: textField, queue: OperationQueue.main, using: {_ in
```
yo1995, line 117: Should this be removed in deinit?

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,246 @@
...
+                }
+            }
+        })
+        present(alert, animated: true)
+    }
+    
+    // Clear selection from all feature layers.
+    func clearSelection() {
+        selectedFeature = nil
+        self.mapView.map?.operationalLayers.forEach { layer in
```
yo1995, line 213:
```suggestion
        mapView.map?.operationalLayers.forEach { layer in
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/README.md
```diff
@@ -0,0 +1,47 @@
...
+Pan and zoom the map to see that the texts on the map are annotations, not labels. Tap one of the address points to update the house number (AD\_ADDRESS) and street name (ST\_STR\_NAM). Tap one of the dashed parcel polylines and tap another location to change its geometry. NOTE: Selection is only enabled for points and straight (single segment) polylines. The feature-linked annotation will update accordingly.
+
+## How it works
+
+1. Load the geodatabase using `AGSLoadable.load(completion:)`. NOTE: Read/write geodatabases should normally come from an `AGSGeodatabaseSyncTask`, but this has been omitted here.
+2. Create `AGSFeatureLayer`s from the geodatabase's array of feature tables.
+3. Create `AGSAnnotationLayer`s from the feature tables.
+4. Add the `AGSFeatureLayer`s and `AGSAnnotationLayer`s to the map's operational layers.
+5. Assign the map view's `touchDelegate` property to itself.
+6. Implement `geoView(_:didTapAtScreenPoint:mapPoint:)` to track taps on the map to either select address points or parcel polyline features.  NOTE: Selection is only enabled for points and straight (single segment) polylines.
```
yo1995, line 22:
```suggestion
6. Implement `geoView(_:didTapAtScreenPoint:mapPoint:)` to track taps on the map to either select address points or parcel polyline features. NOTE: Selection is only enabled for points and straight (single segment) polylines.
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
vquach2404: I'm not sure about the deinit, could you advise @philium ?

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
vquach2404: @philium I tested the closure you provided but it didn't work. I found that the desired feature isn't always the first result. So far, I've seen that it could also be the second result but I can't confirm that the desired features are exclusive to only the first and second results.

yo1995: Maybe bad but simple suggestion: revert forEach back to for ... in, to allow return statement cut off the loop.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
philium: What does the documentation say?

yo1995: The first result from a Google search: https://oleb.net/blog/2018/01/notificationcenter-removeobserver/

vquach2404: The documentation says it's not needed for iOS 9.0 and later. Additionally, "If you forget or are unable to remove the observer, the system cleans up the next time it would have posted to it." Seems like it is now personal preference. I see that there are a few samples that do remove the observer. Should I use it in this case?

philium: When seeking to understand the behavior of a method, start with the documentation for the method itself:

https://developer.apple.com/documentation/foundation/notificationcenter/1411723-addobserver

You should find your answer there.

yo1995: Yes the Apple doc seems to contradict themselves a bit here. In addObserver it says we must call remove explicitly, while in remove it says it is optional and handled by system... So what I typically did is to call remove to avoid this ambiguity. 🤥

philium: There's no contradiction. The documentation for `remove(_:)` says the following:

image

And the documentation for `addObserver(forName:object:queue:using:)` says this:

image

So the automatic unregistration does not apply to this method.

yo1995: Aha! I see - I didn't notice the difference between these 2 `addObserver` signatures. Thanks for clarifying that!

vquach2404:
@philium I've made the remaining changes. Please refer to the earlier [comment](https://github.com/Esri/arcgis-runtime-samples-ios/pull/928#discussion_r469608205) about the `forEach` method.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,253 @@
...
+    // A geodatabase to be loaded.
+    var geodatabase: AGSGeodatabase?
+    // The returned cancelable after identifying the layers.
+    var identifyLayers: AGSCancelable?
+    
+    func loadGeodatabase() {
+        // Obtain the geodatabase URL from portal data.
+        let geodatabaseURL = Bundle.main.url(forResource: "loudoun_anno", withExtension: "geodatabase")!
+        geodatabase = AGSGeodatabase(fileURL: geodatabaseURL)
+        guard let geodatabase = geodatabase else { return }
```
philium, line 39: We know that the geodatabase isn't `nil` since it was just created, so it would make more sense to force unwrap than to guard here. But why is the method even dealing with an optional? Instead of assigning the geodatabase to the property and then accessing the property, it would be better to assign the geodatabase to a local, non-optional variable and then use that variable within the method, including to set the property's value. Or perhaps even better, just make the geodatabase property non-optional and immutable by assigning it a value where it is declared.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,253 @@
...
+        // Load a geodatabase from portal data.
+        geodatabase.load { [weak self] (error: Error?) in
```
philium, line 41: The closure is still strongly capturing the geodatabase. It would be better to access it through `self`.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,253 @@
...
+        }
+    }
+    
+    func selectFeature(at: CGPoint) {
+        // Clear any previously selected features.
+        clearSelection()
+        
+        // Identify across all layers.
+        identifyLayers = mapView.identifyLayers(atScreenPoint: at, tolerance: 10.0, returnPopupsOnly: false) { [weak self] (results: [AGSIdentifyLayerResult]?, error: Error?) in
+            guard let self = self, let results = results else { return }
```
philium, line 69: Guarding that `results` is non-nil here means that lines `70` through `74` are useless. Either `results` or `error` will be non-nil, but not both. If results is non-nil, the error is `nil`. Instead, the `if` statement should really be an `if…else if` statement, with one branch for when `results` is non-nil and the other branch for when `error` is non-nil.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,253 @@
...
+            let streetTextField = alert.textFields?[1]
+            // Set the new values for the attributes and update the selected feature.
+            selectedFeature.attributes["AD_ADDRESS"] = Int((addressTextField?.text)!)
+            selectedFeature.attributes["ST_STR_NAM"] = streetTextField?.text
+            selectedFeature.featureTable?.update(selectedFeature)
+
+            if let streetObserver = streetTextDidChangeObserver, let addressObserver = addressTextDidChangeObserver {
+                NotificationCenter.default.removeObserver(streetObserver)
+                NotificationCenter.default.removeObserver(addressObserver)
+                streetTextDidChangeObserver = nil
```
philium, line 119: The observers aren't properties, so there's no need to set them to `nil` here.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,253 @@
...
+                doneAction.isEnabled = (alert.textFields?.allSatisfy { $0.text?.isEmpty == false })!
+            })
+        }
+        // Add a text field to prompt the user to input the street name.
+        alert.addTextField { (textField) in
+            // Prompt a keyboard to enter the street name.
+            textField.keyboardType = .default
+            textField.text = selectedFeature.attributes["ST_STR_NAM"] as? String
+            // Add an observer to ensure the user does not input an empty string.
+            addressTextDidChangeObserver = NotificationCenter.default.addObserver(forName: UITextField.textDidChangeNotification, object: textField, queue: OperationQueue.main, using: {_ in
```
philium, line 142: There's actually no need for a second observer. Just pass `nil` for the object and the same closure will be evaluated when the text of either field changes.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,253 @@
...
+        // Add the source code button item to the right of navigation bar.
+        (self.navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["EditFeaturesWithFeatureLinkedAnnotationViewController"]
+    }
+}
+
+// MARK: - AGSGeoViewTouchDelegate
+// Select the nearest feature or move the point or polyline vertex to the given screen point.
+extension EditFeaturesWithFeatureLinkedAnnotationViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        if let identifyLayers = identifyLayers {
```
philium, line 238: The meaning of `identifyLayers` at the point of use is a little confusing. Based on the name, I would think that maybe it is an array of layers? Perhaps a better name would be `identifyOperation`.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
vquach2404: I think I resolved this by making geodatabase a local variable. Please let me know if that's right.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,242 @@
...
+    var selectedFeature: AGSFeature?
+    // The returned cancelable after identifying the layers.
+    var identifyOperation: AGSCancelable?
+    
+    func loadGeodatabase() {
+        // Obtain the geodatabase URL from portal data.
+        let geodatabaseURL = Bundle.main.url(forResource: "loudoun_anno", withExtension: "geodatabase")!
+        let geodatabase = AGSGeodatabase(fileURL: geodatabaseURL)
+        // Load a geodatabase from portal data.
+        geodatabase.load { [weak self] (error: Error?) in
```
yo1995, line 38:
```suggestion
        geodatabase.load { [weak self] error in
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
philium: The geodatabase was originally a local variable and I [suggested then](https://github.com/Esri/arcgis-runtime-samples-ios/pull/928#discussion_r467189468) that it should be a property:

> Nothing is holding on to the geodatabase. After the geodatabase is loaded, it will be deallocated when the completion handler is done executing. The geodatabase should be an instance property of the view controller.

arcgis-ios-sdk-samples/Content Display Logic/FileTypes.plist
```diff
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>Geodatabases</key>
+	<array>
+		<string>geodatabase</string>
+	</array>
```
yo1995, line 8: The portal item seems to be a zip file, so instead of put it under `Geodatabases` folder, it need to go under `Portal Data` -> `Archives` -> `loudoun_anno.geodatabase` folder.

Due to this issue I cannot build now 😓 waiting for a fix...

vquach2404: Fixed!

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,241 @@
...
+        // Clear any previously selected features.
+        clearSelection()
+        
+        // Identify across all layers.
+        identifyOperation = mapView.identifyLayers(atScreenPoint: at, tolerance: 10.0, returnPopupsOnly: false) { [weak self] (results: [AGSIdentifyLayerResult]?, error: Error?) in
+            guard let self = self else { return }
+            if let error = error {
+                self.clearSelection()
+                self.presentAlert(error: error)
+                return
```
philium, line 69: This return statement is no longer needed.
```suggestion
```

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,241 @@
...
+                let annotationTableNames = ["ParcelLinesAnno_1", "Loudoun_Address_PointsAnno_1"]
+                let annotationTables = annotationTableNames.compactMap { self.geodatabase.geodatabaseAnnotationTable(withTableName: $0) }
+                let annotationLayers = annotationTables.map(AGSAnnotationLayer.init)
+                // Add the annotation layers to the map.
+                map.operationalLayers.addObjects(from: annotationLayers)
+            }
+        }
+    }
+    
+    func selectFeature(at: CGPoint) {
```
philium, line 59: This should be:
```swift
func selectFeature(at point: CGPoint)
```
So that the name of the point variable is not `at`.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,241 @@
...
+        // Add a "Done" option to complete the editing process and close the alert.
+        let doneAction = UIAlertAction(title: "Done", style: .default) { _ in
+            let addressTextField = alert.textFields?[0]
+            let streetTextField = alert.textFields?[1]
+            // Set the new values for the attributes and update the selected feature.
+            selectedFeature.attributes["AD_ADDRESS"] = Int((addressTextField?.text)!)
+            selectedFeature.attributes["ST_STR_NAM"] = streetTextField?.text
+            selectedFeature.featureTable?.update(selectedFeature)
+            // Remove the observer after editing is complete.
+            if let textFieldObserver = textFieldObserver {
```
philium, line 112: The observer also needs to be removed if the cancel button is tapped.

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,241 @@
...
+        }
+        alert.addAction(doneAction)
+        // Add a text field to prompt the user to input the street address.
+        alert.addTextField { (textField) in
+            // Prompt a number pad for the address.
+            textField.keyboardType = .numberPad
+            // Populate the text fields with the current value.
+            textField.text = (selectedFeature.attributes["AD_ADDRESS"] as! NSNumber).stringValue
+            // Add an observer to ensure the user does not input an empty string.
+            textFieldObserver = NotificationCenter.default.addObserver(forName: UITextField.textDidChangeNotification, object: nil, queue: .main, using: {_ in
```
philium, line 124: Since the observer is no longer tied to a specific text field, it no longer needs to be in the configuration handler for this text field. I would move it out of the closure (into the enclosing method, where `textFieldObserver` is declared).

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -0,0 +1,241 @@
...
+        (self.navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = ["EditFeaturesWithFeatureLinkedAnnotationViewController"]
+    }
+}
+
+// MARK: - AGSGeoViewTouchDelegate
+// Select the nearest feature or move the point or polyline vertex to the given screen point.
+extension EditFeaturesWithFeatureLinkedAnnotationViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        if let identifyOperation = identifyOperation {
+            identifyOperation.cancel()
```
philium, line 227:
```suggestion
            identifyOperation.cancel()
            self.identifyOperation = nil
```

yo1995:
@saratchandrakarumuri Just now I gave it a test and it seems to work fine. Are you using the geodatabase file downloaded by the sample viewer? We are not sure what caused this problem. 🤔 

no-issue

saratchandrakarumuri:
@yo1995 I cleared downloaded geodatabase and tried again with fresh install and ran into it again on couple of devices. But the same thing worked on a sim. So looks like the issue is exclusive with physical devices

philium:
@vquach2404 The location of the geodatabase is readonly on device. I didn't think about that. To fix that, you'll need to copy the geodatabase into a temporary directory. [Here](https://github.com/Esri/arcgis-runtime-samples-ios/blob/5670e79874bd2dff3fa6181e1fa218edf4dd10dd/arcgis-ios-sdk-samples/Maps/Apply%20scheduled%20updates%20to%20preplanned%20map%20area/ApplyScheduledUpdatesToPreplannedMapAreaViewController.swift#L36) is how one sample does the same thing with a mobile map package.

yo1995:
@vquach2404 With this change, the edits are not persisted from session to session as the temp replica is removed after deinit. Should we mention in README?

vquach2404:
> @vquach2404 With this change, the edits are not persisted from session to session as the temp replica is removed after deinit. Should we mention in README?

I don't think this is necessary since most samples behave similarly (e.g. a sample refreshing to a default layer after previously selecting different layers) and the geodatabase is not an online source.

PR #929, opened 2020-08-05
Bugfix: Create save KML file deleted too early

yo1995:
This PR fixes the issue in `/runtime/common-samples/issues/2150`.

In the original implementation, when the save file `UIActivityViewController` is dismissed or completed, the KML file get deleted, without considering the case where user could cancel the process. This will cause the activity VC keep popping up when tapping on save files button again, as it cannot find the file with provided path.

### DIFF
--- a/arcgis-ios-sdk-samples/Edit data/Create and save KML file/CreateAndSaveKMLViewController.swift
+++ b/arcgis-ios-sdk-samples/Edit data/Create and save KML file/CreateAndSaveKMLViewController.swift
@@ -61,8 +61,12 @@ class CreateAndSaveKMLViewController: UIViewController {
         let activityViewController = UIActivityViewController(activityItems: [kmzProvider], applicationActivities: nil)
         activityViewController.popoverPresentationController?.barButtonItem = sender
         present(activityViewController, animated: true)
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Edit data/Create and save KML file/CreateAndSaveKMLViewController.swift
```diff
@@ -61,8 +61,12 @@ class CreateAndSaveKMLViewController: UIViewController {
         let activityViewController = UIActivityViewController(activityItems: [kmzProvider], applicationActivities: nil)
         activityViewController.popoverPresentationController?.barButtonItem = sender
         present(activityViewController, animated: true)
-        activityViewController.completionWithItemsHandler = { (_, _, _, _) in
-            kmzProvider.deleteKMZ()
+        activityViewController.completionWithItemsHandler = { activityType, completed, returnedItems, activityError in
```
philium, line 64: The unused values should still be represented by an underscore.
```suggestion
        activityViewController.completionWithItemsHandler = { _, completed, _, activityError in
```

yo1995:
@saratchandrakarumuri Could you help to verify that this PR addresses the bug? Thanks. 🙂

PR #930, opened 2020-08-06
Fix: Animate images with image overlay slider color

yo1995:
Description: The opacity slider is embedded in a `UIBarButtonItem`. Current version sets slider's `isEnabled` property to false in storyboard, which dims its tint color to indicate it is not interactive. After enabling it in code, the color does not update, which might give wrong clues to the user that the slider is not operable.

This fix invert the settings: enable in storyboard, and disable in code if loading images failed.

Please see the comparison images below.
|Before|After|
|-|-|
|before|after|

ref: #857


yo1995:
> Does the slider dim when disabled in code?

Literally speaking I don't know... 🤥  I'll do some research before merging this PR.

yo1995:
> Does the slider dim when disabled in code?

yes - setting `isEnabled` in code either to true or false both update the tint color. 

Just have no idea why the original false to true change doesn't update its color. 😵

philium:
> Just have no idea why the previous false to true change doesn't update its color. 😵

I have no idea either. Thanks for confirming that.

PR #940, opened 2020-08-18
Update OpenSceneViewController.swift

TADraeseke:
As per https://devtopia.esri.com/runtime/common-samples/issues/2164


yo1995:
> One thing to note - this scene was crashing on some iOS devices. I have seen crashes on iOS and Android due to either a memory leak or just high memory usage. It is probably related to the number of unique textures on the building. I'd suggest we  test navigating around the scene for a few minutes to see if any crashes are easy to reproduce. If so, it might not be a good scene to showcase

yo1995:
The update looks good, though I've a question - this sample does not add the scene with an elevation base surface, thus the buildings in the distance are floating above ground. Is that a problem? @TADraeseke 

floating

PR #941, opened 2020-08-21
Update: Open scene (portal item) with new scene

yo1995:
The previous scene has rendering problem, so we go with a new Berlin, Germany scene for this sample.

Also clean up unneeded code.


philium:
@yo1995 It looks like the log messages can be [safely ignored](https://developer.apple.com/forums/thread/71915?answerId=209510022#209510022).

yo1995:
Thanks Phil! 

@saratchandrakarumuri please take a look at these updates, as they change the look of the sample.

PR #943, opened 2020-08-26
Update: LaunchScreen.storyboard

yo1995:
This PR updates the previous `LaunchScreen.xib` to `LaunchScreen.storyboard`. Per guidelines from [Apple](https://developer.apple.com/ios/submit/#:~:text=Test%20your%20app%20on%20devices)

> Starting June 30, 2020, app updates must use an Xcode storyboard to provide the app’s launch screen. In addition, new apps that run on iPhone must support all iPhone screens and new apps that run on iPad must support all iPad screens.

It also removes the previous layout guides from xib.

This should go after the master release tomorrow.


philium:
@yo1995 It seems like these changes should be part of (or at least branched off of) https://github.com/Esri/arcgis-runtime-samples-ios/pull/915, since that PR modifies the launch screen and introduces a new color (`accentColor`) that needs to be used in the new storyboard. What do you think? I'm fine with replacing the launch nib with a storyboard in that PR instead of modifying the launch nib just to replace it right away.

yo1995:
> @yo1995 It seems like these changes should be part of (or at least branched off of) #915, since that PR modifies the launch screen and introduces a new color (`accentColor`) that needs to be used in the new storyboard. What do you think? I'm fine with replacing the launch nib with a storyboard in that PR instead of modifying the launch nib just to replace it right away.

I see. I was afraid of #915 cannot make our AppStore release so I made this a separate PR, so that we can pass the new submission requirements from Apple even if 915 need more time. 

If we gonna wait after #915 getting merged to release the app, then I'll do this `LaunchScreen.storyboard` change there.

yo1995:
Hmm I'll close this PR and make the changes in #915

PR #945, opened 2020-08-27
Fix: Merge conflict between v.next and master

yo1995:
This PR fixes the merge conflict between `v.next` and `master`, and also update toolkit to 100.9 master.

PR #947, opened 2020-08-27
Update: re-update toolkit version to 100.9

yo1995:
Due to wrong order during merging #945 and #946 , the submodule update to 100.9 is accidentally reverted.

Thus we need to re-update the submodule to 100.9 both on `v.next` and `master`

After this PR we'll need another PR to merge `v.next` to `master`.


yo1995:
Sorry for the miscue 😓  - please re-review so that we have the correct submodule version.

PR #949, opened 2020-08-28
Update group layers sample

vquach2404:
Group layers sample is updated to showcase the new visibility modes.


arcgis-ios-sdk-samples/Layers/Group layers/README.md
```diff
@@ -10,13 +10,17 @@ Group layers communicate to the user that layers are related and can be managed
...
 
 ## How it works
 
 1. Create an `AGSGroupLayer` object.
 2. Add a child layer to the group layer's layers collection.
-3. To toggle the visibility of the group, simply change the group layer's `isVisible` property.
+3. Set the group layer's `groupVisibilityMode` to change its behavior:
+    * `groupVisibilityMode.INDEPENDENT` allows each sublayer to change its visibility independently.
+    * `groupVisibilityMode.EXCLUSIVE` allows only one sublayer to be visible at a time.
+    * `groupVisibilityMode.INHERITED` treats the group layer as if it is one merged layer.
```
yo1995, line 22: I also forget how to we annotate the enum cases.

Was it wrapping cases in backticks or bare strings? i.e. `exclusive` vs exclusive vs `.exclusive`

Wait for Phil's comment.

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
```diff
@@ -97,32 +127,49 @@ class LayersTableViewController: UITableViewController, GroupLayersCellDelegate,
...
+            visibleRows.lazy.filter { $0.section == section }.forEach { indexPath in
+                if let cell = tableView.cellForRow(at: indexPath) {
+                    cell.isUserInteractionEnabled = isEnabled
+                    cell.textLabel?.isEnabled = isEnabled
+                    // Change the tint if the cell is enabled.
+                    if isEnabled {
+                        cell.tintColor = .none
+                    } else {
+                        cell.tintColor = .gray
+                    }
```
yo1995, line 150: With `isUserInteractionEnabled` set, I believe the tint color are managed automatically. Please try to comment out and see if it behaves as expected. 😳

arcgis-ios-sdk-samples/Layers/Group layers/GroupLayersSectionView.swift
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Esri.
+// Copyright 2020 Esri.
```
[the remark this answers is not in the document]
vquach2404: I noticed Hudson update it in Java so I thought we might want to do the same. But I'll wait for Phil's thoughts.

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
vquach2404: I had to include this specifically because the accessory checkmarks for those cells didn't change colors despite changing `isUserInteractionEnabled`.

yo1995: Good point!

philium, line 150: This sort of thing should be done by updating the value of `tintAdjustmentMode`, not by updating the tint color itself. Search the repo for `tintAdjustmentMode` to find an example.

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
```diff
@@ -35,23 +35,43 @@ class LayersTableViewController: UITableViewController, GroupLayersCellDelegate,
...
+            cell.layerVisibilitySwitch.isOn = childLayer.isVisible
+            cell.layerVisibilitySwitch.isEnabled = groupLayer.isVisible
+            
+            // To update the visibility of operational layers on switch toggle.
+            cell.delegate = self
+            
+            return cell
+        case 1:
+            let cell = tableView.dequeueReusableCell(withIdentifier: "exclusiveCell", for: indexPath) as! GroupLayersCell
+            guard let groupLayer = layers[indexPath.section] as? AGSGroupLayer, let childLayers = groupLayer.layers as? [AGSLayer] else { return cell }
```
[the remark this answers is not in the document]
vquach2404: I was slightly confused by this comment but I think I've resolved this issue 😅 Please let me know if further action is needed!

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
```diff
@@ -35,23 +44,52 @@ class LayersTableViewController: UITableViewController, GroupLayersCellDelegate,
...
+            cell.textLabel?.text = formattedValue(of: childLayer.name)
+            // Enable or disable the cell accordingly.
+            cell.isUserInteractionEnabled = groupLayer.isVisible
+            cell.textLabel?.isEnabled = groupLayer.isVisible
+            // Adjust the tint if the cell is enabled.
+            if groupLayer.isVisible {
+                cell.tintAdjustmentMode = .automatic
+            } else {
+                cell.tintAdjustmentMode = .dimmed
+            }
```
philium, line 82: This is correct, but you still need to set the tint color of the cell to the tint color of the view. Setting the `tintAdjustmentMode` means that you don't need to set a different tint color depending on whether the group layer is visible. With the adjustment mode as dimmed, all controls in the cell will have the correct appearance when disabled.

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
```diff
@@ -73,6 +111,18 @@ class LayersTableViewController: UITableViewController, GroupLayersCellDelegate,
...
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let groupLayer = layers[indexPath.section] as! AGSGroupLayer
+        let childLayers = groupLayer.layers as! [AGSLayer]
+        
+        var indexPathsToReload = [indexPath]
+        if let indexOfPreviouslyVisibleLayer = childLayers.firstIndex(where: { $0.isVisible }) {
+            indexPathsToReload.append(IndexPath(row: indexOfPreviouslyVisibleLayer, section: indexPath.section))
+        }
+        childLayers[indexPath.row].isVisible = true
+        tableView.reloadRows(at: indexPathsToReload, with: .automatic)
```
philium, line 123: I had forgot that reloading parts of a table view doesn't play nicely with observing the content size, for some reason. My first idea was that the observation could be removed as soon as a change was observed, but the first observed change isn't the correct size, so that wouldn't work. My other thought would be to have a custom `func reloadRows(at indexPaths: [IndexPath])` function that first removes the content size observation (if it still exists) and then calls `reloadRows` on the table view. What do you think of that idea?

arcgis-ios-sdk-samples/Layers/Group layers/README.md
```diff
@@ -10,13 +10,17 @@ Group layers communicate to the user that layers are related and can be managed
 
 ## How to use the sample
 
-The layers in the map will be displayed in a table of contents. Toggle the switch next to a layer's name to change its visibility. Turning a group layer's visibility off will override the visibility of its child layers.
+The layers in the map will be displayed in a table of contents. Toggle the switch next to a layer's name or tap the cell to change its visibility. Turning a group layer's visibility off will override the visibility of its child layers.
```
philium, line 13: > Toggle the switch next to a layer's name or tap the cell to change its visibility.

That kind of makes it sound like there's two different ways to change a layer's visibility. You might consider changing it to something like this: "For project area group layers, toggle the switch next to the layers name to change its visibility. For buildings group layers, tap a row to make that layer the visible layer in that group."

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
vquach2404: I think that's a good idea. I gave it a try, please let me know what you think!

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
```diff
@@ -87,46 +144,38 @@ class LayersTableViewController: UITableViewController, GroupLayersCellDelegate,
     func didToggleSwitch(_ sectionView: GroupLayersSectionView, isOn: Bool) {
         guard let section = tableView.section(forHeaderView: sectionView) else { return }
         layers[section].isVisible = isOn
-        updateSwitchForRows(in: section, isEnabled: isOn)
+        let rowsInSection = tableView.numberOfRows(inSection: section)
+        let rowArray = 0...rowsInSection - 1
+        let indexPaths = rowArray.map { IndexPath(row: $0, section: section) }
+        reloadRows(at: indexPaths)
```
philium, line 150: I think the table view's method to reload a section is more efficient than reloading all the rows in the section. It would probably be better to add a custom function to reload a section, just like you did for reloading rows. Could you please do that?

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
```diff
@@ -16,42 +16,86 @@ import UIKit
...
-        // To update the visibility of operational layers on switch toggle.
-        cell.delegate = self
-        
-        return cell
+        let groupLayer = layers[indexPath.section]
+        switch groupLayer.visibilityMode {
+        case .independent:
+            let cell = tableView.dequeueReusableCell(withIdentifier: "switchCell", for: indexPath) as! GroupLayersCell
+            let childLayers = groupLayer.layers as! [AGSLayer]
+            let childLayer = childLayers[indexPath.row]
```
philium, line 57: It also looks like this can be moved outside of the switch statement since it is the same for all cases.

arcgis-ios-sdk-samples/Layers/Group layers/LayersTableViewController.swift
vquach2404: Both of these functions are only used once -- would it be better to just remove the observer right before calling `reloadRows` in the original function rather than make two separate functions?

philium: I was under the impression that they were called in multiple places. Since they're not, then I suppose the custom functions aren't needed.

philium (approved):
Thank you so much for implementing this sample!

yo1995 (approved):
LGTM 👍 Pretty smooth.  Do we need Sarat's review for this update?

saratchandrakarumuri (approved):
Changes looks good

PR #951, opened 2020-08-31
Fix selection in OptionsTableViewController

philium:
Before | After
------ | -----
Before | After

I've never been happy with how selection was implemented in `OptionsTableViewController`. The tapped cells are not highlighted as excepted and the selection change isn't animated. This PR addresses those issues by simply reloading the previously selected row and the newly selected row. Rows are now highlighted when tapped and both the highlight and the checkmark changes are animated.

### DIFF
--- a/arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
+++ b/arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
@@ -50,7 +50,7 @@ class OptionsTableViewController: UITableViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        tableView.register(OptionCell.self, forCellReuseIdentifier: "OptionCell")
+        tableView.register(UITableViewCell.self, forCellReuseIdentifier: "OptionCell")
     }
     
     // UITableViewDataSource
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
```diff
@@ -50,7 +50,7 @@ class OptionsTableViewController: UITableViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        tableView.register(OptionCell.self, forCellReuseIdentifier: "OptionCell")
+        tableView.register(UITableViewCell.self, forCellReuseIdentifier: "OptionCell")
```
yo1995, line 53: With this change, should we also remove the `OptionCell` identifier from storyboard? I only find one from `CreateSaveMap.storyboard` and not sure if it should be removed.

philium: No, It should not be removed from `CreateSaveMap.storyboard`. If it were removed, it would [break the sample](https://github.com/Esri/arcgis-runtime-samples-ios/blob/91459c9e25126e34eb52866985a83dca37a69759/arcgis-ios-sdk-samples/Maps/Create%20and%20save%20a%20map/CreateOptionsViewController.swift#L52). Reuse identifiers are not shared across storyboards.

yo1995: I see! Thanks for explanation. 👍

PR #952, opened 2020-09-02
Update: AGSRouteTrackerLocationDataSource in Navigate route

yo1995:
This PR updates the simulated location datasource to an `AGSRouteTrackerLocationDataSource`, based on requests in `/common-samples/issues/2222` and `/common-samples/issues/2045`.

Also refactor the code to decouple functions and make it cleaner.


arcgis-ios-sdk-samples/Route and directions/Navigate route/README.md
```diff
@@ -20,11 +20,12 @@ Tap "Navigate" to simulate traveling and to receive directions from a preset sta
...
+7. Create an `AGSRouteTrackerLocationDataSource` with the route tracker and an `AGSSimulatedLocationDataSource` object to snap the location display to the route.
+8. Use `AGSLocationChangeHandlerDelegate.locationDataSource(_:locationDidChange:)` to track the location of the device and update the route tracking status.
+9. Use `AGSRouteTrackerDelegate.routeTracker(_:didUpdate:)` to be notified of `AGSTrackingStatus` changes, and use them to display updated route information. `AGSTrackingStatus` includes a variety of information on the route progress, such as the remaining distance, remaining geometry or traversed geometry (represented by an `AGSPolyline`), or the remaining time (`Double`), amongst others.
+10. Use `AGSRouteTrackerDelegate.routeTracker(_:didGenerateNewVoiceGuidance:)` to be notified of new voice guidances. From the voice guidance, get the `AGSVoiceGuidance.text` representing the directions and use a text-to-speech engine to output the maneuver directions.
+11. You can also query the tracking status for the current `AGSDirectionManeuver` index, retrieve that maneuver from the `AGSRoute` and get its direction text to display in the GUI.
```
vquach2404, line 27:
```suggestion
11. You can also query the tracking status for the current `AGSDirectionManeuver` index, retrieve that maneuver from the `AGSRoute`, and get its direction text to display in the GUI.
```

arcgis-ios-sdk-samples/Route and directions/Navigate route/README.md
```diff
@@ -20,11 +20,12 @@ Tap "Navigate" to simulate traveling and to receive directions from a preset sta
...
+12. To establish whether the destination has been reached, get the `destinationStatus` from the tracking status. If the destination status is `.reached`, and the `remainingDestinationCount` is 1, we have arrived at the destination and can stop routing. If there are several destinations in your route, and the remaining destination count is greater than 1, switch the route tracker to the next destination.
```
vquach2404, line 28:
```suggestion
12. To establish whether the destination has been reached, get the `destinationStatus` from the tracking status. If the destination status is `.reached` and the `remainingDestinationCount` is 1, we have arrived at the destination and can stop routing. If there are several destinations on your route and the remaining destination count is greater than 1, switch the route tracker to the next destination.
```

vquach2404 (commented):
Please add a comma after `returnStops` on line 3 of How it works. I can't make a suggestion since no edits were made there.

Also, since you're updating this sample, could you add some padding to the left of the directions label? I think the words are too close to the edge of the screen and adding padding would improve the look.

arcgis-ios-sdk-samples/Route and directions/Navigate route/README.md
yo1995: I've fixed the issue in README and replace the screenshot with text label content inset added.

arcgis-ios-sdk-samples/Route and directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -19,51 +19,57 @@ import ArcGIS
...
-    /// A list to keep track of directions solved by the route task.
-    var directionsList: [AGSDirectionManeuver] = []
     /// The route tracker for navigation. Use delegate methods to update tracking status.
     var routeTracker: AGSRouteTracker!
-    /// The route task to solve the route between stops, using the online routing service.
-    var routeTask: AGSRouteTask!
+    /// A list to keep track of directions solved by the route task.
+    var directionsList: [AGSDirectionManeuver] = []
+    
+    /// The original view point that can be reset to later on.
```
vquach2404, line 55:
```suggestion
    /// The original view point that can be reset later on.
```

vquach2404 (approved):
Just one more small fix. Ready for Phil's review!

arcgis-ios-sdk-samples/Route and directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -19,51 +19,57 @@ import ArcGIS
...
+    @IBOutlet var resetBarButtonItem: UIBarButtonItem!
+    /// The button to recenter the map to navigation pan mode.
+    @IBOutlet var recenterBarButtonItem: UIBarButtonItem!
+    /// The map view managed by the view controller.
+    @IBOutlet var mapView: AGSMapView! {
+        didSet {
+            mapView.map = AGSMap(basemap: .navigationVector())
+            let stopSymbol = AGSSimpleMarkerSymbol(style: .diamond, color: .orange, size: 20)
+            let stopGraphics = makeStops().map { AGSGraphic(geometry: $0.geometry, symbol: stopSymbol) }
+            let routeGraphics = [routeAheadGraphic, routeTraveledGraphic]
```
philium, line 38: Could all of this go into `makeRouteOverlay()`? Then the last line of this property observer could just be `mapView.graphicsOverlays.add(makeRouteOverlay())`.

arcgis-ios-sdk-samples/Route and directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -159,6 +154,33 @@ class NavigateRouteViewController: UIViewController {
...
+        let mockDataSource = makeDataSource(route: firstRoute)
+        initialLocation = mockDataSource.locations?.first
+        // Create a route tracker location data source to snap the location display to the route.
+        let routeTrackerLocationDataSource = AGSRouteTrackerLocationDataSource(routeTracker: routeTracker, locationDataSource: mockDataSource)
+        
+        // Set location display.
+        mapView.locationDisplay.dataSource = routeTrackerLocationDataSource
+        mapView.locationDisplay.autoPanMode = .navigation
+        // If the user navigates the map view away from the location display, activate the recenter button.
+        mapView.locationDisplay.autoPanModeChangedHandler = { [weak self] _ in self?.recenterBarButtonItem.isEnabled = true }
```
philium, line 176: Instead of setting `recenterBarButtonItem.isEnabled` potentially multiple times, could you remove the handler when it has been called and then re-add it when the recenter button is tapped? Just a thought.

arcgis-ios-sdk-samples/Route and directions/Navigate route/README.md
```diff
@@ -16,15 +16,16 @@ Tap "Navigate" to simulate traveling and to receive directions from a preset sta
...
+10. Use `AGSRouteTrackerDelegate.routeTracker(_:didGenerateNewVoiceGuidance:)` to be notified of new voice guidances. From the voice guidance, get the `AGSVoiceGuidance.text` representing the directions and use a text-to-speech engine to output the maneuver directions.
+11. You can also query the tracking status for the current `AGSDirectionManeuver` index, retrieve that maneuver from the `AGSRoute`, and get its direction text to display in the GUI.
+12. To establish whether the destination has been reached, get the `destinationStatus` from the tracking status. If the destination status is `.reached` and the `remainingDestinationCount` is 1, we have arrived at the destination and can stop routing. If there are several destinations on your route and the remaining destination count is greater than 1, switch the route tracker to the next destination.
```
philium, line 28:
```suggestion
12. To establish whether the destination has been reached, get the `destinationStatus` from the tracking status. If the destination status is `reached` and the `remainingDestinationCount` is 1, we have arrived at the destination and can stop routing. If there are several destinations on your route and the remaining destination count is greater than 1, switch the route tracker to the next destination.
```

philium: Also, I'm not sure about that usage of "we". Everywhere else it is "you".

arcgis-ios-sdk-samples/Route and directions/Navigate route/NavigateRouteViewController.swift
yo1995: I'm not sure if I understood it correctly. 😳  Please refer to the changes on 

https://github.com/Esri/arcgis-runtime-samples-ios/blob/cc37331b0e0e52cc7140e559b053eff3a4871adb/arcgis-ios-sdk-samples/Route%20and%20directions/Navigate%20route/NavigateRouteViewController.swift#L174-L177

and

https://github.com/Esri/arcgis-runtime-samples-ios/blob/cc37331b0e0e52cc7140e559b053eff3a4871adb/arcgis-ios-sdk-samples/Route%20and%20directions/Navigate%20route/NavigateRouteViewController.swift#L226-L229

arcgis-ios-sdk-samples/Route and directions/Navigate route/NavigateRouteViewController.swift
yo1995: Yes sure. Recently I'm trying with better coding style and had a question: **when should I use instance properties in a function?**

My current understanding is to only access instance properties in a function that does not take arguments and/or does not return. e.g. `func loadGeodatabase()` or `func makeMap()` or `func someAsync(completion: @escaping () -> Void)`.

Otherwise I tend to create a "pure function" that does not deal with instance properties. (to achieve that, sometimes I pass instance property as an argument to a pure function).

Is my understanding correct? Also @vquach2404 as your knowledge in programming should be deeper than mine.

Much appreciated! 😄

vquach2404: I tried to find some information in the [Swift doc](https://docs.swift.org/swift-book/LanguageGuide/Properties.html#:~:text=In%20addition%20to%20properties%2C%20you,property%20is%20not%20accessed%20directly) but it didn't really provide much of an answer. I would suggest reaching out to Mikey from BNR in addition to taking Phil's advice. 🙂

yo1995: > I would suggest reaching out to Mikey from BNR in addition to taking Phil's advice. 

Yes that is a good idea! Thanks for the heads up.

philium: There is no such rule that I know of for when to make a function pure and when not to. Pure functions are nice in some ways, but can lead to less straightforward code, which isn't necessarily good for sample code where the most important thing is to show users how they can use the ArcIGS runtime in their apps.

arcgis-ios-sdk-samples/Route and directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -159,6 +152,36 @@ class NavigateRouteViewController: UIViewController {
...
+        let routeTrackerLocationDataSource = AGSRouteTrackerLocationDataSource(routeTracker: routeTracker, locationDataSource: mockDataSource)
+        
+        // Set location display.
+        mapView.locationDisplay.dataSource = routeTrackerLocationDataSource
+        mapView.locationDisplay.autoPanMode = .navigation
+        // If the user navigates the map view away from the location display, activate the recenter button.
+        mapView.locationDisplay.autoPanModeChangedHandler = { [weak self] _ in
+            self?.recenterBarButtonItem.isEnabled = true
+            self?.mapView.locationDisplay.autoPanModeChangedHandler = nil
+        }
```
philium, line 177: Could this be replaced by a call to `recenter()`? It seems like that method is doing most of this already and it would be good not to do the same thing in multiple places.

arcgis-ios-sdk-samples/Route and directions/Navigate route/NavigateRouteViewController.swift
```diff
@@ -159,6 +152,36 @@ class NavigateRouteViewController: UIViewController {
...
+        let mockDataSource = makeDataSource(route: firstRoute)
+        initialLocation = mockDataSource.locations?.first
+        // Create a route tracker location data source to snap the location display to the route.
+        let routeTrackerLocationDataSource = AGSRouteTrackerLocationDataSource(routeTracker: routeTracker, locationDataSource: mockDataSource)
+        
+        // Set location display.
+        mapView.locationDisplay.dataSource = routeTrackerLocationDataSource
+        mapView.locationDisplay.autoPanMode = .navigation
+        // If the user navigates the map view away from the location display, activate the recenter button.
+        mapView.locationDisplay.autoPanModeChangedHandler = { [weak self] _ in
```
philium, line 174: The documentation for this property states the following:

> The block will be invoked on the same thread on which the event occured, which could be any arbitrary thread. You need to dispatch any UI related work to the main thread.

Please update accordingly. Also, take care to repeatedly access a `weak` variable without unwrapping it first.

yo1995: I made some changes but not sure if my understanding is correct. Please take a look! Thanks.

philium (approved):

yo1995:
Since it does not have visual changes to the sample, I'll not ask Sarat to test again. @vquach2404 please take another look. Thanks 😄

vquach2404 (approved):

PR #958, opened 2020-09-04
Renamed RGB renderer

vquach2404:
Renamed class, files, metadata, and functions.


arcgis-ios-sdk-samples/Layers/RGB renderer/README.metadata.json
```diff
@@ -37,7 +37,7 @@
     ],
     "snippets": [
         "RGB Renderer Cells.swift",
```
yo1995, line 39: 😩 Why they used filename with spaces in the first place...

yo1995 (approved):

arcgis-ios-sdk-samples/Layers/RGB renderer/README.metadata.json
vquach2404: Haha I also wondered if I should rename that file. @philium, what do you think?

philium: Yes please! In fact, you could just move the classes into `RGBRendererSettingsViewController.swift`. There's not really a reason for them to be in a separate file.

arcgis-ios-sdk-samples/Layers/RGB renderer/README.metadata.json
```diff
@@ -37,7 +37,7 @@
...
-        "RGBRendererSettingsVC.swift",
+        "RGBRendererSettingsViewController.swift",
         "RGBRendererViewController.swift"
```
philium, line 41: This file should really come first since it is the main view of the sample. Do you mind reordering these?

arcgis-ios-sdk-samples/Layers/RGB renderer/RGBRendererViewController.swift
```diff
@@ -42,7 +42,7 @@ class RGBRendererViewController: UIViewController, RGBRendererSettingsVCDelegate
     
     // MARK: - RGBRendererSettingsVCDelegate
```
philium, line 43: This needs to be changed.

arcgis-ios-sdk-samples/Layers/RGB renderer/RGBRendererViewController.swift
```diff
@@ -42,7 +42,7 @@ class RGBRendererViewController: UIViewController, RGBRendererSettingsVCDelegate
...
     
-    func rgbRendererSettingsVC(_ rgbRendererSettingsVC: RGBRendererSettingsVC, didSelectStretchParameters parameters: AGSStretchParameters) {
+    func rgbRendererSettingsViewController(_ rgbRendererSettingsVC: RGBRendererSettingsViewController, didSelectStretchParameters parameters: AGSStretchParameters) {
```
philium, line 45: The name of the first parameter has changed in the protocol, so that should be reflected here.

philium (changes requested):

arcgis-ios-sdk-samples/Layers/RGB renderer/README.metadata.json
yo1995: Ahh good catch. In previous README and metadata updates, we ran a script to sort the filenames by alphabetical order.

It makes sense to put the most important file atop so that users see it first on the dev website.

I can remove this ordering rule from the metadata style checker. @vquach2404 what do you think?

vquach2404: I agree, the first file should always be the main view and anything that comes afterward should be alphabetically ordered.

philium: I personally don't think the files should be sorted automatically and I regret that they were. I somehow missed that. The file order is dependent on each sample and is often used to show the flow through a sample. For example, the metadata file for the IWA sample [originally](https://github.com/ChrisFrenchPDX/arcgis-runtime-samples-ios/blob/8176b41c91e64bdce0a29459069ecf600da933ed/arcgis-ios-sdk-samples/Cloud%20%26%20Portal/Integrated%20Windows%20Authentication/README.metadata.json#L23) had the files listed by the order in which they are used in the sample. The first presented view controller was first, the second listed second, and so on.

vquach2404: Hm, we could always sort through all the samples with multiple swift files and revert the snippets portion. It may be a lower priority issue but we can make it a rule to order the list according to order of use from now on.

yo1995: Sounds good to me!

philium (approved):

yo1995 (approved):

PR #964, opened 2020-09-10
Update Edit and sync features

vquach2404:
This PR resolves some issues mentioned in #1707. 

The extent view will remain visible through the editing process and panning, zooming, and rotating will be turned off after the user has generated the geodatabase. This will prevent the user from straying outside the generated geodatabase and help them keep track of where they can apply edits. 

Additionally, edits will only be allowed on the points feature layer. This removes the polygons and lines that caused issues when selected.


arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -85,15 +85,16 @@ class EditAndSyncFeaturesViewController: UIViewController {
             if let error = error {
                 self.presentAlert(error: error)
             } else {
-                if let featureServiceInfo = self.geodatabaseSyncTask?.featureServiceInfo,
-                    let map = self.mapView.map {
-                    for index in featureServiceInfo.layerInfos.indices.reversed() {
```
yo1995, line 90: Does it make a difference to remove `reversed()`?

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -85,15 +85,16 @@ class EditAndSyncFeaturesViewController: UIViewController {
...
+                    let featureTable = AGSServiceFeatureTable(url: layerURL)
+                    featureTable.load { error in
+                        if let error = error {
+                            self.presentAlert(error: error)
+                        } else if featureTable.geometryType == .point {
                             let featureLayer = AGSFeatureLayer(featureTable: featureTable)
-                            featureLayer.name = layerInfo.name
                             map.operationalLayers.add(featureLayer)
                         }
                     }
```
yo1995, line 100: It might not need to explicitly call load here, but I'm not sure if the `geometryType` depends on loading status. Wait for Phil's comments.

yo1995 (commented):
Thanks for making this change. It seems the current README doesn't need update, while the screenshot might need an update. A few comments below.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
vquach2404: It doesn't make a difference since we're only showing the point feature layer. Previously, it prevented the polygon layer from being added on top of the point layer which caused issues when trying to select a point.

yo1995:
LGTM 👍  Feel free to add Phil to take a look.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -85,15 +85,16 @@ class EditAndSyncFeaturesViewController: UIViewController {
...
-                        let layerInfo = featureServiceInfo.layerInfos[index]
-                        // For each layer in the serice, add a layer to the map.
-                        if let layerURL = self.featureServiceURL?.appendingPathComponent(String(index)) {
-                            let featureTable = AGSServiceFeatureTable(url: layerURL)
+                guard let featureServiceInfo = self.geodatabaseSyncTask?.featureServiceInfo, let map = self.mapView.map else { return }
+                for index in featureServiceInfo.layerInfos.indices {
+                    // For each layer in the serice, add a layer to the map.
+                    guard let layerURL = self.featureServiceURL?.appendingPathComponent(String(index)) else { return }
+                    let featureTable = AGSServiceFeatureTable(url: layerURL)
+                    featureTable.load { error in
```
philium, line 93: This closure is capturing self strongly. Also, a loadable holds on to the completion handler, so the handler should store an unowned reference to the loadable to avoid a retain cycle.

However, does the order of the layers matter? There's no guarantee that the tables will load in the order that `load()` is called. If order matters, use `AGSLoadObjects` to load all of the tables before creating the feature layers.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
philium: That's a good question. @vquach2404 Does the sample function as expected without explicitly loading each table?

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -141,6 +142,9 @@ class EditAndSyncFeaturesViewController: UIViewController {
             self.barButtonItem.isEnabled = false
             self.instructionsLabel.text = "Tap on a feature"
             self.mapView.touchDelegate = self
+            self.mapView.interactionOptions.isPanEnabled = false
+            self.mapView.interactionOptions.isZoomEnabled = false
+            self.mapView.interactionOptions.isRotateEnabled = false
```
philium, line 147: Rather than disabling each of these interactions individually, could user interaction be disabled on the map view itself?

vquach2404: No, user interaction must be enabled so that the user can select and move features.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
vquach2404: I played around with it, trying to load only the point layer. It seems like it's unable to determine the `geometryType` of the layer unless it's loaded.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
vquach2404: Originally, the order of the layers did matter but regardless, only the point layer is added now. But it seems like I have to load all the layers in order to determine which is the point layer.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -85,15 +85,18 @@ class EditAndSyncFeaturesViewController: UIViewController {
...
                 self.presentAlert(error: error)
             } else {
-                if let featureServiceInfo = self.geodatabaseSyncTask?.featureServiceInfo,
-                    let map = self.mapView.map {
-                    for index in featureServiceInfo.layerInfos.indices.reversed() {
-                        let layerInfo = featureServiceInfo.layerInfos[index]
-                        // For each layer in the serice, add a layer to the map.
-                        if let layerURL = self.featureServiceURL?.appendingPathComponent(String(index)) {
-                            let featureTable = AGSServiceFeatureTable(url: layerURL)
+                let featureServiceInfo = (self.geodatabaseSyncTask?.featureServiceInfo)!
```
yo1995, line 88:
```suggestion
                let featureServiceInfo = self.geodatabaseSyncTask.featureServiceInfo!
```
Since we are unwrapping anyway, only *bang* ❗  it once.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -85,15 +85,18 @@ class EditAndSyncFeaturesViewController: UIViewController {
...
+                let map = self.mapView.map!
+                for index in featureServiceInfo.layerInfos.indices {
+                    // For each layer in the serice, add a layer to the map.
+                    let layerURL = (self.featureServiceURL?.appendingPathComponent(String(index)))!
```
yo1995, line 92:
```suggestion
                    let layerURL = self.featureServiceURL!.appendingPathComponent(String(index))
```
Same as above.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -85,15 +85,18 @@ class EditAndSyncFeaturesViewController: UIViewController {
...
+                    let featureTable = AGSServiceFeatureTable(url: layerURL)
+                    _ = featureTable.load { [weak self] error in
```
yo1995, line 94: It seems this function has a `Void` return value, rather than an `AGSCancelable` so I wonder if we can safely remove the underscore.

yo1995 (approved):
A few other suggestions. Changes look good. 👍

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
vquach2404: I did so because of Phil's[ comment](https://github.com/Esri/arcgis-runtime-samples-ios/pull/964#discussion_r487309775). Did I misunderstand?

yo1995: I think it is safe to change to `featureTable.load { [weak self] error in ...`, as the return value doesn't have any effect on strong reference or have any use. Wait for Phil's comment.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -85,15 +85,18 @@ class EditAndSyncFeaturesViewController: UIViewController {
...
-                    for index in featureServiceInfo.layerInfos.indices.reversed() {
-                        let layerInfo = featureServiceInfo.layerInfos[index]
-                        // For each layer in the serice, add a layer to the map.
-                        if let layerURL = self.featureServiceURL?.appendingPathComponent(String(index)) {
-                            let featureTable = AGSServiceFeatureTable(url: layerURL)
+                let featureServiceInfo = self.geodatabaseSyncTask.featureServiceInfo!
+                let map = self.mapView.map!
+                for index in featureServiceInfo.layerInfos.indices {
+                    // For each layer in the serice, add a layer to the map.
+                    let layerURL = self.featureServiceURL!.appendingPathComponent(String(index))
```
philium, line 92: I was suggesting that `featureServiceURL` be made non-optional by force unwrapping it at the point of initialization:
```swift
private let featureServiceURL = URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/Sync/WildfireSync/FeatureServer")!
```

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
philium: Since `load(completion:)` returns `Void`, it really doesn't make sense to ignore the return value. What I was suggesting is that the feature table be made an unowned reference _inside the closure_:
```swift
featureTable.load { [weak self, unowned featureTable] error in
```

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
philium: Ah, that makes sense. Thanks!

philium (approved):

yo1995 (approved):

PR #970, opened 2020-09-17
New: Apply raster rendering rule

yo1995:
This PR implements the `Layers` -> `Apply raster rendering rule` sample. Please refer to `common-samples/issues/392` for more details.

Please notice: the delete-add-back process is expected as designed. The image service does not allow changing rendering rule of a raster after it is loaded.

|demo|
|-|
|200917-rasterRenderingRule|


arcgis-ios-sdk-samples/Layers/Raster rendering rule/RasterRenderingRuleViewController.swift
```diff
@@ -0,0 +1,104 @@
...
+        // Load the raster and get a list of rendering rule info supported by the service.
+        imageServiceRaster.load { [weak self] error in
+            guard let self = self else { return }
+            if let serviceInfo = imageServiceRaster.serviceInfo, let extent = serviceInfo.fullExtent {
+                self.mapView.setViewpoint(AGSViewpoint(targetExtent: extent), completion: nil)
+                self.rasterRenderingRules = serviceInfo.renderingRuleInfos
+                self.chooseRenderingRuleBarButtonItem.isEnabled = true
+            } else if let error = error {
+                self.presentAlert(error: error)
+            }
```
vquach2404, line 54: I think we generally keep the error handling at the beginning of the completion handler but I don't see an issue with how it is now. Just a thought!

arcgis-ios-sdk-samples/Layers/Raster rendering rule/RasterRenderingRuleViewController.swift
```diff
@@ -0,0 +1,104 @@
...
+        )
+        rasterRenderingRules.forEach { ruleInfo in
+            let action = UIAlertAction(title: ruleInfo.name, style: .default) { _ in
+                // Clear all raster layers before adding new one.
+                self.mapView.map!.operationalLayers.removeAllObjects()
+                // Create a new `AGSRenderingRule` object with the chosen rule.
+                let renderingRule = AGSRenderingRule(renderingRuleInfo: ruleInfo)
+                let imageServiceRaster = AGSImageServiceRaster(url: self.imageServiceURL)
+                // Add the new raster layer to the map.
+                self.mapView.map!.operationalLayers.add(self.makeRasterLayer(raster: imageServiceRaster, renderingRule: renderingRule))
```
vquach2404, line 87: Would it be good to guard map in this case since it's unwrapped twice?

vquach2404 (commented):
Looks good! Just a few comments and questions

yo1995:
From Friday's review: change the name to Apply raster rendering rule

arcgis-ios-sdk-samples/Layers/Raster rendering rule/RasterRenderingRuleViewController.swift
yo1995: I cannot find the original post, but I think Phil suggested otherwise. Let's wait for Phil's call.

yo1995:
@vquach2404 I made some changes based on your suggestions! Please take another look.

vquach2404 (approved):
Thanks for implementing those changes! Ready for Phil's review.

arcgis-ios-sdk-samples/Layers/Raster rendering rule/RasterRenderingRuleViewController.swift
philium: @vquach2404 We sometimes need to put the error check first when that is the only way to know whether the operation failed. But when possible, I think it makes more sense to put the most likely (no error) case first.

yo1995: Thanks for clarification. I've documented this in the coding style section in our wiki. (/common-samples/wiki/iOS-Samples-Repo:-A-Contribution-Guide#first-success-then-failure)

arcgis-ios-sdk-samples/Layers/Apply raster rendering rule/ApplyRasterRenderingRuleViewController.swift
```diff
@@ -0,0 +1,116 @@
...
+    }
+    
+    /// Create a raster layer from an image service raster with an optional rendering rule.
+    ///
+    /// - Parameters:
+    ///   - raster: An `AGSImageServiceRaster` object.
+    ///   - renderingRule: An optional `AGSRenderingRule` object.
+    /// - Returns: An `AGSRasterLayer` object.
+    func makeRasterLayer(raster: AGSImageServiceRaster, renderingRule: AGSRenderingRule? = nil) -> AGSRasterLayer {
+        raster.renderingRule = renderingRule
```
philium, line 68: Since `renderingRule` is a property of the passed in `raster`, it seems like it should be set by whoever created the raster and passed it into this function in the first place. If it were moved outside of this function, then the only thing left would be `AGSRasterLayer(raster: raster)`, which makes me wonder if this function is worth having. What do you think?

arcgis-ios-sdk-samples/Layers/Apply raster rendering rule/ApplyRasterRenderingRuleViewController.swift
```diff
@@ -0,0 +1,116 @@
...
+            message: nil,
+            preferredStyle: .actionSheet
+        )
+        rasterRenderingRules.forEach { ruleInfo in
+            let action = UIAlertAction(title: ruleInfo.name, style: .default) { _ in
+                let map = self.mapView.map!
+                // Clear all raster layers before adding new one.
+                map.operationalLayers.removeAllObjects()
+                let rasterLayer: AGSRasterLayer
+                if self.rasterLayers[ruleInfo.name] != nil {
```
philium, line 87: Use `if let existingLayer = self.rasterLayers[ruleInfo.name] { ... }` instead of subscripting twice.

arcgis-ios-sdk-samples/Layers/Apply raster rendering rule/ApplyRasterRenderingRuleViewController.swift
yo1995: 😅 😅 That totally makes sense... Removed.

philium (approved):
Looks good!

yo1995:
@saratchandrakarumuri Please take a look at this new sample. Thanks!

saratchandrakarumuri (approved):
Sample looks good. I think the button will automatically follow the new color scheme when merged in `v.next`

yo1995:
Yes it should. After merge I'll double check!

PR #971, opened 2020-09-21
New: Add ENC exchange set

yo1995:
This PR implements the `Add ENC exchange set` sample. Please read its design at `/designs/AddENCExchangeSet`

The original issue is at `/common-samples/issues/483`

Branch: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/New-AddENCExchangeSet-vNext/arcgis-ios-sdk-samples/Layers/Add%20ENC%20exchange%20set


vquach2404 (approved):
Looks good! Just one suggestion but otherwise ready for Phil's review.

arcgis-ios-sdk-samples/Layers/Add ENC exchange set/AddENCExchangeSetViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+    /// Load the ENC dataset and add to the map view.
+    ///
+    /// - Parameter mapView: The map view managed by the view controller.
+    func addENCExchangeSet(mapView: AGSMapView) {
+        guard let map = mapView.map else { return }
+        // Load catalog file in ENC exchange set from bundle.
+        let catalogURL = Bundle.main.url(
+            forResource: "CATALOG",
+            withExtension: "031",
+            subdirectory: "ExchangeSetwithoutUpdates/ExchangeSetwithoutUpdates/ENC_ROOT"
```
philium, line 35: Any chance we could get the portal item updated so that there aren't two `ExchangeSetwithoutUpdates` directories? (Bonus points for getting the _w_ capitalized!) That just looks really odd.

arcgis-ios-sdk-samples/Layers/Add ENC exchange set/AddENCExchangeSetViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+        )!
+        let encExchangeSet = AGSENCExchangeSet(fileURLs: [catalogURL])
+        
+        // URL to the "hydrography" data folder that contains the "S57DataDictionary.xml" file.
+        let hydrographyDirectory = Bundle.main.url(
+            forResource: "S57DataDictionary",
+            withExtension: "xml",
+            subdirectory: "hydrography/hydrography"
```
philium, line 43: Same with `hydrography`. Any way we could update the zip to remove one of the `hydrography` directories?

arcgis-ios-sdk-samples/Layers/Add ENC exchange set/AddENCExchangeSetViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+        )!
+        .deletingLastPathComponent()
+        // Set environment settings for loading the dataset.
+        let environmentSettings = AGSENCEnvironmentSettings.shared()
+        environmentSettings.resourceDirectory = hydrographyDirectory
+        // The SENC data directory is for temporarily storing generated files.
+        environmentSettings.sencDataDirectory = FileManager.default.temporaryDirectory
```
philium, line 50: It would be a good idea to create a subdirectory inside the temporary directory (see [this](https://nshipster.com/temporary-files/) NSHipster article for advice on how to do that). It would also be good to delete the temporary directory when the sample is closed, which would probably involve storing the temporary directory URL and then removing the item at the URL when the view controller is deinited.

arcgis-ios-sdk-samples/Layers/Add ENC exchange set/AddENCExchangeSetViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+        
+        encExchangeSet.load { [weak self] error in
+            guard let self = self else { return }
+            if let error = error {
+                self.presentAlert(error: error)
+            } else {
+                var ENCLayers = [AGSENCLayer]()
```
philium, line 57: FYI: In Swift, variable names almost always begin with a lowercase letter. When the name begins with an acronym, the convention Apple uses is to lowercase all the letters in the acronym (like [`UIColor.cgColor`](https://developer.apple.com/documentation/uikit/uicolor/1621954-cgcolor), for example). So a better name for this would be `encLayers`. By the way, this naming convention differs from Objective-C.

arcgis-ios-sdk-samples/Layers/Add ENC exchange set/AddENCExchangeSetViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+                let loadGroup = DispatchGroup()
+                // A reference to the last error occurred, if there is any.
+                var loadError: Error?
+                // Create a list of ENC layers and load each layer.
+                encExchangeSet.datasets.forEach { dataset in
+                    let layer = AGSENCLayer(cell: AGSENCCell(dataset: dataset))
+                    ENCLayers.append(layer)
```
philium, line 64: Instead of building up the layers array one by one, it would be better to map the datasets to layers and then iterate over the layers, if needed. By the way, is iterating over the layers needed? Do the layers need to be explicitly loaded? If so, consider using `AGSLoadObjects(_:_:)`.

arcgis-ios-sdk-samples/Layers/Add ENC exchange set/AddENCExchangeSetViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+                at: tmpURL,
+                includingPropertiesForKeys: nil,
+                options: .skipsHiddenFiles
+            ) {
+                tmpDirectory.forEach { file in
+                    try? FileManager.default.removeItem(atPath: file.path)
+                }
+            }
+        }
+    }
```
philium, line 107: Ah, the temporary items are being removed on deinit. Good! However, this is way more complicated than it needs to be. As mentioned earlier, a subdirectory of the temporary directory should be used. Then, the subdirectory itself can just be removed using `FileManager.removeItem(at:)`. As far as I can tell, there's no reason to do that asynchronously by dispatching to a different queue.

philium (commented):
Does this look right?

image

That doesn't look anything like the included screenshot.

arcgis-ios-sdk-samples/Layers/Add ENC exchange set/AddENCExchangeSetViewController.swift
yo1995: After discussion with team, the current solution is to create replicas of these portal items with correct folder structure, instead of fixing the original files. But I'm afraid that will add burdens to the data management in the long run. (People may get confused why there are items with same content but different IDs.) 

@mikewilburn what do you think?

yo1995:
Other comments have been addressed except the portal item folder structure problem. Mentioned in our channel and is waiting for feedback.

arcgis-ios-sdk-samples/Layers/Add ENC exchange set/AddENCExchangeSetViewController.swift
mikewilburn: I'm not sure of the history that brought about the seemingly redundant `/ExchangeSetwithoutUpdates/ExchangeSetwithoutUpdates/` and `hydrograpy/hydrography` directory structure.

This is data we didn't produce ourselves, correct? It's supplementary data created by Esri/Runtime to permit the API to function as expected?

yo1995: These 2 items are replaced by replicas with folder structure fixed. However, the links in the README is still pointing to the original items to avoid confusion.

philium (approved):
Looks good!

yo1995:
@saratchandrakarumuri Please take a look at this new sample. Thanks!

saratchandrakarumuri (approved):
Sample looks good

PR #972, opened 2020-09-22
Edit KML ground overlay implementation

vquach2404:
Implementation for issue #1420 in `common-samples`.


yo1995:
Consider revert d116a6f4a94250df890fe9b81effe3af39044f5e and push again, or force push to overwrite these changes... 😁

vquach2404:
> Consider revert [d116a6f](https://github.com/Esri/arcgis-runtime-samples-ios/commit/d116a6f4a94250df890fe9b81effe3af39044f5e) and push again, or force push to overwrite these changes... 😁

I converted this to a draft after realizing this. I'm working on it now, please wait until I re-request your review. Thanks for your patience!

yo1995:
> > Consider revert [d116a6f](https://github.com/Esri/arcgis-runtime-samples-ios/commit/d116a6f4a94250df890fe9b81effe3af39044f5e) and push again, or force push to overwrite these changes... 😁
> 
> I converted this to a draft after realizing this. I'm working on it now, please wait until I re-request your review. Thanks for your patience!

Oops sorry I didn't refresh... sure no rush!

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/README.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+        "imagery",
+        "Keyhole",
+        "KML",
+        "KMZ",
+        "OGC",
+        "AGSKMLDataset",
+        "AGSKMLGroundOverlay",
+        "AGSKMLIcon",
+        "AGSKMLLayer"
+    ],
```
yo1995, line 18:
```suggestion
"keywords": [
        "KML",
        "KMZ",
        "Keyhole",
        "OGC",
        "imagery",
        "AGSKMLDataset",
        "AGSKMLGroundOverlay",
        "AGSKMLIcon",
        "AGSKMLLayer"
    ],
```

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/README.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+    "redirect_from": [],
+    "relevant_apis": [
+        "AGSKMLDataset",
+        "AGSKMLGroundOverlay",
+        "AGSKMLIcon",
+        "AGSKMLLayer"
+    ],
+    "snippets": [
+        "EditKMLGroundOverlayViewController"
```
yo1995, line 27:
```suggestion
        "EditKMLGroundOverlayViewController.swift"
```

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/README.md
```diff
@@ -0,0 +1,39 @@
+# Edit KML ground overlay
+
+Edit the values of a KML ground overlay.
+
+![Image of edit KML ground overlay](edit-kml-ground-overlay.png)
+
+## Use case
+
+KML ground overlays are used for showing aerial imagery, symbology, or other images draped over a scene. Changing the geometry, rotation, and other attributes of a ground overlay after it has been loaded allows for live editing.  For example, editing the geometry and opacity of a historical image draped over present day satellite imagery makes it possible to view change over time.
```
yo1995, line 9:
```suggestion
KML ground overlays are used for showing aerial imagery, symbology, or other images draped over a scene. Changing the geometry, rotation, and other attributes of a ground overlay after it has been loaded allows for live editing. For example, editing the geometry and opacity of a historical image draped over present day satellite imagery makes it possible to view change over time.
```

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+    func makeScene() -> AGSScene {
+        // Create a scene for the scene view.
+        let scene = AGSScene(basemap: .imagery())
+        // Create a geometry for the ground overlay.
+        let overlayGeometry = AGSEnvelope(xMin: -123.066227926904, yMin: 44.04736963555683, xMax: -123.0796942287304, yMax: 44.03878298600624, spatialReference: .wgs84())
+        // Create a KML icon for the overlay image.
+        let imageURL = URL(string: "https://libapps.s3.amazonaws.com/accounts/55937/images/1944.jpg")!
+        let overlayImage = AGSKMLIcon(url: imageURL)
+        // Apply a KML ground overlay.
+        overlay = AGSKMLGroundOverlay(geometry: overlayGeometry, icon: overlayImage)
```
yo1995, line 51: Per its doc
> A new initialized ground overlay or nil if an unsupported geometry is passed.

It seems this initializer will always succeed given that the geometry is fixed. Maybe consider force unwrap here?

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+    // The KML ground overlay.
+    var overlay: AGSKMLGroundOverlay?
+    
+    // MARK: Actions and methods
+    @IBAction func sliderValueChanged(_ slider: UISlider) {
+        // Change the color of the overlay according to the slider's value.
+        let alpha = CGFloat(slider.value)
+        overlay?.color = UIColor(red: 0, green: 0, blue: 0, alpha: alpha)
+        // Update the slider's value label.
+        valueLabel.text = String(format: "%.2f", slider.value)
```
yo1995, line 39: Phil may suggest using `NumberFormatter` for number to string conversion.

yo1995 (approved):
Overall looks good. A few small changes.

yo1995:
constraints
Also, consider setting a max width for the slider, or/and align the opacity label on the left to safe area leading space.

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
vquach2404: If I force unwrap there and allow `overlay` to possibly be nil, won't it cause problems afterward? I'll also have to unwrap `overlay` about 5 times.

arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
```diff
@@ -658,6 +658,14 @@
...
 				</array>
 			</dict>
+			<dict>
+				<key>descriptionText</key>
+				<string>Edit the values of a KML ground overlay.</string>
+				<key>storyboardName</key>
+				<string>EditKMLGroundOverlay</string>
+				<key>displayName</key>
+				<string>Edit KML ground overlay</string>
+			</dict>
```
yo1995, line 668: If you put this ahead of `Apply raster rendering rule`, please also adjust its order in the project file.

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
yo1995: I'm suggesting sth like
```swift
// Implicitly unwrap here.
var overlay: AGSKMLGroundOverlay!

// Since we know the initializer won't fail, it is safe to unwrap ❗  it.
let overlay = AGSKMLGroundOverlay(geometry: overlayGeometry, icon: overlayImage)!
// Assign the newly created object to the instance property.
self.overlay = overlay
```

But it totally up to your choice. It is fine to keep as-is. 😁  I prefer to bang it whenever it is safe.

vquach2404:
> Also, consider setting a max width for the slider, or/and align the opacity label on the left to safe area leading space.

I've aligned the slider view with the safe area which will make the opacity label also align with the safe area. But I don't see much of a visual difference or issue in the screenshot you provided. Is there an issue with the opacity label?

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+    @IBOutlet var valueLabel: UILabel!
+    
+    // The KML ground overlay.
+    var overlay: AGSKMLGroundOverlay!
+    
+    // MARK: Actions and methods
+    @IBAction func sliderValueChanged(_ slider: UISlider) {
+        // Change the color of the overlay according to the slider's value.
+        let alpha = CGFloat(slider.value)
+        overlay?.color = UIColor(red: 0, green: 0, blue: 0, alpha: alpha)
```
yo1995, line 37:
```suggestion
        overlay.color = UIColor(red: 0, green: 0, blue: 0, alpha: alpha)
```

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+        // Apply a KML ground overlay.
+        let overlay = AGSKMLGroundOverlay(geometry: overlayGeometry, icon: overlayImage)!
+        self.overlay = overlay
+        // Set the rotation of the ground overlay.
+        overlay.rotation = -3.046024799346924
+        // Create a KML dataset with the ground overlay as the root node.
+        let dataset = AGSKMLDataset(rootNode: overlay)
+        // Create a KML layer for the scene view.
+        let layer = AGSKMLLayer(kmlDataset: dataset)
+        // Add the layer to the map.
```
yo1995, line 59:
```suggestion
        // Add the layer to the scene.
```

yo1995 (approved):
LGTM 👍

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+        // Create a scene for the scene view.
+        let scene = AGSScene(basemap: .imagery())
+        // Create a geometry for the ground overlay.
+        let overlayGeometry = AGSEnvelope(xMin: -123.066227926904, yMin: 44.04736963555683, xMax: -123.0796942287304, yMax: 44.03878298600624, spatialReference: .wgs84())
+        // Create a KML icon for the overlay image.
+        let imageURL = URL(string: "https://libapps.s3.amazonaws.com/accounts/55937/images/1944.jpg")!
+        let overlayImage = AGSKMLIcon(url: imageURL)
+        // Apply a KML ground overlay.
+        let overlay = AGSKMLGroundOverlay(geometry: overlayGeometry, icon: overlayImage)!
+        self.overlay = overlay
```
philium, line 52: You could make this a pure function by having it take the ground overlay rather than creating it and assigning it to an instance variable: `func makeScene(groundOverlay: AGSKMLGroundOverlay) -> AGSScene`. The advantage there is that the `overlay` property could be made immutable and non-optional: `let overlay: AGSKMLGroundOverlay = { ... }()`. It is up to you, but that is what I would recommend.

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlay.storyboard
```diff
@@ -0,0 +1,88 @@
...
+                <viewController id="NFA-Uf-2BM" customClass="EditKMLGroundOverlayViewController" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <view key="view" contentMode="scaleToFill" id="VhX-3L-oDy">
+                        <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
+                        <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
+                        <subviews>
+                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="QlX-Cr-Hln" userLabel="Scene View" customClass="AGSSceneView">
+                                <rect key="frame" x="0.0" y="0.0" width="414" height="812"/>
+                                <color key="backgroundColor" systemColor="systemBackgroundColor"/>
+                            </view>
+                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="OBw-N8-s44" userLabel="Slider View">
```
philium, line 24: Did you try using a toolbar for this view? It is positioned where a toolbar should be, but it doesn't have the right background color or the right size. Also, the spacing on the right of the view is different than the spacing on the left. Rather than using a constant for the spacing, the first and last views should be constrained to the _margins_ of the superview:

image

That will use the default margins, but they can be changed if desired:

image

philium (changes requested):
Thank you for implementing this sample! It looks pretty good, but I've suggested a few changes.

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlay.storyboard
vquach2404: I considered using a toolbar but decided against it because the only "button" is the slider. I also saw a different sample that put the view inside a toolbar. Would it be best to just put the slider in a toolbar without a view?

yo1995: It is OK to use a toolbar, since our only control is a slider. And we've done that in samples such as #922 .

However, putting a slider in toolbar as a bar button item fixes its width. It may look bad on iPad landscape with a tiny slider in the middle.
It is also not easy to configure the labels in a toolbar (need to embed them in views and set constraints). So I think either using toolbar, or use a custom wrapper view or stack view is fine.

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlay.storyboard
```diff
@@ -0,0 +1,116 @@
...
+                                            <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
+                                            <subviews>
+                                                <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Opacity:" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="Gzf-D2-LS1">
+                                                    <rect key="frame" x="0.0" y="0.0" width="70" height="36"/>
+                                                    <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                                    <nil key="textColor"/>
+                                                    <nil key="highlightedColor"/>
+                                                </label>
+                                            </subviews>
+                                            <color key="backgroundColor" systemColor="systemBackgroundColor"/>
```
yo1995, line 40: Please set this wrapper view's background color to default.

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlay.storyboard
```diff
@@ -0,0 +1,116 @@
...
+                                                    <rect key="frame" x="0.0" y="0.0" width="40" height="33"/>
+                                                    <constraints>
+                                                        <constraint firstAttribute="width" constant="40" id="SAK-yB-6kS"/>
+                                                    </constraints>
+                                                    <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                                    <nil key="textColor"/>
+                                                    <nil key="highlightedColor"/>
+                                                </label>
+                                            </subviews>
+                                            <color key="backgroundColor" systemColor="systemBackgroundColor"/>
```
yo1995, line 75: Same here.

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+        let dataset = AGSKMLDataset(rootNode: groundOverlay)
+        // Create a KML layer for the scene view.
+        let layer = AGSKMLLayer(kmlDataset: dataset)
+        // Add the layer to the scene.
+        scene.operationalLayers.add(layer)
+        // Move the viewpoint to the ground overlay.
+        let overlayCenter = groundOverlay.geometry.extent.center
+        let camera = AGSCamera(lookAt: overlayCenter, distance: 1250, heading: 45, pitch: 60, roll: 0)
+        let targetExtent = groundOverlay.geometry as! AGSEnvelope
+        sceneView.setViewpoint(AGSViewpoint(targetExtent: targetExtent, camera: camera))
```
yo1995, line 69: Since the method still access an instance property `overlay` it is not pure function. Consider moving the set view point portion to scene view `didSet` observer.

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+            return numberFormatter
+        }()
+        valueLabel.text = numberFormatter.string(from: slider.value as NSNumber)
+    }
+    
+    func makeScene(groundOverlay: AGSKMLGroundOverlay) -> AGSScene {
+        // Create a scene for the scene view.
+        let scene = AGSScene(basemap: .imagery())
+        // Set the rotation of the ground overlay.
+        groundOverlay.rotation = -3.046024799346924
```
yo1995, line 58: Could this be set in overlay's init closure?

yo1995 (commented):
Some suggestions below.

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
vquach2404: After moving line 58 (setting rotation), wouldn't it be a pure function since it's not mutating `overlay` or any of its values?

yo1995: Ahh I'm sorry I mean `sceneView` instead of overlay. 😅 
This function access sceneView, which is an instance property. Moving it out can make it a pure function.

yo1995 (dismissed):
LGTM 👍

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+class EditKMLGroundOverlayViewController: UIViewController {
+    // MARK: Storyboard views and properties
+    @IBOutlet var sceneView: AGSSceneView! {
+        didSet {
+            sceneView.scene = makeScene(groundOverlay: overlay)
+            // Move the viewpoint to the ground overlay.
+            let overlayCenter = overlay.geometry.extent.center
+            let camera = AGSCamera(lookAt: overlayCenter, distance: 1250, heading: 45, pitch: 60, roll: 0)
+            let targetExtent = overlay.geometry as! AGSEnvelope
+            sceneView.setViewpoint(AGSViewpoint(targetExtent: targetExtent, camera: camera))
```
philium, line 27: If you know that the overlay's geometry is an envelope (as evidenced by the force cast to `AGSEnvelope`), then there's no need to get the extent of the geometry. This would be equivalent, no?

```suggestion
            let targetExtent = overlay.geometry as! AGSEnvelope
            let camera = AGSCamera(lookAt: targetExtent.center, distance: 1250, heading: 45, pitch: 60, roll: 0)
            sceneView.setViewpoint(AGSViewpoint(targetExtent: targetExtent, camera: camera))
```

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+        return overlay
+    }()
+    
+    // MARK: Actions and methods
+    @IBAction func sliderValueChanged(_ slider: UISlider) {
+        // Change the color of the overlay according to the slider's value.
+        let alpha = CGFloat(slider.value)
+        overlay.color = UIColor.black.withAlphaComponent(alpha)
+        // Update the slider's value label.
+        let numberFormatter: NumberFormatter = {
```
philium, line 54: Since this formatter is used multiple times (every time this method is called), it would be good to create it once and then reuse it. Please make it an instance property.

arcgis-ios-sdk-samples/Layers/Edit KML ground overlay/EditKMLGroundOverlayViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+            let numberFormatter = NumberFormatter()
+            numberFormatter.maximumFractionDigits = 2
```
philium, line 56: Going from two decimal places to zero for `0` and `1` looks a bit odd. I suggest setting `minimumFractionDigits` to the same value so that they appear as `0.00` and `1.00`.

philium (approved):

yo1995 (approved):

saratchandrakarumuri (approved):
Looks good

vquach2404:
@philium @yo1995 Sorry, I had to resolve merge conflicts and updated the value label, please review once more. Thanks!

yo1995 (approved):

philium (approved):

PR #977, opened 2020-10-01
Fixed memory leaks caused by retain cycles

vquach2404:
This PR fixes the memory leaks caused by alert retain cycles.

### DIFF
--- a/arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
+++ b/arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
@@ -153,7 +153,7 @@ class EditFeaturesWithFeatureLinkedAnnotationViewController: UIViewController {
         }
         alert.addAction(doneAction)
         // Add an observer to ensure the user does not input an empty string.
-        textFieldObserver = NotificationCenter.default.addObserver(forName: UITextField.textDidChangeNotification, object: nil, queue: .main, using: { [unowned alert] _ in
+        textFieldObserver = NotificationCenter.default.addObserver(forName: UITextField.textDidChangeNotification, object: nil, queue: .main, using: { [unowned alert, unowned doneAction] _ in
             // Enable the done button if both textfields are not empty.
             doneAction.isEnabled = (alert.textFields?.allSatisfy { $0.text?.isEmpty == false })!
         })
--- a/arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
+++ b/arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
@@ -153,7 +153,7 @@ class EditFeaturesWithFeatureLinkedAnnotationViewController: UIViewController {
         }
         alert.addAction(doneAction)
         // Add an observer to ensure the user does not input an empty string.
-        textFieldObserver = NotificationCenter.default.addObserver(forName: UITextField.textDidChangeNotification, object: nil, queue: .main, using: {_ in
+        textFieldObserver = NotificationCenter.default.addObserver(forName: UITextField.textDidChangeNotification, object: nil, queue: .main, using: { [unowned alert] _ in
             // Enable the done button if both textfields are not empty.
             doneAction.isEnabled = (alert.textFields?.allSatisfy { $0.text?.isEmpty == false })!
         })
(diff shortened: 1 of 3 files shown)


arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
```diff
@@ -153,7 +153,7 @@ class EditFeaturesWithFeatureLinkedAnnotationViewController: UIViewController {
         }
         alert.addAction(doneAction)
         // Add an observer to ensure the user does not input an empty string.
-        textFieldObserver = NotificationCenter.default.addObserver(forName: UITextField.textDidChangeNotification, object: nil, queue: .main, using: {_ in
+        textFieldObserver = NotificationCenter.default.addObserver(forName: UITextField.textDidChangeNotification, object: nil, queue: .main, using: { [unowned alert] _ in
             // Enable the done button if both textfields are not empty.
             doneAction.isEnabled = (alert.textFields?.allSatisfy { $0.text?.isEmpty == false })!
```
yo1995, line 158: I'm a little bit unsure. 
Do we also need to capture the `doneAction` here? When it comes to observers I'm always lost on their life cycles... 🤯

arcgis-ios-sdk-samples/Utility Network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
```diff
@@ -204,18 +204,18 @@ class ConfigureSubnetworkTraceOptionsViewController: UITableViewController {
...
-                // Enable the done button if the textfield is not empty and is a valid number.
-                doneAction.isEnabled = NumberFormatter().number(from: text) != nil
-            } else {
-                doneAction.isEnabled = false
+            // Add an observer to ensure the user does not input an empty string.
+            textFieldObserver = NotificationCenter.default.addObserver(
+                forName: UITextField.textDidChangeNotification,
+                object: textField,
+                queue: .main
+            ) { _ in
```
yo1995, line 212: Same here for the action.

yo1995 (commented):
The changes look good, but I've more questions...

arcgis-ios-sdk-samples/Edit data/Edit features with feature-linked annotation/EditFeaturesWithFeatureLinkedAnnotationViewController.swift
vquach2404: I think since those actions are dependent on the alert, I don't think there's a need to capture them. Could you confirm @philium ?

philium: > Do we also need to capture the `doneAction` here?

Good thought. `doneAction` _is_ being captured here. The question is, how should it be captured? To answer that, consider 1) who has a strong reference to it, and 2) who has a strong reference to those who have a strong reference to it. Here, the observer is retaining the done action. But who is retaining the observer? The Done and Cancel actions? Yes and no. They are holding a reference to the `textFieldObserver` variable, whose value is the observer. Since the variable holds a strong reference to the observer, the actions are as well. There are two different ways to address that:

1) Make the `doneAction` here `unknown`, or
2) Set `textFieldObserver` to `nil` after removing the observer.

Both options break the retain cycle, but they do so in different ways. The first option prevents the retain cycle from being created in the first place. The second option causes the observer to go away when either button is tapped, breaking the retain cycle and allowing the Done button to go away as well.

Does that make sense?

vquach2404: I made `doneAction` unowned. Please review once again!

yo1995 (approved):

arcgis-ios-sdk-samples/Edit data/Edit with branch versioning/EditWithBranchVersioningViewController.swift
```diff
@@ -376,7 +376,7 @@ class EditWithBranchVersioningViewController: UIViewController {
                 object: textField,
                 queue: .main
             ) { _ in
-                let text = alertController.textFields![0].text!.trimmingCharacters(in: .whitespacesAndNewlines)
+                let text = textField.text!.trimmingCharacters(in: .whitespacesAndNewlines)
                 // Enable the done button if branch version textfield is not empty.
                 createAction.isEnabled = !text.isEmpty
```
philium, line 381: There's the same issue here with the observer holding a strong reference to `createAction` and `createAction` extending the lifetime of `textFieldObserver`. Please fix, like you did for `doneAction`, here and in the other samples (if necessary).

philium (changes requested):

arcgis-ios-sdk-samples/Edit data/Edit with branch versioning/EditWithBranchVersioningViewController.swift
yo1995: 🤦‍♂️  My bad, I didn't pay attention to the not changed lines...

philium (approved):

yo1995:
@vquach2404 Are there more changes coming for this PR? Thanks.

vquach2404:
> @vquach2404 Are there more changes coming for this PR? Thanks.

Nope, it's ready for review! I thought I had already requested your review, my mistake!

yo1995 (approved):

PR #978, opened 2020-10-05
CI: Add GitHub Actions workflow and style checking tools

yo1995:
This PR adds a set of scripts and 2 GitHub Actions workflow YAML files to `v.next` branch, to enforce styling for `README.md` and `README.metadata.json` files.

Steps to review:

1. (Optional) Read about the background in /common-samples/issues/2018
1. (Optional) Read about the GitHub Actions wiki in /common-samples/wiki/README-and-metadata-status-check-with-GitHub-Actions
1. Try it out to see how it works in my forked repo: https://github.com/yo1995/arcgis-runtime-samples-ios/pulls. 
    
    I've set the repo up exactly the same as this PR. When you create a PR to `v.next` in that repo, it will run the checks and give success/failure status, and print error output into the console.
1. Review the files
    - `main.yml` and `action.yml` are GitHub Actions configuration files
    - `style.rb` is a `markdownlint` configuration file, and has been reviewed by samples team
    - `Dockerfile` is a docker to setup the environment to run Python scripts and Ruby markdownlint package
    - `entry.py` is the entry point (glue code) script called by the docker, that call the other scripts
    - The other 4 Python scripts check various rules against README and metadata. There are duplicate code in them because each script is designed to be a standalone tool that can run separately

    These checkers have been used and adjusted to deliver update 9 and should work pretty well, given that the website doesn't have noticeable issue right now.

1. In #869 in this [comment](https://github.com/Esri/arcgis-runtime-samples-ios/pull/869#pullrequestreview-451994085) you can find explanation of what are being checked by the scripts.

---

In short, feel free to skim through and give approval as the scripts have been proven working well. No sample is changed so this PR won't break the build.

We can make improvements afterwards if these checker scripts run into issues in production.

### DIFF
--- a/Scripts/CI/README_Metadata_StyleCheck/description_differ.py
+++ b/Scripts/CI/README_Metadata_StyleCheck/description_differ.py
@@ -1,0 +1,314 @@
+#!/usr/bin/env python3
+
+import os
+import re
+import json
+import plistlib
+import argparse
+from typing import List
+
+
+# region Global sets
+# A set of category folder names in current sample viewer.
+categories = {
+    'Maps',
+    'Layers',
+    'Features',
+    'Display information',
+    'Search',
+    'Edit data',
+    'Geometry',
+    'Route and directions',
+    'Analysis',
+    'Cloud and portal',
+    'Scenes',
+    'Utility network',
+    'Augmented reality'
+}
+
+excluded_samples = {
+    'Map loaded',
+    'Animate 3D graphic',
+    'Densify and generalize',
+    'Add graphics with renderer'
+}
+# endregion
+
+
+# region Static functions
+def sub_special_char(string: str) -> str:
+    """
+    Check and substitute if a string contains special characters.
+
+    :param string: The input string.
+    :return: A new string without special characters.
+    """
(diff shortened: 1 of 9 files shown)


vquach2404 (approved):
Thanks for writing these scripts! They'll be really useful!

Scripts/CI/README_Metadata_StyleCheck/description_differ.py
```diff
@@ -0,0 +1,314 @@
...
+import re
+import json
+import plistlib
+import argparse
+from typing import List
+
+
+# region Global sets
+# A set of category folder names in current sample viewer.
+categories = {
```
philium, line 13: It looks like the list of categories exists in multiple places. Would it be possible to extract them into a separate file and then pass them in as an argument of the script? Just a thought.

philium (approved):

Scripts/CI/README_Metadata_StyleCheck/description_differ.py
yo1995: Yes. Eventually I want to combine all the helper functions and classes into a single file so that they are easier to maintain. Meanwhile I'll first put these checks into production. If they work well during this release, we can proceed with further optimization.

PR #979, opened 2020-10-06
Update: remove -j flag from download script and fix portal item

yo1995:
This PR fixes the issue in #857 (Animate images with image overlay) and #865 , where we decided to add "-j" flag to call `unzip` so that it omits all enclosing folders and extract files to a flattened directory. It also replaced the item that caused trouble.

Removing the flag means now extracting zip archive will preserve their folder structure, and we need to

- Add the folder tree as-is when introducing new item to the `Portal Data` folder
- (Negotiate with sample designer to) ensure future zip archives do not contain irrelevant hidden files or redundant folders

### DIFF
--- a/arcgis-ios-sdk-samples/Content Display Logic/PortalItems.plist
+++ b/arcgis-ios-sdk-samples/Content Display Logic/PortalItems.plist
@@ -6,12 +6,6 @@
 	<array>
 		<dict>
 			<key>identifier</key>
-			<string>74c0c9fa80f4498c9739cc42531e9948</string>
-			<key>filename</key>
-			<string>loudoun_anno.geodatabase</string>
-		</dict>
-		<dict>
-			<key>identifier</key>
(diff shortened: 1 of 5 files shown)


arcgis-ios-sdk-samples/Content Display Logic/PortalItems.plist
```diff
@@ -66,9 +60,15 @@
 		</dict>
 		<dict>
 			<key>identifier</key>
-			<string>9465e8c02b294c69bdb42de056a23ab1</string>
+			<string>d1453556d91e46dea191c20c398b82cd</string>
```
philium, line 63: Please update the item identifier in the sample's README.

philium (changes requested):

philium (approved):

yo1995:
@saratchandrakarumuri Please verify `Animate images with image overlay` sample still work as expected. Thanks~

vquach2404 (approved):
This works fine on my end!

PR #981, opened 2020-10-09
Update Vector tiled layer (custom style)

vquach2404:
Please refer to common-samples/340. This PR updates the vector tiled layer (custom style) sample to include usage of `AGSArcGISVectorTiledLayer(vectorTileCache:itemResourceCache:)` which creates a vector tiled layer using offline data.


arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,88 @@ import UIKit
...
+            // Start the job.
+            exportVectorTilesJob?.start(statusHandler: { (status) in
+                SVProgressHUD.show(withStatus: status.statusString())
+            }, completion: { [weak self] (result, error) in
+                SVProgressHUD.dismiss()
+                guard let self = self else { return }
+                if let error = error {
+                    // Handle errors.
+                    self.presentAlert(error: error)
+                } else if let result = result {
```
yo1995, line 73: Consider first result then error.

yo1995 (commented):
A few suggestions.

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,88 @@ import UIKit
...
+            map.basemap = AGSBasemap(baseLayer: vectorTiledLayer)
+            // Set the viewpoint.
+            let centerPoint = AGSPoint(x: 1990591.559979, y: 794036.007991, spatialReference: .webMercator())
+            self.mapView.setViewpoint(AGSViewpoint(center: centerPoint, scale: 88659253.829259947))
+        }
+    }
+    
+    private func getItemResourceCacheURL() -> URL {
+        // Get a suitable directory to place files.
+        let directoryURL = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
```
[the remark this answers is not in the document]
yo1995: Also you can use a closure to make this a get-only property rather than a function.

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyle.storyboard
```diff
@@ -299,6 +301,63 @@
...
+                                                            <integer key="value" value="5"/>
+                                                        </userDefinedRuntimeAttribute>
+                                                    </userDefinedRuntimeAttributes>
+                                                </view>
+                                                <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Offlilne custom style" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontForContentSizeCategory="YES" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="LgN-yo-i5R">
+                                                    <rect key="frame" x="68" y="12" width="338" height="20"/>
+                                                    <fontDescription key="fontDescription" style="UICTFontTextStyleBody"/>
+                                                    <color key="textColor" red="0.0" green="0.0" blue="0.0" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
```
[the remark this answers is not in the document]
vquach2404: What color issues are you referring to and how can I fix them?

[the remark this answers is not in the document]
yo1995: It would be great if you can check every color with `colorSpace="custom"` in the storyboard source code, and fix any that hasn't to be custom color.

But it is also OK to only fix the label colors for the cells (change them to default), so that the text are white in darkmode.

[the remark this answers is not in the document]
vquach2404: Should I delete `colorSpace="custom"` altogether or should I set it to something else?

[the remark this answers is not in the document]
yo1995: For each custom color in the storyboard, find which view/property it belongs to, and change it to default color if it applies.

For example, for each static cell, when you tap on the label, it will show its label color is "custom". Reset it to default and in the storyboard it won't have the `colorSpace="custom"` anymore.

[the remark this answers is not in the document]
yo1995: There are other colors like cell background, tableview background, etc. that was set to custom due to legacy issues. It would be great if we can fix them whenever we see them.

vquach2404:
I updated the code so that the sample loads the custom style and local .vtpk only once upon the start of the sample. That way, only one temporary URL is created and the user can reload the layer quicker. Please let me know what you think!

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyle.storyboard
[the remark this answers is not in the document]
yo1995: Colors look great now!

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyle.storyboard
```diff
@@ -11,15 +14,14 @@
...
-                        <rect key="frame" x="0.0" y="0.0" width="600" height="600"/>
+                        <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
                             <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="k3a-nA-IYu" customClass="AGSMapView">
-                                <rect key="frame" x="0.0" y="0.0" width="600" height="556"/>
-                                <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                <rect key="frame" x="0.0" y="44" width="414" height="774"/>
                             </view>
                             <toolbar opaque="NO" clipsSubviews="YES" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="Sqg-eH-RPH">
```
yo1995, line 23: Please remove the toolbar then add again, to make it fill the bottom safe area.

image

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,89 @@ import UIKit
...
+        // Create a unique name for the item resource cache based on current timestamp.
+        let formattedDate = ISO8601DateFormatter().string(from: Date())
+        // Create and return the full, unique URL.
+        return directoryURL.appendingPathComponent("\(formattedDate)")
+    }()
+    
+    // The item ID of the shown layer.
+    var shownItemID: String?
+    // The job to export the item resource cache.
+    var exportVectorTilesJob: AGSExportVectorTilesJob?
```
yo1995, line 44: It seems this property is only written to but not read by the method, consider removing it and using a local variable.

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,89 @@ import UIKit
...
+    
+    private func showSelectedItem(_ itemID: String) {
+        guard let map = mapView.map else { return }
+        shownItemID = itemID
+        if itemID == itemIDs.last {
+            // If the custom style is chosen, assign the local vector tile package to the basemap.
+            map.basemap = AGSBasemap(baseLayer: offlineVectorTiledLayer!)
+            // Set the viewpoint to display the Palm Springs Convention Center.
+            let point = AGSPoint(x: -116.5384, y: 33.8258, spatialReference: .wgs84())
+            self.mapView.setViewpoint(AGSViewpoint(center: point, scale: 3_000))
```
yo1995, line 82:
```suggestion
            mapView.setViewpoint(AGSViewpoint(center: point, scale: 3_000))
```

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,89 @@ import UIKit
...
+            }
+        }
+    }
+    
+    private func showSelectedItem(_ itemID: String) {
+        guard let map = mapView.map else { return }
+        shownItemID = itemID
+        if itemID == itemIDs.last {
+            // If the custom style is chosen, assign the local vector tile package to the basemap.
+            map.basemap = AGSBasemap(baseLayer: offlineVectorTiledLayer!)
```
yo1995, line 79: While it is usually OK to assume the offline layer is already loaded before user select it, this is not safe in the following situation:
- When a user has downloaded the on-demand resource, but lost network access afterwards. When they select the  offline custom style, the sample will crash.

Consider disabling the UI or blocking user interaction if the offlineVectorTiledLayer is nil.

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,89 @@ import UIKit
...
+    private let itemIDs = ["1349bfa0ed08485d8a92c442a3850b06",
                            "bd8ac41667014d98b933e97713ba8377",
                            "02f85ec376084c508b9c8e5a311724fa",
-                           "1bf0cc4a4380468fbbff107e100f65a5"]
-    private var shownItemID: String? {
-        return ((mapView.map?.basemap.baseLayers.firstObject as? AGSArcGISVectorTiledLayer)?.item as? AGSPortalItem)?.itemID
+                           "1bf0cc4a4380468fbbff107e100f65a5",
+                           "2056bf1b350244d69c78e4f84d1ba215"]
+    
+    let temporaryURL: URL? = {
```
yo1995, line 32:
```suggestion
    let temporaryURL: URL = {
```

1. This doesn't need to be optional

2. Since this is only used by the load function, consider making it a local constant.

yo1995 (commented):
It works, and here are a few more suggestions.

Feel free to add Phil to review if the fixes are done before next week.

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
vquach2404: I originally had this as a local variable but apparently with most, if not all `AGSJob`s, you have to keep a global reference or else the job won't complete. If it were a local variable, the job would go out of scope and get deallocated so the callback would never get called.

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyle.storyboard
vquach2404: I removed it and added it again but I'm not sure if that solved the issue. Please let me know if there's something else I should do!

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,89 @@ import UIKit
...
 class VectorTileCustomStyleViewController: UIViewController, VectorStylesVCDelegate {
-    @IBOutlet var mapView: AGSMapView!
-    
-    private var itemIDs = ["1349bfa0ed08485d8a92c442a3850b06",
+    @IBOutlet var mapView: AGSMapView! {
+        didSet {
+            mapView.map = AGSMap()
+        }
+    }
+    @IBOutlet var changeStyleBBI: UIBarButtonItem!
```
philium, line 25: Please [avoid abbreviations](https://swift.org/documentation/api-design-guidelines/#use-terminology-well) except for highly recognized standard ones, like URL:

image

For variables whose type is `UIBarButtonItem`, I recommend following Apple's [convention](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621471-editbuttonitem) of using `BarItem` as a suffix.

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyle.storyboard
```diff
@@ -11,48 +14,41 @@
...
-                                <rect key="frame" x="0.0" y="0.0" width="600" height="556"/>
-                                <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                <rect key="frame" x="0.0" y="44" width="414" height="769"/>
                             </view>
-                            <toolbar opaque="NO" clipsSubviews="YES" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="Sqg-eH-RPH">
-                                <rect key="frame" x="0.0" y="556" width="600" height="44"/>
-                                <constraints>
-                                    <constraint firstAttribute="height" relation="lessThanOrEqual" constant="44" id="6ah-Dy-HjX"/>
-                                </constraints>
+                            <toolbar opaque="NO" clipsSubviews="YES" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="pRt-Jl-9L6">
```
philium, line 23: The toolbar does not extend into the safe area. Please fix.

image

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyle.storyboard
```diff
@@ -299,6 +291,62 @@
...
+                                            </subviews>
+                                            <constraints>
+                                                <constraint firstItem="Eva-Qw-W63" firstAttribute="centerY" secondItem="Wft-Cj-G9I" secondAttribute="centerY" id="JOy-kX-YB8"/>
+                                                <constraint firstItem="LgN-yo-i5R" firstAttribute="leading" secondItem="Eva-Qw-W63" secondAttribute="trailing" constant="8" id="Zae-Iw-78C"/>
+                                                <constraint firstItem="LgN-yo-i5R" firstAttribute="top" secondItem="Wft-Cj-G9I" secondAttribute="top" constant="12" id="ghY-dt-saR"/>
+                                                <constraint firstAttribute="bottom" secondItem="LgN-yo-i5R" secondAttribute="bottom" constant="12" id="j7r-wN-gbO"/>
+                                                <constraint firstAttribute="leadingMargin" secondItem="Eva-Qw-W63" secondAttribute="leading" id="wT9-FB-pbC"/>
+                                                <constraint firstAttribute="trailing" secondItem="LgN-yo-i5R" secondAttribute="trailing" constant="8" id="zXZ-Ty-Bbp"/>
```
philium, line 346: I realize this cell was likely a copy of one of the existing cells, but this constraint's relation should really be `greaterThanOrEqual` so that the label is only as wide as needed to fit its text.

Equal | Greater Than or Equal
----- | -----
Screen Shot 2020-10-17 at 1 16 23 AM 1 | Screen Shot 2020-10-17 at 1 20 23 AM

philium (changes requested):
Thank you for updating this sample! Changes look good for the most part, but I have requested a few minor changes.

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,89 @@ import UIKit
...
+                           "2056bf1b350244d69c78e4f84d1ba215"]
+    // The item ID of the shown layer.
+    var shownItemID: String?
+    // The job to export the item resource cache.
+    var exportVectorTilesJob: AGSExportVectorTilesJob?
+    // The vector tiled layer created by the local VTPK.
+    var offlineVectorTiledLayer: AGSArcGISVectorTiledLayer?
+    
+    func loadOfflineLayer() {
+        let temporaryURL: URL = {
```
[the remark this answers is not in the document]
vquach2404: Hm I suppose. I gave it a try and I don't think it'll cause any issues where I placed the deletion. Please give it a look!

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,90 @@ import UIKit
...
+                let itemresourceCahce = result.itemResourceCache
+                // Create a vector tiled layer with the vector tiled cache and the item resource cache.
+                self.offlineVectorTiledLayer = AGSArcGISVectorTiledLayer(vectorTileCache: vectorTileCache, itemResourceCache: itemresourceCahce)
+                self.changeStyleBarItem.isEnabled = true
+            } else if let error = error {
+                // Handle errors.
+                self.presentAlert(error: error)
+            }
+        }
+        try? FileManager.default.removeItem(at: temporaryURL)
```
philium, line 72: This deletes the temporary directory before the job has even started. This should go inside of the completion handler.

philium (changes requested):
My comment about the label's trailing constraint being equal instead of greater-than-or-equal was marked as resolved, but I'm still seeing it as an equal constraint:

image

Am I missing something?

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
vquach2404: I tried to add it inside the completion handler but then the layer didn't appear at all. I thought it was possibly deleting the directory before the job finished. Is there an additional step I should do to include the deletion in the completion handler?

vquach2404:
> My comment about the label's trailing constraint being equal instead of greater-than-or-equal was marked as resolved, but I'm still seeing it as an equal constraint:

Hm I'm not sure why that change wasn't committed. It should be fixed now, sorry about that!

arcgis-ios-sdk-samples/Content Display Logic/PortalItems.plist
```diff
@@ -77,6 +77,10 @@
 			<string>SA_EVI_8Day_03May20.zip</string>
 		</dict>
 		<dict>
+			<key>filename</key>
+			<string>PSCC_vector.vtpk</string>
+			<key>identifier</key>
+			<string>aae86ca2cf8f4e6087065fb8bf1b6ecd</string>
```
philium, line 83: These should be in their own dictionary. As part of this one, this dictionary now has multiple values for the keys `filename` and `identifier`. In the plist editor, the last one read wins out:

image

philium (commented):
> It should be fixed now

It is. Thanks! Although, I noticed another issue with all of those trailing constraints. They are constrained to the superview with a constant of `8`. They should be constrained to the superview margin with a constant of `0`, since the margin includes the appropriate padding for a table view row. It might be a good idea to create an issue to fix all those trailing constraints.

vquach2404:
Trevor has proposed new data for this sample. It won't be long until the new data is ready so I'll hold off on requesting reviews until then.

vquach2404:
@yo1995 I made some major changes including adding new data and reorganizing the code structure. Please give this another look, thanks!

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,122 @@ import UIKit
...
+            } else if let error = error {
+                // Handle errors.
+                self.presentAlert(error: error)
+            }
+            }
+        )
+        try? FileManager.default.removeItem(at: temporaryURL)
     }
     
     private func showSelectedItem(_ itemID: String) {
```
yo1995, line 95: By tapping on the custom style twice, it seems to disappear. See the screenshot below.
select-bug

yo1995 (commented):
There is a small bug for show selected item. Please fix that and then let me do another review! Also a few minor issues.

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -4600,6 +4604,7 @@
 					MontereyElevation,
 					MontereyElevationRaster,
 					NDVIRaster,
+					PSCC_vector,
```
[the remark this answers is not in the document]
yo1995: Oh good it seems this is solved!

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -821,6 +821,7 @@
 		F140E1EE23F36707008AC44E /* DisplayAnnotationViewController.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = F140E1EA23F35E92008AC44E /* DisplayAnnotationViewController.swift */; };
 		F14A8F012474AD9D00242D96 /* ConfigureSubnetworkTrace.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = F14A8F002474AD9D00242D96 /* ConfigureSubnetworkTrace.storyboard */; };
 		F14F2CA323FC768200693196 /* DisplaySubtypeSettingsViewController.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = F1987D0223F7677E00DF41CC /* DisplaySubtypeSettingsViewController.swift */; };
+		F1510FD3254A18E200C29150 /* dodge_city.vtpk in Resources */ = {isa = PBXBuildFile; fileRef = F1510FD2254A18E200C29150 /* dodge_city.vtpk */; settings = {ASSET_TAGS = (DodgeCity, ); }; };
```
yo1995, line 824: Hmm it seems during the update the source code and README are dropped. Could you add them back?

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleViewController.swift
```diff
@@ -17,40 +17,122 @@ import UIKit
...
+        // Set the viewpoint to display Dodge City, KS.
+        let point = AGSPoint(x: -100.01766, y: 37.76528, spatialReference: .wgs84())
+        mapView.setViewpoint(AGSViewpoint(center: point, scale: 40000))
+        // Get the vector tiled layer URL.
         let vectorTiledLayerURL = URL(string: "https://arcgisruntime.maps.arcgis.com/home/item.html?id=\(itemID)")!
-        let vectorTiledLayer = AGSArcGISVectorTiledLayer(url: vectorTiledLayerURL)
-        mapView.map?.basemap = AGSBasemap(baseLayer: vectorTiledLayer)
+        switch itemID {
+        case "e01262ef2a4f4d91897d9bbd3a9b1075":
+            if dayVectorTiledLayer == nil {
```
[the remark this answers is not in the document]
vquach2404: I actually tried to change` itemIDs` into key value pairs since dictionaries don't always load in the same order. But I can't store the offline `AGSArcGISVectorTiledLayer`s in the key value pairs because they must be loaded using an `AGSExportVectorTilesJob`; the result of which can't be assigned on the main thread. Please let me know if I can clarify that. It's a little difficult to explain 😅

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
vquach2404: I don't see any files that we deleted or missing. Which files are you referring to?

yo1995: The README and source code. When tapping on the upper right info button, the README doesn't load. Correct me if I'm wrong and I can take another look.

vquach2404: Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-05 at 16 41 33
Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-05 at 16 41 35
Do you mean the info button in the Simulator? They loaded okay for me

yo1995: I see. I'll take another look in next review! 😳

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyle.storyboard
yo1995: This is fixed for the remaining entries.

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorStylesViewController.swift
```diff
@@ -26,8 +26,8 @@ class VectorStylesViewController: UITableViewController {
     
     weak var delegate: VectorStylesVCDelegate?
     
-    override func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
```
yo1995, line 30: By setting UI in `viewWillAppear`, it will cause `UITableViewAlertForLayoutOutsideViewHierarchy` warning.

I cannot think of a good way to update a static table before it is loaded, so maybe we'll need Phil's help.

With this change there will be a short delay before the checkmark appears, while it doesn't throw warnings anymore.

yo1995 (commented):
@vquach2404  I've done some refactoring based on your version, but I don't've more time to do a thorough clean. Feel free to make additional changes! (e.g. README, screenshot, etc.)

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorStylesViewController.swift
philium: I haven't tested it, but I think the same technique used by `OptionsTableViewController` could be employed. Simply override `tableView(_:cellForRowAt:)` to set the accessory type:

```swift
override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
    let cell = super.tableView(tableView, cellForRowAt: indexPath)
    cell.accessoryType = itemIDs[indexPath.row] == selectedItemID ? .checkmark : .none
    return cell
}
```

And then reload/deselect rows as appropriate in `tableView(_:didSelectRowAt:)`.

vquach2404: I couldn't reload the cells because they came out blank after. I fixed the checkmark delay but it may not be as efficient as reloading the cells. Please let me know what you think!

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorStylesViewController.swift
```diff
@@ -21,31 +21,34 @@ protocol VectorStylesVCDelegate: AnyObject {
...
-        super.viewWillAppear(animated)
-        if let selectedItemID = selectedItemID,
-            let row = itemIDs.firstIndex(of: selectedItemID) {
-            let indexPath = IndexPath(row: row, section: 0)
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = super.tableView(tableView, cellForRowAt: indexPath)
+        if itemIDs[indexPath.row] == selectedItemID {
+            // Select and indicate the displayed item.
+            cell.accessoryType = .checkmark
             tableView.selectRow(at: indexPath, animated: false, scrollPosition: .none)
```
philium, line 36: This seems odd to me. The "selected" row is both selected _and_ checked? It seems like it should be one or the other. Could this line be removed?

vquach2404: `selectRow(at:animated:scrollPosition:)` doesn't prompt `tableView(_:didSelectRowAt:)` (ref: [Apple doc](https://developer.apple.com/documentation/uikit/uitableview/1614875-selectrow#discussion) ) so a checkmark won't display when the view is initially loaded. And if I don't select the row, `tableView(_:didDeselectRowAt:)`won't prompt appropriately. I could try to work around it but I think this is the cleanest solution. What do you think?

philium: Ah! I see what you mean. Right now, selection is actually being tracked in two places (sort of): 1) the table view's selected row (used to check/uncheck rows), and 2) `selectedItemID`, although currently that is only used to check a row when the table view is first displayed. Instead, I propose that `selectedItemID` should be used for both purposes, and that could easily be done by adding the following lines to `tableView(_:didSelectRowAt:)`:

```swift
if let selectedItemID = selectedItemID, let row = itemIDs.firstIndex(of: selectedItemID) {
    tableView.cellForRow(at: IndexPath(row: row, section: 0))?.accessoryType = .none
}
selectedItemID = itemID
```

Then the implementation of `tableView(_:didDeselectRowAt:)` isn't needed.

vquach2404: Great idea! I cleaned up the code a bit and tried to make it easier to follow. Please let me know what you think!

philium (approved):

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorStylesViewController.swift
```diff
@@ -16,36 +16,34 @@
...
+        // Indicate which cell has been selected.
         tableView.cellForRow(at: indexPath)?.accessoryType = .checkmark
-    }
-    
-    override func tableView(_ tableView: UITableView, didDeselectRowAt indexPath: IndexPath) {
-        tableView.cellForRow(at: indexPath)?.accessoryType = .none
+        
+        if let previousItemID = selectedItemID, let previousRow = itemIDs.firstIndex(of: previousItemID) {
+            tableView.cellForRow(at: IndexPath(row: previousRow, section: 0))?.accessoryType = .none
+        }
```
yo1995, line 46: I'd suggest swap the order of these 2 lines. Otherwise, when tapping on a cell for multiple times, it won't display the checkmark.
Or consider `guard let` early in the function to abort further changes if `selectedItemID` is not changed.
multiple

yo1995 (changes requested):

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/VectorStylesViewController.swift
vquach2404: Nice catch! I reorganized. Please take a look!

yo1995 (approved):

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/README.md
```diff
@@ -15,14 +15,24 @@ Pan and zoom to explore the vector tile basemap.
 
 ## How it works
 
-1. Construct an `AGSArcGISVectorTiledLayer` with the URL of a custom style.
+1. Construct an `AGSArcGISVectorTiledLayer` with the URL of a custom style from AGOL.
+    * Follow these steps to create a vector tiled layer with a custom style from offline resources:
+    1. Construct an `AGSVectorTileCache` using the name of the local vector tile package.
```
saratchandrakarumuri, line 20: I feel using a different scheme like a,b,c or i,ii,iii would look better a sublist here

saratchandrakarumuri (commented):
Everything looks good except for one comment below

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/README.md
yo1995: Haha Markdown does render as i, ii, iii 👇 even it is written as 1, 2, 3... 😁 
image

philium (dismissed):

arcgis-ios-sdk-samples/Layers/ArcGIS vector tiled layer (custom style)/README.md
saratchandrakarumuri: Ohh... that is weird... I haven't check on git markdown, but the readme page in our viewer renders like this

IMG_838FAB48DA4B-1

yo1995: > Ohh... that is weird... I haven't check on git markdown, but the readme page in our viewer renders like this

I see. Yeah apparently GitHub is using a more advanced Markdown parser... 🤥 
On our developer site it also doesn't render correctly as secondary numbered list style. i.e. https://github.com/Esri/arcgis-runtime-samples-ios/pull/981#discussion_r527280529

So yeah maybe consider replacing them with unordered list asterisk, or manually do a, b, c etc.

vquach2404: I manually made the roman numerals. Please let me know what you think, thanks!

yo1995 (approved):

saratchandrakarumuri (approved):
Looks good

philium (approved):

yo1995:
@vquach2404 It seems the status check on https://github.com/Esri/arcgis-runtime-samples-ios/pull/981/commits/a005df3ab0691e574e6cf302037550d3a514f8fe gave errors for trailing spaces (and apparently all of us missed the big red ❌  ! 🤦 ) 

Can you make another PR to fix it? Thanks.

vquach2404:
> @vquach2404 It seems the status check on [a005df3](https://github.com/Esri/arcgis-runtime-samples-ios/commit/a005df3ab0691e574e6cf302037550d3a514f8fe) gave errors for trailing spaces (and apparently all of us missed the big red ❌ ! 🤦 )
> 
> Can you make another PR to fix it? Thanks.
> 
> Probably we should consider disable merging if status check fails.

The trailing spaces were intentional. I had to make the sublist manually so I had to make new lines without extra spacing after each step.

yo1995:
> The trailing spaces were intentional. I had to make the sublist manually so I had to make new lines without extra spacing after each step.

Hmm that's not ideal. There are a few workarounds

- newline between lines: this increases the spacing between lines, which is not good
- backslash at the end of the line: the Markdown renderer in app does not render it correctly
- put `<br/>` at the end of the line: works everywhere, but doesn't look good in source

I replaced the 2 whitespaces with `<br/>` in #1010 , so that it won't trigger style check error.
But it might be better to revert to numbered list. What do you think?

vquach2404:
Why isn't it ideal? I think the roman numeral sublist makes the readme easier to follow, especially with the small iPhone screen so we should try to preserve it.

yo1995:
> Why isn't it ideal? I think the roman numeral sublist makes the readme easier to follow, especially with the small iPhone screen so we should try to preserve it.

It breaks the status check... Did you notice the  ❌  on the right of the commit sha?
image

We can definitely opt-out the trailing spaces rule in markdownlint to ignore this case, but that will allow any trailing spaces to go into our README, which we tried to avoid in our README clean-up effort. 🤔

vquach2404:
I understand that it breaks the status check but does it affect the entire repo? My understanding is that the status checks only act as an aid to help us catch any unintentional errors. Can we allow this specific case?

yo1995:
> ...Can we allow this specific case?

That is a great idea. I'll make the following changes to the rules so that it allows the exception to use spaces as line break.
https://github.com/updownpress/markdown-lint/blob/master/rules/009-no-trailing-spaces.md

PR #983, opened 2020-10-20
Update: Replace SVProgressHUD with ProgressHUD

yo1995:
Please refer to `/common-samples/issues/2102` for more details.

|Original|light|dark|
|-|-|-|
|original|light|dark|

## Initiative

The original `SVProgressHUD` is the only ObjC dependency in our app, and it is not Dark Mode compatible. Therefore, we would replace it with a modern, Swift-native progress HUD that looks and feels great. It should follow these requirements

- [x] Short enough - not like `SVProgressHUD` that span across multiple files with thousands lines of code.
- [x] No bundle image needed
- [x] Dark Mode compatible
- [ ] *Cancel button (see below)
- [x] Not to deal with UIWindow or root VC
- [x] Translucent and look and feel modern
- [ ] No parallax effect - the original SVProgressHUD does have that, but it is trivial...

## Leverage `ProgressHUD`

Currently other OSAs don't have an in-house HUD implementation. Esri's mapbook app maintained by Eli [uses `PKHUD`](https://github.com/Esri/mapbook-ios/tree/master/mapbook-iOS/PKHUD), which still has lots of issues.

[`ProgressHUD`](https://github.com/relatedcode/ProgressHUD) is a short and clean 3rd party Swift HUD that has been recently maintained by the author. It behaves almost the same as the previous HUD, while including many more features.

The only drawback is that ProgressHUD does not come with a configurable cancel button. See below.

### Cancel Button

Per Phil's request, we want our new HUD to have a cancel button, so that we may add the cancel action as a closure for an `AGSCancelable` task. However, it is not easy to bind an action closure to a `UIButton`. Before iOS 13, it involves `objc_setAssociatedObject` to manage the button's lifecycle which doesn't look great. 

With [`UIAction`](https://developer.apple.com/documentation/uikit/uiaction) or `Combine`, it is easier to add a closure as the action of a button with fewer lines of code. Therefore, I decided not to add the cancel button for now, and we may eventually add it when our samples app drops iOS 12.

More discussions can be found below

- https://stackoverflow.com/questions/25919472/adding-a-closure-as-target-to-a-uibutton
- https://www.swiftbysundell.com/tips/creating-closure-based-ui-controls-with-uiaction/

### Changes in the source code

Refer to c87b3b9efeb27a72eac5e67f2d42331702ac57b9 for the changes that make it compatible with iOS 12.

## Usage

The API is pretty much the same as before

```swift
// Show the HUD with system default `UIActivityIndicatorView` spinner and a prompt
ProgressHUD.show("Loading")
// Dismiss the HUD
ProgressHUD.dismiss()
```

---
This PR will be marked as ready after #899 is merged.


yo1995:
@vquach2404 Please give it a try and share your thoughts! No need to delve into the source code as it is mostly adapted from the original implementation, but feel free to if you want.

vquach2404 (commented):
I tested each sample and they all seem to work as expected except for `Analysis/Viewshed (geoprocessing). `After loading, there seems to be an error and the results are not displayed. However, I don't suspect this is an issue due to the changes in this PR since I saw the same behavior in `v.next`. Can you confirm this issue? 
viewshedGeoprocessing

vquach2404:
> I tested each sample and they all seem to work as expected except for `Analysis/Viewshed (geoprocessing). `After loading, there seems to be an error and the results are not displayed. However, I don't suspect this is an issue due to the changes in this PR since I saw the same behavior in `v.next`. Can you confirm this issue?

We've concluded that it is indeed a service issue. Ready for Phil's review 👍🏼

philium:
> The API is pretty much the same as before

Was this not implemented as a modal view controller? One of the downsides to SVProgressHUD is how it is presented. It behaves as a singleton view that is added to the app's window. That is not a good pattern to imitate. Just like a popover or an alert, I think the HUD should be a view controller that is presented and dismissed as needed. I'm curious: did you explore that approach?

yo1995:
> ...Just like a popover or an alert, I think the HUD should be a view controller that is presented and dismissed as needed. ...

Hi Phil, I'm intermittently working on this design and have a few questions/thoughts

1. Do we want to keep the current signature, or use the native methods for a view controller?

i.e. current

```swift
// Show the HUD with system default `UIActivityIndicatorView` spinner and a prompt
ProgressHUDController.show(self, "Loading")
// Dismiss the HUD
ProgressHUDController.dismiss()
```

or native

```swift
let controller = ProgressHUDController(/* ...some initializer here */)
// ...
present(controller, animated: true)
controller.dismiss()
```

2. Since the new implementation is a view controller, it is not feasible to show a progress value, as we need to dismiss and present again to update a progress view. So I didn't implement that.

3. Currently for some `AGSJob`, the call site look like this

```swift
someAGSJob.start(
    statusHandler: { (status: AGSJobStatus) in
        SVProgressHUD.show(withStatus: status.statusString()) //Show job status.
    },
    completion: { [weak self] (_, error: Error?) in
        SVProgressHUD.dismiss()
        // handle the result
    }
)
```
And the `statusHandler` might be called several times. This will involve presenting and dismissing the HUD controller several times. Does it make sense to disable the animation for `present` and `dismiss` in this situation?

Thanks.

When it is finished, I'll build a demo app to make it easier to try it out.

philium:
> Do we want to keep the current signature, or use the native methods for a view controller?

If the HUD becomes a view controller, it should be presented and dismissed as one. So no, I would not try to maintain the same API as SVProgressHUD.

> Since the new implementation is a view controller, it is not feasible to show a progress value, as we need to dismiss and present again to update a progress view.

Why is that? Why couldn't it be created with an instance of `Progress` and observe that to update the progress view, like `UIProgressView` does?

> Does it make sense to disable the animation for `present` and `dismiss` in this situation?

It doesn't make sense to dismiss and represent in that situation. If the HUD is a view controller, the sample could store a reference to the presented view controller and update whatever properties are needed while the view controller is presented.

I hope that helps. In a way, getting rid of `SVProgressHUD` is a lot like the transition away from `UIAlertView`. An alert view could be created anywhere by anyone and presented simply by calling `show()`. It was awful! Apple replaced it with `UIAlertController` which, as the name implies, is a view controller and therefore can only be presented by one. That really was a monumental change for the better and I see the removal of `SVProgressHUD` the same way. If it could be replaced by a `UIViewController`-based implementation, I am all for that!

yo1995:
Eli and Mark implemented a `UIViewController`-based solution.

References
- https://github.com/Esri/data-collection-ios/pull/284
- https://github.com/Esri/data-collection-ios/pull/289

esreli:
> Eli and Mark implemented a `UIViewController`-based solution.
> 
> References
> 
> * [Esri/data-collection-ios#284](https://github.com/Esri/data-collection-ios/pull/284)
> * [Esri/data-collection-ios#289](https://github.com/Esri/data-collection-ios/pull/289)

Just a small note here. This solution works really well for our purposes but has not been put through rigorous review. With the business decision surrounding OSA, we afforded this task only marginal attention.

yo1995:
Close and use #1059 instead.

PR #986, opened 2020-10-23
Update: Export tiles sample with new .tpkx format

yo1995:
This PR updates the `Export tiles` sample to support exporting the new `.tpkx` file format. Details at: `common-samples/issues/2255`

This PR is part of the whole efforts of upgrading all `.tpk` tile package files to the new format.


arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
```diff
@@ -17,154 +17,242 @@ import UIKit
...
+    // MARK: Properties
+    
+    /// The tiled layer created from world street map service.
+    let tiledLayer = AGSArcGISTiledLayer(url: URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/World_Street_Map/MapServer")!)
+    /// The export task to request the tile package with the same URL as the tile layer.
+    let exportTask = AGSExportTileCacheTask(url: URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/World_Street_Map/MapServer")!)
+    /// An export job to download the tile package.
+    var job: AGSExportTileCacheJob! {
+        didSet {
+            exportTilesBarButtonItem.title = job == nil ? "Export tiles" : "Cancel"
```
yo1995, line 66: Actually it is impossible to tap the cancel button, as the `SVProgressHUD` blocks all user interaction when being presented. This issue also exists in the original sample, so I didn't make a change.

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
```diff
@@ -17,154 +17,242 @@ import UIKit
...
-                if (error as NSError).code != NSUserCancelledError {
-                    self.presentAlert(error: error)
-                }
-            } else if let tileCache = result {
+            if let tileCache = result {
                 self.visualEffectView.isHidden = false
                 
                 let newTiledLayer = AGSArcGISTiledLayer(tileCache: tileCache)
                 self.previewMapView.map = AGSMap(basemap: AGSBasemap(baseLayer: newTiledLayer))
+                self.zoomToViewpoint(mapView: self.previewMapView, scale: 1e7)
```
yo1995, line 156: This line is added to cope with the new `.tpkx` package format. Without this line

- The original `.tpk` package loads correctly
- But the new `.tpkx` package does not show up at the correct extent

Comment out this line to see the difference. Is it expected?

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
```diff
@@ -17,154 +17,242 @@ import UIKit
...
-        let minPoint = self.mapView.screen(toLocation: frame.origin)
-        let maxPoint = self.mapView.screen(toLocation: CGPoint(x: frame.origin.x + frame.width, y: frame.origin.y + frame.height))
-        let extent = AGSEnvelope(min: minPoint, max: maxPoint)
-        return extent
+    // MARK: Properties
+    
+    /// The tiled layer created from world street map service.
+    let tiledLayer = AGSArcGISTiledLayer(url: URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/World_Street_Map/MapServer")!)
+    /// The export task to request the tile package with the same URL as the tile layer.
+    let exportTask = AGSExportTileCacheTask(url: URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/World_Street_Map/MapServer")!)
```
yo1995, line 62: Both share the same service URL. I tend to use a static let for the URL, but Phil might suggest against it. 🤔

yo1995 (commented):
Below are some comments for the refactoring.

yo1995:
@vquach2404 Though it seems to be lots of changes in src, the functionalities actually didn't change much 😅 and the PR is mostly about bug fixing and refactoring . You may compare this PR with our current AppStore release to see the difference. Thanks!

vquach2404 (commented):
Simulator Screen Shot - iPhone 11 Pro Max - 2020-10-23 at 16 46 16
I zoomed in as much as I could and exported the tiles but got a blank result for both .tpk and .tpkx. Is this expected behavior? 

The close button is also difficult to read because of the purple text against the dark background. Could you add a tint or maybe a toolbar to make the button more legible?

yo1995:
> I zoomed in as much as I could and exported the tiles but got a blank result for both .tpk and .tpkx. Is this expected behavior?

Ah my fault. I happened to hard code the scale of preview. Will fix that.

> The close button is also difficult to read because of the purple text against the dark background. Could you add a tint or maybe a toolbar to make the button more legible?

This is addressed here: https://github.com/Esri/arcgis-runtime-samples-ios/pull/899#discussion_r510427480 I'll test again with the Dark Mode changes.

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
```diff
@@ -17,154 +17,238 @@ import UIKit
...
-            print("deleted all local data")
-        } catch {
-            print(error)
         }
     }
+    
+    deinit {
+        // Remove the temporary folder and all content in it.
+        try? FileManager.default.removeItem(at: temporaryFolderURL)
+    }
```
vquach2404, line 253: Is it necessary to have both this and the `removeTemporaryTilePackages()` function?

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
vquach2404: Hm, I think the cancel button should be removed since it'll never be used.

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
yo1995: Yes. The `removeTemporaryTilePackages` function removes all the files in the folder, but doesn't remove the folder, so if a user want to export another tile package at a different viewpoint, he can save to disk without overriding existing files.
The remove in `deinit`, on the other hand, remove all the files including the folder as a clean-up.

Think about a case like this: user taps export at San Diego once, then closes the preview, then export at LA, and quit the app without closing the preview.

However, I agree that the `removeTemporaryTilePackages` function might be moved into the close action. Currently it is get called when new export is initiated. What do you think?

vquach2404: Hm, yeah it might make more sense to move the function call to the close action.

vquach2404 (dismissed):

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
```diff
@@ -17,154 +17,233 @@ import UIKit
...
+    @IBAction func closeButtonTapped(_ sender: UIButton) {
+        // Hide the preview and background.
+        visualEffectView.isHidden = true
+        // Release the map in order to free the tiled layer.
+        previewMapView.map = nil
+        // Remove all files in the sample-specific temporary folder.
+        guard let files = try? FileManager.default.contentsOfDirectory(at: temporaryFolderURL, includingPropertiesForKeys: nil), !files.isEmpty else { return }
+        files.forEach { filePath in
+            try? FileManager.default.removeItem(at: filePath)
+        }
```
philium, line 225: Could the temporary directory just be removed altogether?

```suggestion
        do {
            try FileManager.default.removeItem(at: temporaryFolderURL)
        } catch {
            print("Error removing temporary directory: \(error.localizedDescription)")
        }
```

arcgis-ios-sdk-samples/Layers/Export tiles/README.md
```diff
@@ -11,7 +11,7 @@ Field workers with limited network connectivity can use exported tiles as a base
 
 ## How to use the sample
 
-Pan and zoom into the desired area, making sure the area is within the red boundary. Tap the "Export tiles" button to start the process. On successful completion you will see a preview of the downloaded tile package.
+Pan and zoom into the desired area, making sure the area is within the red boundary. Tap the "Export tiles" button and choose an export format to start the process. On successful completion you will see a preview of the downloaded tile package.
```
philium, line 14: I think there should be a comma after `completion`.

```suggestion
Pan and zoom into the desired area, making sure the area is within the red boundary. Tap the "Export tiles" button and choose an export format to start the process. On successful completion, you will see a preview of the downloaded tile package.
```

arcgis-ios-sdk-samples/Layers/Export tiles/README.md
```diff
@@ -33,6 +33,8 @@ Pan and zoom into the desired area, making sure the area is within the red bound
 
 ArcGIS tiled layers do not support reprojection, query, select, identify, or editing. Visit the [ArcGiS Online Developer's portal](https://developers.arcgis.com/ios/latest/swift/guide/layer-types-described.htm#ESRI_SECTION1_30E7379BE7FE4EC2AF7D8FBFEA7BB4CC) to learn more about the characteristics of ArcGIS tiled layers.
 
+Refer to [Tile Package Specification](https://github.com/Esri/tile-package-spec) on GitHub for more information of the tile package format.
```
philium, line 36:
```suggestion
Refer to the [Tile Package Specification](https://github.com/Esri/tile-package-spec) on GitHub for more information on the tile package format.
```

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
```diff
@@ -17,154 +17,233 @@ import UIKit
...
+    // MARK: Actions
+    
+    @IBAction func exportTilesBarButtonTapped(_ sender: UIBarButtonItem) {
+        if let exportJob = job {
+            // If it is downloading, cancel download.
+            exportJob.progress.cancel()
+            job = nil
+        } else {
+            // Otherwise, try to download when exporting tiles is allowed.
+            guard let mapServiceInfo = exportTask.mapServiceInfo, mapServiceInfo.exportTilesAllowed else { return }
```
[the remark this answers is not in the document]
yo1995: Thanks for catching it! Forgot to clean it up after trying with different ideas during refactoring. 😅

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
yo1995: Mentioned above in https://github.com/Esri/arcgis-runtime-samples-ios/pull/986#discussion_r512336655

If the temp folder is removed before a user quit the sample, there would possibly be a case that the user want to export another tile package, while the temp folder does not exist, which leads to an error.

philium: Could the temp directory (not the URL, but the directory itself) be created in `makeDownloadURL(fileFormat:)`? That way if it is deleted, it will be re-created when needed, right?

yo1995: Great idea! In this way I can also remove the closure to init the temp directory URL.

yo1995:
From Friday's review it's mentioned that README needs some update. Will fix that later today.
- minScale
- initial location

yo1995:
Per discussion with Trevor and Luke, this update is temporarily block due to a server side issue.

yo1995:
The UI requires a small change (remove options) to mitigate the extent issue and make the sample future proof.

yo1995:
This PR is ready for another review.

arcgis-ios-sdk-samples/Layers/Export tiles/README.md
```diff
@@ -11,15 +11,16 @@ Field workers with limited network connectivity can use exported tiles as a base
...
 ## How it works
 
-1. Create an `AGSExportTileCacheTask`, passing in the URL of the tiled layer.
-2. Create default `AGSExportTileCacheParameters` for the task, specifying the area of interest, minimum scale and maximum scale.
-3. Use the parameters and a path to create an `AGSExportTileCacheJob` from the task.
-4. Start the job, and when it completes successfully, get the resulting `AGSTileCache`.
-5. Use the tile cache to create an `AGSArcGISTiledLayer` and display it in the map.
+1. Create a map and set its `minScale` to 10,000,000. Limiting the scale in this sample limits the potential size of the selection area, thereby keeping the exported tile package to a reasonable size.
+2. Create an `AGSExportTileCacheTask`, passing in the URL of the tiled layer.
+3. Create default `AGSExportTileCacheParameters` for the task, specifying the area of interest, minimum scale and maximum scale.
```
vquach2404, line 20:
```suggestion
3. Create default `AGSExportTileCacheParameters` for the task, specifying the area of interest, minimum scale, and maximum scale.
```

vquach2404 (approved):
Looks good, just one comma suggestion!

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
```diff
@@ -17,154 +17,218 @@ import UIKit
...
+            job = nil
+        } else if let mapServiceInfo = exportTask.mapServiceInfo, mapServiceInfo.exportTilesAllowed {
+            // Otherwise, try to download when exporting tiles is allowed.
+            if mapServiceInfo.exportTileCacheCompactV2Allowed {
+                // Export using the CompactV2 (.tpkx) if it is supported.
+                self.initiateDownload(exportTask: self.exportTask, downloadFileURL: self.makeDownloadURL(fileFormat: .tpkx))
+            } else {
+                // Otherwise, use the CompactV1 (.tpk) format.
+                self.initiateDownload(exportTask: self.exportTask, downloadFileURL: self.makeDownloadURL(fileFormat: .tpk))
             }
```
philium, line 200:
```suggestion
            let tilePackageFormat: TilePackageFormat
            if mapServiceInfo.exportTileCacheCompactV2Allowed {
                // Export using the CompactV2 (.tpkx) if it is supported.
                tilePackageFormat = .tpkx
            } else {
                // Otherwise, use the CompactV1 (.tpk) format.
                tilePackageFormat = .tpk
            }
            self.initiateDownload(exportTask: exportTask, downloadFileURL: makeDownloadURL(fileFormat: tilePackageFormat))
```

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
yo1995: > Is there still a cancel button? If not, should "Cancel" be removed?

I have split views on this... a bit unsure

- If we keep the cancel button, then it make more sense to replace the HUD with a progress view that displays the progress of the export job. I'll need to redesign the UI a bit if we go this way.
- However, since the export tile task it the major functionality to demonstrate, maybe we shouldn't allow user to cancel it. And showing how to canceling an `AGSJob` is trivial for this sample.

What do you think? I prefer to remove the cancel button.

philium: I'm fine either way; whatever you think is best.

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
```diff
@@ -17,154 +17,217 @@ import UIKit
...
+    /// A view to provide a dark blurry background to preview the exported tiles.
+    @IBOutlet var visualEffectView: UIVisualEffectView!
+    /// A map view to preview the exported tiles.
+    @IBOutlet var previewMapView: AGSMapView! {
+        didSet {
+            previewMapView.layer.borderColor = UIColor.white.cgColor
+            previewMapView.layer.borderWidth = 8
+        }
     }
+    /// A bar button to initiate or cancel the download task.
```
philium, line 51:
```suggestion
    /// A bar button to initiate the download task.
```

arcgis-ios-sdk-samples/Layers/Export tiles/ExportTilesViewController.swift
```diff
@@ -17,154 +17,217 @@ import UIKit
...
+            .appendingPathExtension(fileFormat.fileExtension)
+    }
+    
+    // MARK: Actions
+    
+    @IBAction func exportTilesBarButtonTapped(_ sender: UIBarButtonItem) {
+        if let exportJob = job {
+            // If an existing job is downloading, cancel the download.
+            exportJob.progress.cancel()
+            job = nil
```
philium, line 187: Is this code still valid?

yo1995: No... Great catch!

philium (approved):

yo1995:
@saratchandrakarumuri The sample look a little bit different with the changes above, but it should work pretty much the same as before. Please take a look! Thanks.

saratchandrakarumuri (approved):
Sample looks good

PR #987, opened 2020-10-27
Update: Create terrain from local tile package with new .tpkx format

yo1995:
This PR updates the `Create terrain from local tile package` sample to read the new .tpkx file format. 
Details at: common-samples/issues/2255

This PR is part of the whole efforts of upgrading all .tpk tile package files to the new format.

Notes:

- The new .tpkx file features the exact same extent of the original package
- The new .tpkx file has 1 less level of detail for tiles, which reduced its size by ~30 MB
- The [original .tpk file on the portal](https://arcgisruntime.maps.arcgis.com/home/item.html?id=cce37043eb0440c7a5c109cf8aad5500) is not removed, to ensure older releases of the sample viewer can still access this sample

Also inviting @TADraeseke to take a look.


philium (approved):

philium (approved):

yo1995:
@saratchandrakarumuri Please take a look and this sample should work exactly the same as before... 😄

saratchandrakarumuri (approved):
Sample works as expected

yo1995:
According to @TADraeseke this might need to be reverted before U10 release. Will keep an eye on that.

PR #989, opened 2020-10-28
Use default value for ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES

philium:
I happened to notice that the built app was embedding the Swift standard libraries unnecessarily. This PR updates the `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` to use the default value, which is `NO`. In my brief testing, this shaves 8 to 10 seconds off of builds times.

Screen Shot 2020-10-28 at 12 52 42 PM


yo1995 (approved):
👍  Paste are some [context here.](https://stackoverflow.com/questions/58063771/swift-package-manager-how-to-always-embed-swift-standard-libraries)

vquach2404 (approved):

PR #991, opened 2020-11-05
Fix: Change info.plist to fix compliance warning

yo1995:
This PR fixes the Compliance warning on AppStore Connect so that next time the app is uploaded, it will automatically go into internal testing after processing is done.

|1|2|
|-|-|
|compliance-1|compliance-2|

When the app is uploaded to AppStore Connect, it won't immediately go into internal testing, but get stuck by the compliance warning. Since we are not using encrypted communication in our app, simply setting it to false will solve the problem. Read more at

- https://developer.apple.com/documentation/bundleresources/information_property_list/itsappusesnonexemptencryption
- https://stackoverflow.com/questions/35841117

### DIFF
--- a/arcgis-ios-sdk-samples/Info.plist
+++ b/arcgis-ios-sdk-samples/Info.plist
@@ -45,6 +45,8 @@
 	</array>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<key>ITSAppUsesNonExemptEncryption</key>
+	<false/>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>NSAppTransportSecurity</key>
@@ -124,8 +126,6 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
-	<key>ITSAppUsesNonExemptEncryption</key>
-	<false/>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
 </dict>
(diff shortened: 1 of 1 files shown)


vquach2404 (approved):

philium (approved):

philium:
Wait!

philium:
Did you manually edit the plist? The new key/value pair isn't ordered correctly. Key/value pairs should be sorted by key. When the file is re-saved by Xcode, it makes the following change:
image

yo1995:
> Did you manually edit the plist? The new key/value pair isn't ordered correctly. Key/value pairs should be sorted by key. When the file is re-saved by Xcode, it makes the following change

Hmm interesting... I right-clicked on the blank space to "Add Row" to the plist. Why Xcode is not adding the row with the correct order in the first place 🤔 . Nevertheless I'll fix that.

philium:
Did you edit the file directly or through the target's Info tab?
image
A plist is just an XML file, so sorting isn't part of the spec. Editing a plist through Xcode's plist editor will preserve whatever order you use. But `Info.plist` is a special plist file managed by Xcode, so it should be modified through the Info tab, which does maintain a certain order. Does that make sense?

yo1995:
> Did you edit the file directly or through the target's Info tab?

Ah I edited by directly opening the plist from the project navigator.

> ...But `Info.plist` is a special plist file managed by Xcode, so it should be modified through the Info tab, which does maintain a certain order?

Thanks for pointing this out! I wasn't aware of it. Great to learn some new quirks about Xcode. 😆

philium (approved):

vquach2404 (approved):

PR #995, opened 2020-11-06
[Dark Mode] Update: Geometry

yo1995:
This PR updates the storyboards in `Geometry` category to adapt to Dark Mode, and update the screenshots.


vquach2404 (commented):
These background colors in these two call outs remain the same in dark and light mode. Is this expected?
Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-19 at 14 41 52
Also, the text in Project is illegible in dark mode. 
Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-19 at 14 17 35

yo1995:
I'm not quite sure what is the correct approach. Both are expected, but I won't merge this PR until we reach a consensus.
For the `Nearest vertex` sample, it uses the native `title` and `detail` properties of a callout; for the `Project` sample, it uses a custom view.

> These background colors in these two call outs remain the same in dark and light mode. Is this expected?

It is expected. Currently the callout is not Dark Mode adaptive, so the background and text colors are hardcoded.

> Also, the text in Project is illegible in dark mode.

Since the labels are in a custom (stack) view, they adapt to Dark Mode changes, whereas the callout's background color doesn't change... so it looks weird in Dark Mode.

---

@philium For now, should we hard code label text's color to black in a custom view in a callout? 
I see the attribution footer is currently being updated by Zach for Dark Mode, and I wonder if callout will also be updated soon. 😳

vquach2404 (dismissed):
Everything else looks good, feel free to add more reviewers!

philium:
> should we hard code label text's color to black in a custom view in a callout?

Yes, I would say change what is needed so that it displays properly now.

saratchandrakarumuri (approved):
Changes are good

philium (approved):

PR #996, opened 2020-11-10
[Dark Mode] Update: Analysis

vquach2404:
This PR updates the Analysis category to support dark mode.


vquach2404:
There seems to be a glitch with the popover in the Statistical query (group and sort) sample. 
ezgif-6-b9269e61d69a

I've stepped through the code and it doesn't seem to be caused by anything on our end. Could you confirm @yo1995 ?

yo1995:
> I've stepped through the code and it doesn't seem to be caused by anything on our end. This occurs on `v.next` but not on the AppStore version. Could you confirm @yo1995 ?

Yes please refer to `/common-samples/issues/2149` It is OK to leave it there and we may tackle it later.

arcgis-ios-sdk-samples/Analysis/Viewshed (geoprocessing)/ViewshedGeoprocessing.storyboard
```diff
@@ -19,26 +20,27 @@
...
                                         <rect key="frame" x="8" y="5" width="398" height="36"/>
                                         <fontDescription key="fontDescription" type="system" pointSize="15"/>
                                         <nil key="textColor"/>
                                         <nil key="highlightedColor"/>
                                     </label>
                                 </subviews>
-                                <color key="backgroundColor" red="0.97254901959999995" green="0.97254901959999995" blue="0.97254901959999995" alpha="0.9186059397163121" colorSpace="calibratedRGB"/>
+                                <color key="backgroundColor" systemColor="tertiarySystemBackgroundColor"/>
```
vquach2404, line 33: I set the instructions view background color to `tertiarySystemBackgroundColor` because this view overlaps the map view; unlike other samples where the top of the map view is constrained to the bottom of the instructions view. I can update this sample so that the views do not overlap or update the others so that they do overlap. This is open for discussion as we should establish general rules for instruction views.

yo1995: In this case, please set the background view's background color to `statusLabelBackgroundColor` at top in the named colors section.

Also it is preferred to overlap the status label atop a map view or scene view. But it is up to you to make this change as it is trivial.

yo1995 (commented):
Overall looks great 👍  Please see my comment on status label background color below.

yo1995 (approved):
Looks good! Please replace the `line-of-sight-location.png` with a 276*597 screenshot.
Ready for others to review.

yo1995 (approved):

saratchandrakarumuri (commented):
Noticed these couple of things... Rest looks good

* Date field have different color gradients (may be specific to iOS 14)
IMG_7F0AB789B71E-1

* Can these sort images be swapped dynamically? They blend well in light mode but stand out in dark mode
IMG_FC4710D32A1A-1

vquach2404:
> * Date field have different color gradients (may be specific to iOS 14)

Yes, this `compact` style is specific to iOS 14. The background color and the constraints of the UI seem to be fixed. Now that you point it out, I think this style looks a bit odd here; especially since it won't span the entire row. I've changed it to the `wheels` style now which is what the style was originally before the update. I can revert it back to `compact` if necessary.Please let me know what you think! 

> * Can these sort images be swapped dynamically? They blend well in light mode but stand out in dark mode

Unfortunately, these images are pulled from the `Assets.xcassets` folder so I can't change the background color.

saratchandrakarumuri (dismissed):
`wheels` style looks good in dark mode and consistent with iOS 13.

philium:
> I've changed it to the `wheels` style now which is what the style was originally before the update.

@vquach2404 It looks like the screenshot needs to be updated as it doesn't reflect what the users will see. At least, it doesn't match what I'm seeing.

vquach2404:
@philium Thanks for catching that! I've updated the screenshot, please give another review.

yo1995 (approved):

saratchandrakarumuri (approved):
@philium I didn't realized that calender UI is in screenshots... Thanks for catching

philium (approved):

PR #999, opened 2020-11-12
Fix: Change view controller abbreviation

yo1995:
This PR fixes most remaining "VC" abbreviation in source code and replace them with "ViewController".


arcgis-ios-sdk-samples/Content Display Logic/Controllers/SegmentedViewController.swift
```diff
@@ -19,9 +19,6 @@ class SegmentedViewController: UIViewController {
     @IBOutlet private weak var infoContainerView: UIView!
     @IBOutlet private weak var codeContainerView: UIView!
     
-    private var sourceCodeVC: SourceCodeViewController!
-    private var sampleInfoVC: SampleInfoViewController!
```
yo1995, line 23: These 2 VC references are not used in current class, so I removed them.

philium (approved):

vquach2404 (approved):

PR #1016, opened 2020-12-01
Update Change basemap sample

vquach2404:
This PR updates the Change basemap sample to accommodate for the deprecated basemaps. Some basemaps do not have new basemap styles so I included a few of the new styles to supplement for the old ones. Please let me know what you think!


vquach2404:
Ready for a review! @yo1995

arcgis-ios-sdk-samples/Maps/Change basemap/SwitchBasemapViewController.swift
```diff
@@ -16,40 +16,37 @@ import UIKit
...
+        (.init(style: .arcGISLightGrayBase), "Light Gray Canvas (Raster)"),
+        (.init(style: .arcGISNavigation), "Navigation (Vector)"),
+        (.init(style: .arcGISNavigationNight), "Navigation Night (Vector)"),
+        (.init(style: .arcGISOceans), "Oceans (Raster)"),
+        (.init(style: .osmStreets), "OpenStreetMap (Raster)"),
+        (.init(style: .arcGISStreets), "Streets (Raster)"),
+        (.init(style: .arcGISStreetsNight), "Streets Night (Vector)"),
+        (.init(style: .osmStreetsRelief), "Streets w/ Relief (Vector)"),
+        (.init(style: .arcGISTerrain), "Terrain w/ Labels (Raster)"),
+        (.init(style: .arcGISTerrainDetail), "Terrain w/ Detail (Vector)"),
```
yo1995, line 40: It seems this style renders without terrain.

yo1995 (approved):
LGTM 👍  Feel free to add others to review.
p.s. I've changed the base branch to `basemap-updates` so that we can stash the basemap changes and merge later to `v.next`.

arcgis-ios-sdk-samples/Maps/Change basemap/SwitchBasemapViewController.swift
vquach2404: I was also unsure about this style, `.osmStreetsRelief`, and `.arcGISTopographic`. They are most likely server issues and I assume that would be fixed by the release so I kept them there.

philium: `arcGISTerrainDetail` is documented as "A vector basemap with minimal map content like populated place names, admin and water labels with boundary lines as the reference layer. Designed to be overlaid on `AGSBasemapStyleArcGISTerrainBase` and hillshade." Therefore I don't think it is a server issue.

vquach2404: @philium I see. I removed it from the sample. I also corrected the names of the basemaps. Please give this a review, thanks!

philium: Thanks! Since there's no longer vector/non-vector counterparts, perhaps we could just get rid of that distinction in the basemap names? That information is available in the documentation if desired.

vquach2404: Good point! I got rid of them. Ready for your review!

philium (approved):

saratchandrakarumuri (commented):
This sample is now prompting for credentials, which is expected as per Nimesh. But qt sampleviewer did some additional changes so that this sample won't prompt for login. Should the design be the same for iOS viewer too?

vquach2404:
> This sample is now prompting for credentials, which is expected as per Nimesh. But qt sampleviewer did some additional changes so that this sample won't prompt for login. Should the design be the same for iOS viewer too?

Hm are you referring to the addition of the API keys? After iOS finalizes support for the API keys, this sample will only request AGOL access if API keys have not been provided. This will take into affect after #1013 is merged.

saratchandrakarumuri:
@vquach2404 So should we update read me to mention about requirement of adding api key or authenticating? Because even with ting's branch the default behavior is not including api keys

yo1995:
> @vquach2404 So should we update read me to mention about requirement of adding api key or authenticating? Because even with ting's branch the default behavior is not including api keys

image

Oops there seems to be some miscommunication. 
Currently we are staging all the changes to basemap onto `basemap-updates` branch.
Originally I thought #1013 will get into `v.next` soon, and after it's done I'll merge `v.next` back into `basemap-updates` branch. It turns out that the timing has changed, so I'll update #1013 tomorrow to also include API keys. 
When we finally merge `basemap-updates` into `v.next`, we can do the test then.

vquach2404:
> When we finally merge `basemap-updates` into `v.next`, we can do the test then.

Are you saying we should hold off on testing this and instead just test the `basemap-updates` branch before merging into. `v.next`?

yo1995:
> Are you saying we should hold off on testing this and instead just test the `basemap-updates` branch before merging into. `v.next`?

IMO For now we can do some ad-hoc testing to verify that the sample work as normal; we may wait until merging to run the automated tests like image comparison and others.

saratchandrakarumuri (approved):
@vquach2404 In that case we can go ahead with this PR and worry about api keys when this going into `v.next`

PR #1017, opened 2020-12-01
Update: Create save map

yo1995:
This PR updates the `Create save map` sample with the new basemap APIs.

Note: Since the new basemap constructors require either signing in with a Named User, or an API key, the array in the table is changed to hold `AGSBasemapStyle`s.

Also, it seems the Named User authentication is not working with this sample. Not sure why but maybe related to the OAuth used in this sample... 😳  So please use API key to verify the changes.

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Create and save a map/CreateOptionsViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Create and save a map/CreateOptionsViewController.swift
@@ -33,7 +33,7 @@ class CreateOptionsViewController: UITableViewController {
     private var selectedLayerIndices: IndexSet = []
     
     weak var delegate: CreateOptionsViewControllerDelegate?
-
+    
     // MARK: - UITableViewDataSource
     
     override func numberOfSections(in tableView: UITableView) -> Int {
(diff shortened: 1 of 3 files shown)


vquach2404 (approved):
Looks good! I'd recommend updating the comments to sentence case while we are editing these files.

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateOptionsViewController.swift
```diff
@@ -20,7 +20,7 @@ protocol CreateOptionsViewControllerDelegate: AnyObject {
 }
 
 class CreateOptionsViewController: UITableViewController {
-    private let basemaps: [AGSBasemap] = [.streets(), .imagery(), .topographic(), .oceans()]
+    private let basemapStyles: [(name: String, style: AGSBasemapStyle)] = [("Streets", .arcGISStreets), ("Imagery", .arcGISImageryStandard), ("Topographic", .arcGISTopographic), ("Oceans", .arcGISOceans)]
```
philium, line 23: Could `KeyValuePairs` be used here?

yo1995:
Update: On Monday's planning it is mentioned that there is a bug related to the conflict between Named User and API key authentication in this (and a few other) samples.

- Running the sample without an API key: makes the basemap display incorrectly
- Running the sample with an API key: might lead to wrong saving behavior on AGOL

philium (approved):

PR #1020, opened 2020-12-02
Fix: Group layer grouped toggle

yo1995:
This issue was spotted by Sarat. The PR fixes the toggle that controls group visibility.

To review: build with `v.next` and this branch to see the difference.

|Before|After|
|-|-|
|before|after|

### DIFF
--- a/arcgis-ios-sdk-samples/Layers/Group layers/GroupLayersSectionView.xib
+++ b/arcgis-ios-sdk-samples/Layers/Group layers/GroupLayersSectionView.xib
@@ -1,11 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="14460.31" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
-    <device id="retina4_7" orientation="portrait">
-        <adaptation id="fullscreen"/>
-    </device>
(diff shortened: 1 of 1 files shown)


vquach2404 (approved):
Looks good but what an odd bug. Out of curiosity, do you know why the connection ID wasn't a random unique ID?

arcgis-ios-sdk-samples/Layers/Group layers/GroupLayersSectionView.xib
```diff
@@ -24,7 +22,7 @@
                 <switch opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="750" verticalHuggingPriority="750" contentHorizontalAlignment="center" contentVerticalAlignment="center" on="YES" translatesAutoresizingMaskIntoConstraints="NO" id="9ku-4w-xzb">
                     <rect key="frame" x="310" y="6.5" width="51" height="31"/>
                     <connections>
-                        <action selector="toggleSwitch:" destination="-2" eventType="valueChanged" id="ndi-hb-abW"/>
```
philium, line 27: The ol' destination -2 bug. Gets ya every time. 🙄

philium (approved):

arcgis-ios-sdk-samples/Layers/Group layers/GroupLayersSectionView.xib
yo1995: Ohh I'm also quite curious - I tried to search for it but couldn't find the right keyword. 😲  Is it a well-known bug?

philium: I've never heard of it before! 😂

saratchandrakarumuri (approved):
Toggle switches are now working good in the sample

PR #1022, opened 2020-12-10
Update: Basemap constructor updates

yo1995:
This PR merges the staging branch `basemap-updates` for basemap constructor changes to `v.next`.

Please use the secrets file and uncomment the license line in AppDelegate to test.

Known issues:
- `Create and save map` sample might not work as expected when saving the map to AGOL.


yo1995:
Will add reviewers after #1013 is merged.

philium:
All of these changes have already been reviewed, correct? So this review is more of a formality?

yo1995:
> All of these changes have already been reviewed, correct? So this review is more of a formality?

Yes, all the changes have been reviewed already.

philium (approved):

saratchandrakarumuri (commented):
Changes looks good, but I have couple of question

* There are couple of samples which are using older basemap style for national geographic map. I realized that there is no equivalent to that map in new style, but can we just use a different basemap instead of using one with old style?
* Looks like basemapstyle can be used with scenes too, but basemaps in none of the scene samples are updated. Is it happening in a different PR?

yo1995:
> * There are couple of samples which are using older basemap style for national geographic map. I realized that there is no equivalent to that map in new style, but can we just use a different basemap instead of using one with old style?

From what I learned the natGeo basemap style will be added in future releases, so Mike suggested to stick to it, until we have to replace it due to deprecation while it is somehow not added.

> * Looks like basemapstyle can be used with scenes too, but basemaps in none of the scene samples are updated. Is it happening in a different PR?

No. In U10 the basemapStyle constructors is not fully compatible with 3D scenes (some imagery ones are working, but other vector styles are not). That is also a (big?) reason why we revert the deprecation of old basemap APIs. @nixta could weigh in here.

saratchandrakarumuri (approved):
Basemap changes looks good

vquach2404 (approved):

nixta:
> No. In U10 the basemapStyle constructors is not fully compatible with 3D scenes (some imagery ones are working, but other vector styles are not). That is also a (big?) reason why we revert the deprecation of old basemap APIs. @nixta could weigh in here.

That's absolutely correct. Since all but 3 of the new basemap styles involve vector tile layers in one way or another, we cannot use them for scenes.

PR #1023, opened 2020-12-11
Fix: Remove credentials when leaving OAuth samples

yo1995:
This PR fixes the issue of

- common-samples/issues/1613

by removing credentials on deinit, the samples won't share credential cache;
~~by reset the presentation style to default, the OAuth webpage cannot be dismissed with iOS 13 interactive swipe (which hangs the auth process in an unknown state) on both iPhone and iPad~~ this will be fixed in the SDK

### DIFF
--- a/arcgis-ios-sdk-samples/Augmented reality/Navigate in AR/NavigateARRoutePlannerViewController.swift
+++ b/arcgis-ios-sdk-samples/Augmented reality/Navigate in AR/NavigateARRoutePlannerViewController.swift
@@ -202,6 +202,7 @@ class NavigateARRoutePlannerViewController: UIViewController {
     
     deinit {
         AGSAuthenticationManager.shared().oAuthConfigurations.remove(oAuthConfiguration)
+        AGSAuthenticationManager.shared().credentialCache.removeAllCredentials()
     }
 }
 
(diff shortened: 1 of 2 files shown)


vquach2404 (dismissed):

vquach2404 (approved):

arcgis-ios-sdk-samples/Augmented reality/Navigate in AR/NavigateARRoutePlannerViewController.swift
```diff
@@ -230,7 +231,6 @@ extension NavigateARRoutePlannerViewController: AGSGeoViewTouchDelegate {
 
 extension NavigateARRoutePlannerViewController: AGSAuthenticationManagerDelegate {
     func authenticationManager( _ authenticationManager: AGSAuthenticationManager, wantsToShow viewController: UIViewController) {
-        viewController.modalPresentationStyle = .overFullScreen
```
philium, line 233: Without this, how is the view controller presented on iPad? If I remember right, it presents fullscreen, which really does not look great.

yo1995: Right. There are 2 ways to fix this issue: 

1. Like the changes in this PR, we use the default fullScreen presentation style that doesn't allow dismiss by interactive swipe (dragging down)
2. Or we can set [isModalInPresentation](https://developer.apple.com/documentation/uikit/uiviewcontroller/3229894-ismodalinpresentation) to disable that in iOS 13+. But that will add a few other lines to check API availability. So I chose the first method even if it is a bit bad-looking.

philium: I actually think using `isModalInPresentation` on iOS 13+ is the way to go. It is more verbose, but I think it looks better (on screen) and shows users how they can do it in their own code.

vquach2404 (approved):

philium (approved):

PR #1024, opened 2020-12-11
Fix: Change download build phase output to avoid error

yo1995:
This PR proposes to fix the issue `/common-samples/issues/2004` that the project gives "Build input file cannot be found" error during the first build.

With this change, a user who downloads and builds the project for the first time shouldn't see the error anymore.

To test:  1. Xcode `Product` -> `clean build folder` to do a fresh build. 2. remove the existing Portal Data folder to ensure it doesn't effect our test

- `v.next` will give you the "Build input file cannot be found" error
- This PR will show 2 notifications without giving error.

### DIFF
--- a/arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
+++ b/arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
@@ -4594,6 +4594,7 @@
 			buildPhases = (
 				3E23A9DC1AFC28F6002E2214 /* Sources */,
 				3E23A9DD1AFC28F6002E2214 /* Frameworks */,
+				4CB506D1224C2C20006A3471 /* Download Portal Item Data */,
 				3E23A9DE1AFC28F6002E2214 /* Resources */,
 				3E03F0951B056DC40078EB36 /* CopyFiles */,
 				3ED029A21B8E5D4D00ACA70D /* ShellScript */,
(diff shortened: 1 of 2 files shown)


arcgis-ios-sdk-samples.xcodeproj/xcshareddata/xcschemes/arcgis-ios-sdk-samples.xcscheme
```diff
@@ -1,10 +1,28 @@
...
       buildImplicitDependencies = "YES">
+      <PreActions>
+         <ExecutionAction
+            ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
+            <ActionContent
+               title = "Download Portal Item Data"
+               scriptText = "/usr/bin/osascript -e &apos;display notification &quot;Build will start after download is complete&quot; with title &quot;Downloading Portal Data&quot; subtitle &quot;Prebuild&quot; sound name &quot;Funk&quot;&apos;&#10;&#10;PORTAL_ITEMS=&quot;${SRCROOT}/arcgis-ios-sdk-samples/Content Display Logic/PortalItems.plist&quot;&#10;FILE_TYPES=&quot;${SRCROOT}/arcgis-ios-sdk-samples/Content Display Logic/FileTypes.plist&quot;&#10;DOWNLOAD_DIRECTORY=&quot;${SRCROOT}/Portal Data&quot;&#10;xcrun --sdk macosx swift &quot;${SRCROOT}/Scripts/download-portal-item-data.swift&quot; &quot;$PORTAL_ITEMS&quot; &quot;$FILE_TYPES&quot; &quot;$DOWNLOAD_DIRECTORY&quot;&#10;&#10;/usr/bin/osascript -e &apos;display notification &quot;Project will start to build&quot; with title &quot;Download Complete&quot; subtitle &quot;Prebuild&quot; sound name &quot;Glass&quot;&apos;&#10;">
```
yo1995, line 13: This is the same script from the build phase, with 2 notifications added to indicate that the pre-action is in progress.

However, during the download process, user won't get additional information and may think the build process halt or run into problems. Probably we can instead send notifications when each file completes?

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -4594,7 +4594,6 @@
 			buildPhases = (
 				3E23A9DC1AFC28F6002E2214 /* Sources */,
 				3E23A9DD1AFC28F6002E2214 /* Frameworks */,
-				4CB506D1224C2C20006A3471 /* Download Portal Item Data */,
```
yo1995, line 4597: By removing the build phase, the output of download swift script is written to stdout/stderr instead, which means we cannot see it in the Xcode build log. We may need to find other ways to display these information, such as using notifications.

saratchandrakarumuri:
This is running the download portal data step not only before the build but also before the clean, is it supposed to happen?

yo1995:
> This is running the download portal data step not only before the build but also before the clean, is it supposed to happen?

I also see this side-effect, and it is [reported](https://stackoverflow.com/questions/19152267/xcode-5-pre-action-scripts-are-executed-during-project-cleaning) by others as well. One way to work around it is to add a line to check if it is a build rather than a Clean build folder action. I'll experiment to see if it is a good choice.

It only happens when we first manually remove the Portal Data folder, then run Clean Build Folder. It won't happen in the following cases:

- User downloads and builds for the first time
- User has built for 1/several times and decide to clean build folder, but not manually removing the Portal Data folder
- User remove the whole project and set it up again

philium:
I understand the bug is, well, a bug, but this seems like it might be a case of the solution being worse than the original problem itself. For a variety of reasons, some of which have already been mentioned, a build phase is a much more natural place for this script to run than a scheme pre-build action. I'll just mention another reason: as a build phase, the script is guaranteed to be run whenever the app is built. A part of a shared scheme, the script will only be run when that scheme is used. Users may have their own custom scheme(s) and by removing the build phase, the script will no longer be run for those schemes.

yo1995:
>  A part of a shared scheme, the script will only be run when that scheme is used. Users may have their own custom scheme(s) and by removing the build phase, the script will no longer be run for those schemes.

Yeah this makes sense. 🤭  I'll close this PR and revert the change to the schemes in the API keys PR.
What do you think if we instead print out to the build log a helpful message if a build phase fails "on purpose", i.e. the first build will fail because of missing `.secrets` file and no `Portal Data` items downloaded?

Thanks

philium:
> What do you think if we instead print out to the build log a helpful message if a build phase fails

That might be helpful. What did you have in mind? Create a PR or just send me a branch with your idea and I'll take a look.

yo1995:
Not an easy fix. Close for now.

PR #1025, opened 2020-12-12
Update: Spatial operations README

yo1995:
Sarat spotted that the README doesn't read well for the `difference` spatial operation.

Difference actually means the part which geometry1 differs from geometry2, or let's say all of geometry1 except the overlapping part of geometry2.


saratchandrakarumuri (approved):
Looks good

vquach2404 (approved):

PR #1028, opened 2021-01-21
[Linting] Lint comment for comment_spacing rule

yo1995:
In SwiftLint 0.42.0 the new `comment_spacing` rule (Prefer at least one space after slashes for comments) gives hundreds of warnings. This PR runs `swiftlint autocorrect` in the root directory to correct these issues.

Initially I tried to batch find and replace the comment strings, but it introduced too many unwanted changes. For example, when I replace `//set up the map` by finding `//s` and replace it to `// S`, it will also replace the string `https://sampleserver6...`. So I retreat to the built-in autocorrect tool in SwiftLint.

I don't know if we want to go with this for now, because it only addresses the spacing, but not capitalization or punctuation formatting. 

@philium can you weigh in here? We have a few options
- Don't make this fix, and temporarily disable the rule
- ✅  Apply this PR, and wait for future improvements to the comment strings
- Use this PR as a base branch, and incrementally fix all the comment manually


philium (approved):

vquach2404 (approved):
I think it's fine to address the spacing for now. We should make an effort to update the comment styles soon though.

PR #1031, opened 2021-01-27
[Fix] Slider default value bug in iOS 14

yo1995:
This PR fixes the bug in `common-samples/issues/2148` - slider doesn't honor the default value set in storyboard, in iOS 14.

image

There is no official bug reporting thread found on Apple forum, but the same issue was spotted by other developers. See section 2 in this post: https://fahimfarook.medium.com/xcode-12-and-ios-14-developer-bugs-and-issues-ada35920a104

This issue only affects iOS 14.

- [x] Generate offline map (overrides) > Max scale level slider
    Fixed by setting the slider's value to max in code
- [x] Line of sight (geoelement)
    Fixed by setting the default value to the height of ESB
- [x] Surface placements
    Fixed by setting the value to the half of min + max
- [x] ColorPickerViewController (display grid, change map background, etc.)
    - not changed, the expected behavior is set to 0 each time. Reset the values in storyboard.


saratchandrakarumuri (approved):
Verified on iOS 14 device and iOS 13 simulator and changes are good everywhere

philium (approved):
Just out of curiosity, were the storyboard changes necessary? They don't seem to be.

yo1995:
> Just out of curiosity, were the storyboard changes necessary? They don't seem to be.

No they are not must-have. Probably I was making a wrong move...

I changed the default value in storyboard so that the value is the same as what we set in the code.

Two reasons:

- if one day we want to remove the setter in code, the behavior of the sample won't change
- Sarat compared on device values against storyboard values, so I thought it would be good to keep them the same

philium:
Keep in mind that just because Xcode makes a bunch of silly changes to a storyboard file, that doesn't mean you have to commit them. Sometimes it is better to just commit the one line that was intentionally changed and ignore the rest.

PR #1036, opened 2021-02-19
Fix: Change reference to the raster used in raster cell sample to fix build error

yo1995:
This PR fixes the issue in

- `runtime/common-samples/issues/2004`
- `runtime/devops/issues/823`
- #1024 

by adjusting the on-demand raster used in `Identify raster cell` (#866) from `Create Groups` to `Create Folder References`, since the resource files come as a whole zip archive.

## How to review

- Do a fresh clone of the project from this [branch](https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/Try-ReplaceRaster) and do a clean build. 
- On the first build, `v.next` will give an `Build input file cannot be found: '$SRCROOT/Portal Data/Archives/SA_EVI_8Day_03May20/SA_EVI_8Day_03May20.tif'` error. The following builds won't give error since no additional portal item is added.
- This branch won't give any error

The error is hard to replicate because it only happens on the very first build of the project, or when the `Portal Data` folder is deleted or partially changed.

<details><summary>Below attach some details</summary>

1. Initially I thought this problem come when we made changes to the `download-portal-item-data.swift` script in #865 where we included better zip archive support for the download script.
2. Later we found out the problem is actually with the build system upgrade that [came with Xcode 10](http://www.gietal.net/blog/xcode10-and-prebuild-script-output-files-xcfilelist). In Xcode 10 we need to explicitly provide the build system a list of file paths, if those files are generated by one of the build phase. Otherwise the build system won't pick them up during the current build, which result in a `Build input file cannot be found` error.
3. In #1024 Ting tried to move the `Download Portal Item Data` from build phase to a `Pre-Action` in the build scheme. But Phil indicated that is even worse than having an one-time error.
4. Ting then tried to use `xcfilelist` to actually log the file paths downloaded. However, when going down this path, Ting found out the only files that cause this error are the rasters used in `Identify raster cell`. Adjusting their references and re-tag the on-demand resource tag solve the error.
5. Therefore, we are not adding additional patches to the project in case the problem is solved.

</details>

### DIFF
--- a/arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
+++ b/arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
@@ -46,6 +46,7 @@
 		008D175A24EF35800001BB8F /* EditWithBranchVersioningViewController.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = 008D175624EEF3FE0001BB8F /* EditWithBranchVersioningViewController.swift */; };
 		009D8BEF24F9AC8D00FD7E76 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 009D8BEE24F9AC8D00FD7E76 /* LaunchScreen.storyboard */; };
 		00A3766A256C932800EC433A /* AppSecrets.swift.masque in Sources */ = {isa = PBXBuildFile; fileRef = 007A4250256C7A400025F582 /* AppSecrets.swift.masque */; };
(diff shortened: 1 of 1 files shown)


saratchandrakarumuri (approved):
This fixes the build issue with download data

vquach2404 (approved):

philium (approved):
Nice! Thank you so much for investigating and fixing this bug.

PR #1037, opened 2021-02-23
New: Display device location with NMEA data sources

yo1995:
This PR implements the new `Display device location with NMEA data sources` sample for design at `common-samples/pull/2469`

The preview branch at [here](https://github.com/Esri/arcgis-runtime-samples-ios/blob/Ting/New-NMEA-vNext/arcgis-ios-sdk-samples/Route%20and%20directions/Display%20device%20location%20with%20NMEA%20data%20sources).

|Demo|
|-|
|demo|


arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/README.md
```diff
@@ -0,0 +1,39 @@
...
+* AGSLocation
+* AGSLocationDisplay
+* AGSNMEALocationDataSource
+* AGSNMEASatelliteInfo
+
+## About the data
+
+A list of NMEA sentences is used to initialize a `SimulatedNMEADataSource` object. This simulated data source provides NMEA data periodically, and allows the sample to be used on devices without a GPS dongle that produces NMEA data.
+
+The route taken in this sample features a 2-minute driving trip around Redlands, CA.
```
vquach2404, line 35: The data seems to come from the ArcGIS portal – could you link the data in this section?

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+        case .galileo:
+            return "The European Union Global Navigation Satellite System"
+        case .BDS:
+            return "The BeiDou Navigation Satellite System"
+        case .QZSS:
+            return "The Quasi-Zenith Satellite System"
+        case .navIC:
+            return "The Navigation Indian Constellation"
+        default:
+            return "Unknown GNSS type"
```
vquach2404, line 148: I'm not very familiar with NMEA. The GNSS type didn't change when I tested the sample. In what situation would it change?

vquach2404 (commented):
Looking good! It seems like these files aren't grouped in the sidebar. Could you fix this?
![Screen Shot 2021-02-24 at 3 16 27 PM](https://user-images.githubusercontent.com/52215269/109079105-4c85e900-76b3-11eb-9229-c841a1d1c4de.png]

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
yo1995: Since the NMEA sentences are pre-baked, all the satellite information are included in the sentences and all of the satellites were GPS. That is to say, in our sample with the mock data, they won't change.

In real world, it depends on what device to use and where the users are. Maybe some devices in Europe and Asia will use the different satellites such as Galileo, BeiDou, GLONASS, etc. 

I believe at least 90+% chances [people will get GPS](https://forums.quectel.com/t/nmea-active-satellites-and-beidou/6517/2), but I still included all the possible cases in case someone want to dig up the details about our NMEA support in Runtime.

vquach2404: Ah I see, thanks for the explanation! I agree, it'd be good to exemplify all cases 👍🏼

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/SimulatedNMEADataSource.swift
```diff
@@ -0,0 +1,104 @@
...
+    weak var delegate: SimulatedNMEADataSourceDelegate?
+    
+    /// Load locations from NMEA sentences.
+    /// Read mock NMEA sentences line by line and group them by the timestamp.
+    /// - Parameters:
+    ///   - nmeaSourceFile: The URL of the NMEA source file.
+    ///   - speed: The playback speed multiplier.
+    init(from nmeaSourceFile: URL, speed: Double = 1.0) {
+        // An empty container for NMEA data.
+        var dataSplittedBySecond = [Data]()
```
vquach2404, line 39:
```suggestion
        var dataSplitBySeconds = [Data]()
```
I think this is a bit more grammatically correct. Or maybe just `dataBySeconds`?

vquach2404 (approved):
Ready for Phil's review!

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+        nmeaLocationDataSource.push(nmeaData)
+    }
+}
+
+// MARK: AGSNMEALocationDataSourceDelegate
+
+extension DisplayDeviceLocationWithNMEADataSourcesViewController: AGSNMEALocationDataSourceDelegate {
+    func nmeaLocationDataSource(_ NMEALocationDataSource: AGSNMEALocationDataSource, satellitesDidChange satellites: [AGSNMEASatelliteInfo]) {
+        // Update the satellites info status text.
+        let satelliteSystemsText = Set(satellites.map { $0.system }).map { $0.label }.joined(separator: ", ")
```
philium, line 120: The end result of the second map is an array of unique strings. Instead of mapping twice, I think you could combine the two to just map once. Also, it would be a good idea to sort the strings so that you don't end up with multiple list items flip flopping.

```suggestion
        let satelliteSystemsText = Set(satellites.map(\.system.label))
            .sorted()
            .joined(separator: ", ")
```

I wish we could use `ListFormatter`. You could, but it isn't available on iOS 12, so you'd still need to use `joined(separator:)` for that.

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+        let idText = satellites.map { String($0.satelliteID) }.joined(separator: ", ")
+        let statusText =
+            """
+            \(satellites.count) satellites in view
+            System(s): \(satelliteSystemsText)
+            IDs: \(idText)
```
philium, line 126: Can there be only one ID?

```suggestion
            ID(s): \(idText)
```

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/README.md
```diff
@@ -0,0 +1,39 @@
...
+* AGSLocation
+* AGSLocationDisplay
+* AGSNMEALocationDataSource
+* AGSNMEASatelliteInfo
+
+## About the data
+
+A list of NMEA sentences is used to initialize a `SimulatedNMEADataSource` object. This simulated data source provides NMEA data periodically, and allows the sample to be used on devices without a GPS dongle that produces NMEA data.
+
+The route taken in this sample features a [2-minute driving trip around Redlands, CA](http://arcgis.com/home/item.html?id=d5bad9f4fee9483791e405880fb466da).
```
philium, line 35:
```suggestion
The route taken in this sample features a [2-minute driving trip around Redlands, CA](https://arcgis.com/home/item.html?id=d5bad9f4fee9483791e405880fb466da).
```

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/SimulatedNMEADataSource.swift
```diff
@@ -0,0 +1,104 @@
...
+    private var timer: Timer?
+    /// A delegate to notify view controllers.
+    weak var delegate: SimulatedNMEADataSourceDelegate?
+    
+    /// Load locations from NMEA sentences.
+    /// Read mock NMEA sentences line by line and group them by the timestamp.
+    /// - Parameters:
+    ///   - nmeaSourceFile: The URL of the NMEA source file.
+    ///   - speed: The playback speed multiplier.
+    init(from nmeaSourceFile: URL, speed: Double = 1.0) {
```
philium, line 37: FYI:

image

There are a few rare cases where Apple has violated that guideline, but I would recommend following unless you have a really good reason not to.

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/SimulatedNMEADataSource.swift
```diff
@@ -0,0 +1,104 @@
...
+        nmeaDataIterator = CircularIterator(elements: dataBySeconds)
+        playbackSpeed = speed
+    }
+    
+    func start() {
+        guard !nmeaDataIterator.elements.isEmpty else { return }
+        // Invalidate timer to stop previous mock data generation.
+        timer?.invalidate()
+        // Duration in second.
+        let duration = TimeInterval(1) / playbackSpeed
```
philium, line 67: This actually creates an integer first, and then a time interval (double) from that integer:

image

```suggestion
        let duration: TimeInterval = 1 / playbackSpeed
```

However, I would also suggest renaming `duration`. The value is used as the time interval of the timer, so `interval` might be a better name.

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSources.storyboard
```diff
@@ -0,0 +1,103 @@
...
+                                <viewLayoutGuide key="safeArea" id="AxU-Fm-eyW"/>
+                                <color key="backgroundColor" name="statusLabelBackgroudColor"/>
+                                <constraints>
+                                    <constraint firstItem="AuA-N5-5HY" firstAttribute="trailing" secondItem="8tL-4E-K67" secondAttribute="trailingMargin" id="KNU-0K-aM8"/>
+                                    <constraint firstItem="AuA-N5-5HY" firstAttribute="leading" secondItem="8tL-4E-K67" secondAttribute="leadingMargin" id="WqK-hR-BZJ"/>
+                                    <constraint firstItem="AuA-N5-5HY" firstAttribute="bottom" secondItem="8tL-4E-K67" secondAttribute="bottom" id="cEp-ea-ZbS"/>
+                                    <constraint firstItem="AuA-N5-5HY" firstAttribute="top" secondItem="8tL-4E-K67" secondAttribute="top" id="hGl-Zo-qSF"/>
+                                </constraints>
+                            </view>
+                            <toolbar opaque="NO" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="0Y8-SV-RLM">
```
philium, line 43: This view has duplicate bottom constraints:

image

Please remove one of them.

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        reset()
+    }
+}
+
+// MARK: SimulatedNMEADataSourceDelegate
+
+extension DisplayDeviceLocationWithNMEADataSourcesViewController: SimulatedNMEADataSourceDelegate {
+    func dataSource(didUpdate nmeaData: Data) {
```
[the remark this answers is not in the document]
yo1995: I see. I saw including self in a delegate call as a common practice, but didn't know why. Thanks for the explanation!

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
yo1995: Not sure about this. The NMEA [spec](http://aprs.gids.nl/nmea/#gsv) doesn't say the range of satellite count in the GSV messages. But my intuition says we'll need [at least 3 satellites](https://www.quora.com/What-is-the-minimum-number-of-satellites-a-GPS-system-requires-in-order-to-function) to locate a point on the earth surface... 🤔 

The mock data always shows 12.

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
yo1995: BTW I love the backslashdot-map here! 🤩 
Now I start to wonder if I could've used that in those previous samples...

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
philium: That makes sense.

arcgis-ios-sdk-samples/Route and directions/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -0,0 +1,155 @@
...
+        mapView.locationDisplay.start()
+        // Recenter the map and set pan mode.
+        recenter()
+    }
+    
+    @IBAction func recenter() {
+        mapView.locationDisplay.autoPanMode = .recenter
+        recenterBarButtonItem.isEnabled = false
+        mapView.locationDisplay.autoPanModeChangedHandler = { [weak self] _ in
+            DispatchQueue.main.async { [weak self] in
```
philium, line 63: `self` is already `weak` in this context.

```suggestion
            DispatchQueue.main.async {
```

philium (approved):

yo1995:
Thanks for the review! 👍  I'll add Sarat for verification. 
Could you give another approval as I applied the suggestion?

philium (approved):

saratchandrakarumuri (dismissed):
Sample looks good, but the only query I have is regular display location sample is in `Maps` but this is `routing` category. Is this how it is decided to be?

yo1995:
> Sample looks good, but the only query I have is regular display location sample is in `Maps` but this is `routing` category. Is this how it is decided to be?

🤦‍♂️ 🤦‍♂️  OMG good point... It should probably better go under `Maps` category.

Initially I put it under routing because it receives update from a certain location data source object (which is typical in a routing sample). But apparently the sample doesn't involve any route solving or direction updates...

I'll move the sample to Maps and request your review again.

saratchandrakarumuri (approved):
Looks good

philium (approved):

PR #1038, opened 2021-02-24
New: Create symbol styles from web styles

yo1995:
This PR implements the `Create symbol styles from web styles` sample with its design at `common-samples/pull/2560`

Link to the branch.

A demo below.

https://user-images.githubusercontent.com/9660181/108929767-7a0e5c00-75f9-11eb-9fa1-f367cb38c0a5.mp4


arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesLegendTableViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+class CreateSymbolStylesFromWebStylesLegendTableViewController: UITableViewController {
+    /// The symbols to display in the legend table.
+    var symbols: [(name: String, symbol: AGSSymbol)]?
```
vquach2404, line 20:
```suggestion
    var symbols = [(name: String, symbol: AGSSymbol)]()
```
Would it be possible to initialize the array like so to avoid optional chaining?

vquach2404 (commented):
Well done! I would suggest making the table cells unselectable.

vquach2404 (dismissed):
Feel free to add Phil!

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStyles.storyboard
```diff
@@ -0,0 +1,96 @@
...
+        <scene sceneID="h4J-H6-1A1">
+            <objects>
+                <tableViewController id="YKo-hR-7jI" customClass="CreateSymbolStylesFromWebStylesLegendTableViewController" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <tableView key="view" clipsSubviews="YES" contentMode="scaleToFill" alwaysBounceVertical="YES" dataMode="prototypes" style="plain" separatorStyle="default" rowHeight="-1" estimatedRowHeight="-1" sectionHeaderHeight="28" sectionFooterHeight="28" id="jHa-tn-JGS">
+                        <rect key="frame" x="0.0" y="0.0" width="414" height="842"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <color key="backgroundColor" systemColor="systemBackgroundColor"/>
+                        <prototypes>
+                            <tableViewCell clipsSubviews="YES" contentMode="scaleToFill" preservesSuperviewLayoutMargins="YES" selectionStyle="none" indentationWidth="10" reuseIdentifier="SymbolLegendCell" id="vTx-X9-l6a">
```
philium, line 68: The cell's style should be Basic (`IBUITableViewCellStyleDefault`), not Custom.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesLegendTableViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+        let cell = tableView.dequeueReusableCell(withIdentifier: "SymbolLegendCell", for: indexPath)
+        // Display the name of the symbol.
+        cell.textLabel?.text = symbols[indexPath.row].name
+        // Create an icon swatch for the symbol and update the cell.
+        symbols[indexPath.row].symbol.createSwatch { (image, _) in
+            if let updateCell = tableView.cellForRow(at: indexPath) {
+                updateCell.imageView?.image = image
+                updateCell.setNeedsLayout()
+            }
+        }
```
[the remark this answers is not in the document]
yo1995: The reason why I have them here is I copied from https://github.com/Esri/arcgis-runtime-samples-ios/blob/d58660d404c3bc757f4d3e0a235721d5d8e41211/arcgis-ios-sdk-samples/Display%20information/Show%20legend/MILLegendTableViewController.swift#L77-L82

because I tend to keep unrelated UI code consistent among samples. But in this particular case it doesn't seem to be a good practice.

I'll take your suggestions and re-implemented this section of code. ✅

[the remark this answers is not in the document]
philium: [`ReadSymbolsFromMobileStyleSymbolSettingsViewController`](https://github.com/Esri/arcgis-runtime-samples-ios/blob/d58660d404c3bc757f4d3e0a235721d5d8e41211/arcgis-ios-sdk-samples/Display%20information/Read%20symbols%20from%20a%20mobile%20style/ReadSymbolsFromMobileStyleSymbolSettingsViewController.swift#L183-L185) would make a much better reference for how to do asynchronous image loading for table view cells.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+    }
+    /// The button to show legend table.
+    @IBOutlet var legendBarButtonItem: UIBarButtonItem!
+    
+    // MARK: Instance properties
+    
+    /// The observation on the map view's `mapScale` property.
+    var mapScaleObservation: NSKeyValueObservation?
+    /// The symbols created from the symbol style to render the feature layer.
+    var pointSymbols = [(category: SymbolCategory, symbol: AGSSymbol)]()
```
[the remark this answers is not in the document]
yo1995: Good point. I forgot to sort the array after the dispatch group calls back.

I compared `Dict` vs `KeyValuePairs` vs `Array<Tuple>` and used the latter with a few reasons

- Since there is no need to look up a key in this sample, a linear array is sufficient
- Named tuple makes it more readable to access each element in a tuple
- Only need to sort once

Taken the other suggestions into consideration, I'll probably consider creating an observable array of symbols, and whenever it gets changed, re-sort the array and let the table observe the change.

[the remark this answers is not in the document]
yo1995: And it really makes me think of using [`UITableViewDiffableDataSource` ](https://developer.apple.com/documentation/uikit/uitableviewdiffabledatasource) in U11 with iOS 13, to handle these dynamic async tables. Will explore later.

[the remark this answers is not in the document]
philium: For cases like this, where the underlying data is only assigned once, we can use segue actions starting with U11. The would simplify things somewhat since we can make the underlying data immutable.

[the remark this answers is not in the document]
yo1995: That is nice! 🤩  Will try out soon.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesLegendTableViewController.swift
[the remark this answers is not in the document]
yo1995: Thanks for the suggestions and the example. I've extracted these methods into a data source class, so that

- Even if the `AGSSymbol`s get updated, the legends still reload/display correctly
- Get rid of the additional custom `Create...LegendTableViewController` class, since all we need is a plain table
- Load image swatches in the correct place and allow cancellation

I'm not sure if I'm making things too complex here though. Since there are quite a few changes, no hurry to review!

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,306 @@
...
+    }
+    
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        if segue.identifier == "LegendTableSegue",
+           let controller = segue.destination as? UITableViewController {
+            controller.presentationController?.delegate = self
+            controller.tableView.dataSource = symbolsDataSource
+            // Hold a weak reference to the table view to asynchronously reload
+            // image swatches for the symbols.
+            symbolsDataSource.tableView = controller.tableView
```
philium, line 102: `UITableViewDiffableDataSource` takes the table view as a parameter to the initializer and then automatically assigns itself as the table view's data source. You might consider doing something similar with `SymbolsDataSource.tableView` so that you don't need both of these.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,306 @@
...
+
+private class SymbolsDataSource: NSObject {
+    /// A weak reference to the table view which the data source itself connects.
+    weak var tableView: UITableView?
+    
+    /// The symbols created from a symbol style.
+    private var symbols = [(category: SymbolCategory, symbol: AGSSymbol)]() {
+        didSet {
+            // Keep the array sorted whenever it is modified.
+            symbols.sort { $0.category < $1.category }
```
philium, line 122: I would suggest that the data source should display the data that is given in the order in which it was given. Therefore I suggest doing the sorting before assigning to to `symbols`, not after. That way it is only mutated once, not twice.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,306 @@
...
+    
+    /// A cache for the image swatches of the symbols.
+    private var cachedImages = [IndexPath: UIImage]()
+    /// A cache for cancelable operations.
+    private var imageOperations = [IndexPath: AGSCancelable]()
+    
+    private func resetCaches() {
+        imageOperations.forEach { _, operation in
+            operation.cancel()
+        }
```
philium, line 136:
```suggestion
        imageOperations.values.forEach { $0.cancel() }
```

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,306 @@
...
+                    getSymbolsGroup.leave()
+                }
+                // Add the symbol to the result collection and ignore any error.
+                if let symbol = symbol {
+                    symbols.append((category, symbol))
+                }
+            }
+        }
+        getSymbolsGroup.notify(queue: .main) { [weak self] in
+            self?.symbols = symbols
```
philium, line 163: It is rather odd for a table view data source to be retrieving and setting its own data. I suggest moving this back to the sample view controller and then creating the data source with the symbols. That way, `SymbolsDataSource.symbols` could be immutable.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,306 @@
...
+            completion()
+        }
+    }
+    
+    /// Create an `AGSUniqueValueRenderer` to render feature layer with symbol styles.
+    /// - Parameter fieldNames: The attributes to match the unique values against.
+    /// - Returns: An `AGSUniqueValueRenderer` object.
+    func makeUniqueValueRenderer(fieldNames: [String]) -> AGSUniqueValueRenderer {
```
philium, line 171: The data source seems like a really odd place for this function as it has nothing to do with the table view. I would suggest moving it back to the sample view controller.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,306 @@
...
+            values = ["Public High Schools", "Public Elementary Schools", "Private and Charter Schools"]
+        case .trail:
+            values = ["Trails"]
+        }
+        return values
+    }
+    
+    static func < (lhs: SymbolCategory, rhs: SymbolCategory) -> Bool {
+        return lhs.symbolName < rhs.symbolName
+    }
```
philium, line 305: This is unnecessary as the compiler will [automatically synthesize](https://github.com/apple/swift-evolution/blob/main/proposals/0266-synthesized-comparable-for-enumerations.md) an implementation.

yo1995: 😮  Wow. Nice to learn!

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesLegendTableViewController.swift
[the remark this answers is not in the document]
yo1995: > Another good practice is to handle changes to the backing store. Right now, if symbols were to change after the table is loaded, nothing would happen. Adding a property observer that reloads the table view, if the view has already loaded, solves that problem.

Hi Phil, I'm not sure if I'm understanding correctly here. 😕 Could you explain a bit more?

Do you mean to pass the `symbols` as an `inout` parameter to the data source class's initializer, then observe the value; or add a property observer to the `symbols` in the view controller, and once it get changed, notify the data source class to reload the tableview?

Thanks.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+    
+    /// The observation on the map view's `mapScale` property.
+    var mapScaleObservation: NSKeyValueObservation?
+    /// The Esri 2D point symbol style created from a web style.
+    let symbolStyle = AGSSymbolStyle(styleName: "Esri2DPointSymbolsStyle", portal: .arcGISOnline(withLoginRequired: false))
+    
+    private var symbolsDataSource: SymbolsDataSource?
+    private var symbols = [(category: SymbolCategory, symbol: AGSSymbol)]() {
+        didSet {
+            symbolsDataSource?.symbols = symbols
```
yo1995, line 47: I'm trying to address the second comment here: https://github.com/Esri/arcgis-runtime-samples-ios/pull/1038#discussion_r583294741 about detecting potential changes to the symbols.

Initially I thought it is as easy as passing the array as a reference. But there are a few impediments. Therefore I came up with this rather not ideal solution. Please let me know the better way to do this!

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+private class SymbolsDataSource: NSObject {
+    /// A weak reference to the table view which the data source itself connects.
+    private weak var tableView: UITableView?
+    
+    /// The symbols created from a symbol style.
+    var symbols: [(category: SymbolCategory, symbol: AGSSymbol)] {
+        didSet {
+            // Reset caches to ensure legends are updated correctly.
+            resetCaches()
+            tableView?.reloadData()
```
yo1995, line 170: The diffable data source could achieve partial refresh easily, but here I simply reload the whole legend table when the symbols array is mutated.

yo1995 (commented):
The code is still WIP but I've a few questions. 😳 
Please let me know your thoughts and I'll clean up and finalize the code later.

Thank you!

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesLegendTableViewController.swift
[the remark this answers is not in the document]
philium: No, I meant using a property observer, like you've done.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+    var mapScaleObservation: NSKeyValueObservation?
+    /// The Esri 2D point symbol style created from a web style.
+    let symbolStyle = AGSSymbolStyle(styleName: "Esri2DPointSymbolsStyle", portal: .arcGISOnline(withLoginRequired: false))
+    
+    private var symbolsDataSource: SymbolsDataSource?
+    private var symbols = [(category: SymbolCategory, symbol: AGSSymbol)]() {
+        didSet {
+            symbolsDataSource?.symbols = symbols
+        }
+    }
```
philium, line 49: Okay, I think I might not have been very clear. If the data used _inside_ the data source (in this case an array of category and symbol pairs) is mutable, then the data source should be prepared to handle changes to that data. However, outside of the data source is a bit of a different story. Here, you know that this property will only be set once after it has been initialized, so this property observer will only be called once. You also know that the table view can only be displayed after this property has been updated, so when this property observer is called, `symbolsDataSource` will be `nil`, since the table view could not have been presented yet. Therefore this property observer isn't needed.

```suggestion
    private var symbols = [(category: SymbolCategory, symbol: AGSSymbol)]()
```

Note: Why treat the data source different? Because the data source doesn't (or shouldn't) know where the data comes from before it gets it. All it knows is that it has a mutable property, so it could be set from the outside world, so it should be prepared to handle changes to it. Does that make sense?

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+        super.viewDidDisappear(animated)
+        mapScaleObservation = nil
+    }
+    
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        if segue.identifier == "LegendTableSegue",
+           let controller = segue.destination as? UITableViewController {
+            controller.presentationController?.delegate = self
+            // The data source for the legend table display.
+            symbolsDataSource = SymbolsDataSource(tableView: controller.tableView, symbols: symbols)
```
philium, line 148: I realized that then the legend view controller is dismissed, the data source hangs around. You might want to add an unwind segue to set the `symbolsDataSource` to `nil` when the legend view controller goes away.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+    private weak var tableView: UITableView?
+    
+    /// The symbols created from a symbol style.
+    var symbols: [(category: SymbolCategory, symbol: AGSSymbol)] {
+        didSet {
+            // Reset caches to ensure legends are updated correctly.
+            resetCaches()
+            tableView?.reloadData()
+        }
+    }
```
philium, line 172: If you remove `CreateSymbolStylesFromWebStylesViewController.symbols`'s property observer, then this property can be made immutable:

```swift
let symbols: [(category: SymbolCategory, symbol: AGSSymbol)]
```

Makes `resetCaches()` unnecessary. This was made possible by changing to use a custom data source rather than a custom table view controller.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
yo1995: Yes that makes sense. It was the comment

> ... right now, if symbols were to change after the table is loaded, nothing would happen. 

that made me a bit confused. Because I thought you were suggesting that 

- the `SymbolsDataSource.symbols` should be mutable, so that whenever `CreateSymbolStylesFromWebStylesViewController.symbols` get updated, the table should also reload, even if the table is being presented
- even if in this sample the `CreateSymbolStylesFromWebStylesViewController.symbols` is only set once, we should still show how to handle async updates as good practices

With your new comments, now my understanding is that

- I'll make the `SymbolsDataSource.symbols` property immutable, and remove its property observers, since it is only initialized when the button is tapped.
- In this sample we don't reload the table when it is being presented

Does it sound correct? I'll make some changes and request your review again... 😅

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
yo1995: I added an unwind segue action to release the data source. 
However it would cause the icon swatches to be downloaded each time we open the legend table. Is that OK?

philium: Ah, I didn't think of that. In that case, I suppose it makes sense to keep the data source around. Either that or retrieve them _before_ creating the data source, in which case the data for each row might look something like this:

```swift
struct LegendItem {
    var name: String
    var image: UIImage
}
```

yo1995: Sounds good. I'll let you know once it's ready for review!

yo1995:
I've made changes based on suggestions, to cache the image swatches before allowing to show the legend table.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,295 @@
...
+        
+        categories.forEach { category in
+            getSymbolsGroup.enter()
+            symbolStyle.symbol(forKeys: [category.symbolName]) { symbol, _ in
+                defer {
+                    getSymbolsGroup.leave()
+                }
+                // Add the symbol to the result collection and ignore any error.
+                if let symbol = symbol {
+                    symbols.append((category, symbol))
```
philium, line 102: Instead stopping here and leaving the group, keep going and create the swatch for the symbol. Then once the swatch has been created, create an instance of legend item, append it to the array, and leave the dispatch group. That way there would be no need for an image cache. Is the symbol still needed for the `makeUniqueValueRenderer(fieldNames:)` method? Okay, add it to the legend item.

When refactoring code (like you just did in the last few commits), it is always a good idea to step back and re-evaluate what else can or should be changed. Take the image cache for example. It was only needed so that the table view data source could create an image once and then access it any number of items. The data source wasn't free to modify the array of symbols that was given to it in order to store the images there. But by moving creation of the swatches out of the data source, new possibilities have opened up. Does that make sense?

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,295 @@
...
+    ///   - symbolStyle: An `AGSSymbolStyle` object.
+    ///   - categories: The types of symbols to search in the symbol style.
+    ///   - completion: A closure executed upon success.
+    private func getSymbols(symbolStyle: AGSSymbolStyle, categories: [SymbolCategory], completion: @escaping () -> Void) {
+        let getSymbolsGroup = DispatchGroup()
+        var symbols = [(SymbolCategory, AGSSymbol)]()
+        
+        categories.forEach { category in
+            getSymbolsGroup.enter()
+            symbolStyle.symbol(forKeys: [category.symbolName]) { symbol, _ in
```
philium, line 96: `symbolName` is computed multiple times (here, in `makeUniqueValueRenderer(fieldNames:)`, etc.), but I think that can be avoided. Calculate it once by storing it to be local variable and then use the value of that variable here and, if you take my suggestion to eliminate the image cache, when creating the legend item. Then `makeUniqueValueRenderer(fieldNames:)` can use the legend item's name directly. Does that make sense?

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
yo1995: Yeah I totally agree 🤕  Last Friday when I was trying to refactor the code, I felt I'm losing track of some of the earlier suggestions.

yo1995:
I've applied the suggestions to refactor the code, removing excessive calls to the computed variables and adjusted the comment docstrings. Hope the changes make it clearer! 🙏

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,282 @@
...
+    /// The Esri 2D point symbol style created from a web style.
+    let symbolStyle = AGSSymbolStyle(styleName: "Esri2DPointSymbolsStyle", portal: .arcGISOnline(withLoginRequired: false))
+    /// The observation on the map view's `mapScale` property.
+    var mapScaleObservation: NSKeyValueObservation?
+    /// The data source for the legend table.
+    private var symbolsDataSource: SymbolsDataSource?
+    /// A list of symbols created from the web style, with their associated names and values.
+    private var symbols = [(symbolName: String, symbol: AGSSymbol, symbolCategoryValues: [String])]()
+    /// The legend items for legend table.
+    private var legendItems = [LegendItem]()
```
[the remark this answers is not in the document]
yo1995: I agree. I was hoping to cram all the required properties into a single struct, to avoid holding stuff piece by piece, such as

```swift
private struct LegendItem {
    let symbolName: String
    let image: UIImage
    let symbol: AGSSymbol
    let symbolCategoryValues: [String]
}
```

But in this case the name of the struct `LegendItem` doesn't seem proper. Do you have any suggestion on this?
Thanks.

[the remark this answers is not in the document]
philium: Yeah, I played around with trying to combine them as well and I didn't like what I was coming up with. But I just realized that the `symbols` property isn't really needed either. It is only read by `makeUniqueValueRenderer(fieldNames:)`, which is called by the completion handler for `getSymbols(symbolStyle:types:completion:)`. You could modify `makeUniqueValueRenderer(fieldNames:)` to take the symbols as an argument. Then, you could either modify the completion handler to pass back the symbols or move the call to `makeUniqueValueRenderer(fieldNames:)` to where the completion handler is called. If you do that, then you could move `legendBarButtonItem.isEnabled = true` to a property observer of `symbolsDataSource` and get rid of the completion handler altogether. Does that make sense?

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -56,28 +56,27 @@ class CreateSymbolStylesFromWebStylesViewController: UIViewController, UIAdaptiv
...
-    /// Create a map with an `AGSFeatureLayer` added to its operational layers.
-    /// - Parameter featureLayer: An `AGSFeatureLayer` object.
-    /// - Returns: An `AGSMap` object.
-    func makeMap(featureLayer: AGSFeatureLayer) -> AGSMap {
+    /// Create a map with a reference scale.
+    func makeMap(referenceScale: Double) -> AGSMap {
         let map = AGSMap(basemapStyle: .arcGISLightGray)
-        map.referenceScale = 1e5
-        map.operationalLayers.add(featureLayer)
+        map.referenceScale = referenceScale
```
yo1995: I realized that if I add the feature layer before setting its unique value renderer, the layer may load with its default renderer/symbols first (displays as black dots). So instead I add the feature layer to the map *after* setting its renderer.

This afternoon I'll discuss with Divesh to learn a bit more about how feature layer and its renderer works.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -56,28 +56,27 @@ class CreateSymbolStylesFromWebStylesViewController: UIViewController, UIAdaptiv
...
     }
     
     /// Create an `AGSUniqueValueRenderer` to render feature layer with symbol styles.
-    /// - Parameter fieldNames: The attributes to match the unique values against.
+    /// - Parameters:
+    ///   - fieldNames: The attributes to match the unique values against.
+    ///   - symbolInfos: A dictionary of symbols and their associated infos.
     /// - Returns: An `AGSUniqueValueRenderer` object.
-    func makeUniqueValueRenderer(fieldNames: [String]) -> AGSUniqueValueRenderer {
+    func makeUniqueValueRenderer(fieldNames: [String], symbolInfos: [AGSSymbol: (name: String, values: [String])]) -> AGSUniqueValueRenderer {
```
yo1995, line 71: Passed an argument to the method of a dict of symbols and infos. A few reasonings

- Instead of directly using the element-wise initializer of `AGSUniqueValueRenderer` in the `DispatchGroup.notify` block, I created this method because I think it is important to separately showcase how we map the symbols' names and category values to create unique values.
- To avoid large tuples, I used dictionary to pass those computed strings
- Naming: I'm thinking either `symbolAttributes` or `symbolInfos` and chose the latter.

though I'm still not quite satisfied. Love to hear more feedback!

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -87,36 +86,40 @@ class CreateSymbolStylesFromWebStylesViewController: UIViewController, UIAdaptiv
...
-                    }
+                defer { getSymbolsGroup.leave() }
+                // Add the symbol to result dictionary and ignore any error.
+                guard let symbol = symbol else { return }
+                symbolInfos[symbol] = (symbolName, symbolCategoryValues)
+                
+                // Get the image swatch for the symbol.
+                getSymbolsGroup.enter()
+                symbol.createSwatch { image, _ in
+                    defer { getSymbolsGroup.leave() }
```
yo1995, line 108: Enter the dispatch group twice for each symbol, to ensure we don't get stuck if any error occurs.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -166,11 +162,6 @@ class CreateSymbolStylesFromWebStylesViewController: UIViewController, UIAdaptiv
         // Ensure that the settings are shown in a popover on small displays.
         return .none
     }
-    
-    func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
-        // Release the data source when the table view controller is dismissed.
-        symbolsDataSource = nil
```
yo1995: Don't release the symbols data source as we are using it to hold the legend items.

yo1995 (commented):
Based on the suggestions, I made a few more changes. 

I'll request your review later when it's ready.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
```diff
@@ -0,0 +1,273 @@
...
+            )
+        }
+    }
+    /// The button to show legend table.
+    @IBOutlet var legendBarButtonItem: UIBarButtonItem!
+    
+    // MARK: Instance properties
+    
+    /// The Esri 2D point symbol style created from a web style.
+    let symbolStyle = AGSSymbolStyle(styleName: "Esri2DPointSymbolsStyle", portal: .arcGISOnline(withLoginRequired: false))
```
philium, line 40: This doesn't need to be a property either. I can just be created in `getSymbols` since it is only used there, right?

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
philium: I think what you have here is fine. Another name idea is `symbolDetails`.

philium (commented):
Looking good! I think this PR is almost ready.

arcgis-ios-sdk-samples/Display information/Create symbol styles from web styles/CreateSymbolStylesFromWebStylesViewController.swift
yo1995: I'm not really sure why, but the `AGSSymbolStyle.symbol(forKeys:completion:)` method doesn't callback if I don't hold a strong reference to the `symbolStyle`, e.g. if I create a local variable in the `viewDidLoad` method. 🤔

philium: Oh, duh. I knew that. Yes, the property is needed to prevent the object from being deallocated while fetching the symbols. Never mind.

philium (approved):

saratchandrakarumuri (approved):
Sample looks good

PR #1026, opened 2020-12-16
[iOS] Update Sample Server 7 to 10.8.1

vquach2404:
This PR updates samples that use sample server 7. The URLs have been updated and credentials have been added. For testing purposes, you will need to be on the VPN and update the hosts file on your machine. Please see `/common-samples/issues/2270` for more info and see Zack's comment about updating the hosts file. 

Note: The Naperville Electric feature service has changed slightly but will be updated to match its previous state. Display subtype feature layer, Display utility associations, and trace utility network may appear different as a result.

PR #858, opened 2020-04-29
Perform valve isolation trace

yo1995:
Implementation of the `Perform valve isolation trace` sample. 

A link to the PR branch folder: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/PerformValveIsolationTrace_vNext/arcgis-ios-sdk-samples/Utility%20Network/Perform%20valve%20isolation%20trace

A brief demo below.

|`Isolated features included`|`Isolated features excluded`|
|:------:|:------:|
|||

PR #1039, opened 2021-02-24
Update utility network samples to use service geodatabase

vquach2404:
This PR updates utility network samples so that the maps use service geodatabases. Please see `/common-samples/2410` for more info.


arcgis-ios-sdk-samples/Utility network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -83,17 +54,54 @@ class TraceUtilityNetworkViewController: UIViewController, AGSGeoViewTouchDelega
...
+                            defaultSymbol: AGSSimpleLineSymbol()
+                        )
+                    }
+                    return layer
+                }
+            }()
+            // Add the utility network feature layers to the map for display.
+            self.map.operationalLayers.addObjects(from: self.layers)
+        }
+
```
yo1995, line 104:
```suggestion
```
Remove blank lines, linter suggestion.

arcgis-ios-sdk-samples/Utility network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -83,17 +54,54 @@ class TraceUtilityNetworkViewController: UIViewController, AGSGeoViewTouchDelega
...
-        
+        // Create the service geodatabase that matches the utility network.
+        serviceGeodatabase = AGSServiceGeodatabase(url: featureServiceURL)
         super.init(coder: aDecoder)
+        // Load the service geodatabase.
+        serviceGeodatabase.load { [weak self] error in
+            guard let self = self else { return }
+            // Create electrical distribution line layer ./115 and electrical device layer ./100.
+            self.layers = {
+                return [115, 100].map {
```
yo1995, line 72: There were a suggestion by Phil not to use the path component concatenation pattern: https://github.com/Esri/arcgis-runtime-samples-ios/pull/858#discussion_r418925647

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -104,6 +91,31 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
     }
     
+    /// Load the service geodatabase and initialize the layers.
+    func loadServiceGeodatabase() {
+        serviceGeodatabase = AGSServiceGeodatabase(url: featureServiceURL)
+        serviceGeodatabase.load { [ weak self ] error in
+            guard let self = self else { return }
+            // The gas line layer ./3 and gas device layer ./0 are created from the service geodatabase.
+            self.layers = {
+                let layerIDs = [
```
yo1995, line 101: Same here: https://github.com/Esri/arcgis-runtime-samples-ios/pull/858#discussion_r418925647 we may just directly create an array with hardcoded values.

arcgis-ios-sdk-samples/Utility network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -83,17 +54,54 @@ class TraceUtilityNetworkViewController: UIViewController, AGSGeoViewTouchDelega
...
         map = AGSMap(basemapStyle: .arcGISStreetsNight)
+        
         // Create the utility network, referencing the map.
         utilityNetwork = AGSUtilityNetwork(url: featureServiceURL, map: map)
-        
+        // Create the service geodatabase that matches the utility network.
+        serviceGeodatabase = AGSServiceGeodatabase(url: featureServiceURL)
         super.init(coder: aDecoder)
+        // Load the service geodatabase.
+        serviceGeodatabase.load { [weak self] error in
```
yo1995, line 68:
```suggestion
        serviceGeodatabase.load { [weak self] _ in
```
Since the error is not checked here, replace it with underscore.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -104,6 +91,31 @@ class PerformValveIsolationTraceViewController: UIViewController {
         return parameters
     }
     
+    /// Load the service geodatabase and initialize the layers.
+    func loadServiceGeodatabase() {
+        serviceGeodatabase = AGSServiceGeodatabase(url: featureServiceURL)
+        serviceGeodatabase.load { [ weak self ] error in
```
yo1995, line 97:
```suggestion
        serviceGeodatabase.load { [weak self] error in
```
Also what about moving the error checking to the top of this load method? It seems odd to check error at the end. Though I'm not sure what error would be emitted other than network or load error.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -54,30 +50,21 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
     /// - Parameter layers: The feature layers for the utility network.
-    /// - Returns: An `AGSMap` object.
-    func makeMap(with layers: [AGSFeatureLayer]) -> AGSMap {
+    func setMap(with layers: [AGSFeatureLayer]) {
         let map = AGSMap(basemapStyle: .arcGISStreetsNight)
         // Add the utility network feature layers to the map for display.
         map.operationalLayers.addObjects(from: layers)
-        return map
+        // Create the utility network, referencing the map.
+        utilityNetwork = AGSUtilityNetwork(url: featureServiceURL, map: map)
```
yo1995, line 66: Hmm it seems to be an odd place to initialize the `utilityNetwork` property... Leave it to Phil's judgement.

yo1995 (commented):
The changes look good, and I cannot really spot any difference with ad-hoc testing - which is great! Also, great job for paving the roads for other platforms! 👍 

In addition to the comments below, a few more thoughts

- Since we still don't know when SS7 will go live, it would be painful to try to keep both PRs (#1039 and #1026 ) up-to-date. So once this PR is approved, feel free to put it aside until we revisit it, and solve the merge conflict then. 
- And we may give another approval after testing with the new service URLs.
- It seems without using SS7 URLs, these changes also work without issue... Is that expected? i.e. I can still do the trace in both TUN and PVIT samples.

vquach2404:
> * Since we still don't know when SS7 will go live, it would be painful to try to keep both PRs (#1039 and #1026 ) up-to-date. So once this PR is approved, feel free to put it aside until we revisit it, and solve the merge conflict then.
Yes, I agree. I'll wait until SS7 goes live and update this PR accordingly.
> * It seems without using SS7 URLs, these changes also work without issue... Is that expected? i.e. I can still do the trace in both TUN and PVIT samples.
Oh I see. My hosts file is still edited to accommodate for the SS7 updates. If your hosts file is still edited like mine, you'll have to remove the line in order to test.

yo1995 (commented):
LGTM 👍  just one last comment.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -104,6 +91,30 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
+            guard let self = self else { return }
+            // The gas line layer ./3 and gas device layer ./0 are created from the service geodatabase.
+            self.layers = {
+                let featureTables = [
+                    self.serviceGeodatabase.table(withLayerID: 3),
+                    self.serviceGeodatabase.table(withLayerID: 0)
+                ]
+                return featureTables.map { (featureTable) -> AGSFeatureLayer in
+                    return AGSFeatureLayer(featureTable: featureTable!)
+                }
```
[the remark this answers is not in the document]
vquach2404: That's a good idea! I made a similar change. Please let me know what you think!

yo1995 (dismissed):

arcgis-ios-sdk-samples/Utility network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -27,38 +27,9 @@ class TraceUtilityNetworkViewController: UIViewController, AGSGeoViewTouchDelega
...
-                    uniqueValues: [mediumVoltageValue, lowVoltageValue],
-                    defaultLabel: "",
-                    defaultSymbol: AGSSimpleLineSymbol()
-                )
-            }
-            return layer
-        }
-    }
-    
+    private var layers = [AGSFeatureLayer]()
```
philium, line 30: It looks like this property isn't even needed as it is only used in `init(coder:)`.

vquach2404:
I've marked this as a draft for now until we can test with the new SS7 updates.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
vquach2404: I forwarded this question to Jennifer N. from the utility networks team:

> Yup that is correct, using UN constructor with map parameter will add this UN to the map's UtilityNetworks[]collection. If you use UN constructor with just URL, before loading it, you can add to map.UtilityNetworks property.. but once loaded, you cannot add to a map anymore. This map and UN relationship allows the tables on map to be used by UN - sharing the same instance.

vquach2404:
SS7 updates have gone live! This PR is now ready for reviews. You no longer have to edit your hosts file in order to test. 
Please note that Create load report and Configure subnetwork trace do not need service geodatabases since they don't have maps.

yo1995 (dismissed):

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
philium: Thanks!

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -104,6 +90,24 @@ class PerformValveIsolationTraceViewController: UIViewController {
         return parameters
     }
     
+    /// Load the service geodatabase and initialize the layers.
+    func loadServiceGeodatabase() {
+        let serviceGeodatabase = AGSServiceGeodatabase(url: featureServiceURL)
```
philium, line 95: Methinks this would need to be a property rather than an instance variable as the way it is now, only the completion handler is keeping the geodatabase around, which isn't great. The completion handler should access the geodatabase either through a property of `self` or as an unowned reference.

vquach2404: This change required me to adjust the way the credentials are handled. I also added an additional status message because there were times when the viewpoint took a while to adjust and the status bar said "utility network loaded" which was confusing.

yo1995: I'm a bit confused here - why the credentials handling need to be changed? Could we still apply the credentials to the UN or even to the geodatabase? `serviceGeodatabase.credential = ...`

yo1995 (commented):
Please see my comment above.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
vquach2404: By making the `serviceGeodatabase` a property, authentication is prompted twice – once for `serviceGeodatabase` and then again for `utilityNetwork`. Using the `AGSAuthenticationManager` allows me to set the credentials once and it will apply the credentials whenever prompted. Otherwise, I'd have to set the same credentials twice for separate instance variables.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -104,10 +91,28 @@ class PerformValveIsolationTraceViewController: UIViewController {
         return parameters
     }
     
+    /// Load the service geodatabase and initialize the layers.
+    func loadServiceGeodatabase() {
+        serviceGeodatabase = AGSServiceGeodatabase(url: featureServiceURL)
```
philium, line 96: Instead of creating the geodatabase here (and needing it to be mutable and optional), could it be created where the property is declared?

```swift
let serviceGeodatabase = AGSServiceGeodatabase(url: featureServiceURL)
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -104,10 +91,28 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
+            }
+        }
+    }
+    
+    /// Load the utility network.
     func loadUtilityNetwork() {
         setStatus(message: "Loading utility network…")
         // NOTE: Never hardcode login information in a production application. This is done solely for the sake of the sample.
-        utilityNetwork.credential = AGSCredential(user: "viewer01", password: "I68VGU^nMurF")
+//        utilityNetwork.credential = AGSCredential(user: "viewer01", password: "I68VGU^nMurF")
```
philium, line 115:
```suggestion
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
yo1995: I think the problem is `featureServiceURL` is a property, which makes the `serviceGeodatabase` cannot access it in the first place.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
yo1995: I tried to set the credential on geodatabase and not the UN, and it doesn't prompt me for credentials. This is why I asked...

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
philium: Sounds like `featureServiceURL` should be `static`.

yo1995: ✅  I think this is the right time to pitch the question - should we

- Create all the string/URL literals in a sample as static constants?
- Only create the strings that are required by other properties (like this URL) as static, and other strings as instance property?
- Avoid using static variables, and duplicate the strings when they are required to instantiate an object?

We may agree on a pattern so that we don't have these issues in the future.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -53,31 +49,22 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
+    var layers = [AGSFeatureLayer]()
     
     // MARK: Initialize map and utility network
     
     /// Create a map.
     ///
     /// - Parameter layers: The feature layers for the utility network.
-    /// - Returns: An `AGSMap` object.
-    func makeMap(with layers: [AGSFeatureLayer]) -> AGSMap {
+    func setMap(with layers: [AGSFeatureLayer]) {
```
philium, line 61: Now that creation of the utility network doesn't depend on the map, could we go back to using the `makeMap` method?

yo1995: And since both the `utilityNetwork` and `layers` are properties, probably we can remove the parameters from this method.

arcgis-ios-sdk-samples/Utility network/Trace utility network/TraceUtilityNetworkViewController.swift
```diff
@@ -83,16 +53,54 @@ class TraceUtilityNetworkViewController: UIViewController, AGSGeoViewTouchDelega
...
     }()
     
-    // MARK: Initialize map and Utility Network
+    // MARK: Initialize map, utility network, and service geodatabase
     required init?(coder aDecoder: NSCoder) {
         // Create the map
         map = AGSMap(basemapStyle: .arcGISStreetsNight)
+        
         // Create the utility network, referencing the map.
         utilityNetwork = AGSUtilityNetwork(url: featureServiceURL, map: map)
```
yo1995, line 62: Should we also update this to `map.utilityNetworks.add(utilityNetwork)`?

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
vquach2404: Hm, I could use the `makeMap` method but would have to remove the layers parameter since the service geodatabase must be loaded before I can instantiate the layers array. Please let me know what you think!

philium (dismissed):

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
philium: Yeah, I guess it makes sense to remove the layers parameter. Since the layers need to be added later, it makes sense to do that through the `map` property of the map view, just like you did.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
philium: I don't know that there's a one-size-fits-all pattern we can follow. In this case it made sense for the URL to be static, but sometimes the URL is just used in one place, in which case it doesn't even need to be a property.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -53,30 +53,20 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
     /// - Parameter layers: The feature layers for the utility network.
-    /// - Returns: An `AGSMap` object.
-    func makeMap(with layers: [AGSFeatureLayer]) -> AGSMap {
+    func makeMap() -> AGSMap {
         let map = AGSMap(basemapStyle: .arcGISStreetsNight)
-        // Add the utility network feature layers to the map for display.
-        map.operationalLayers.addObjects(from: layers)
+        // Add the utility network to the map's array of utility networks.
+        map.utilityNetworks.add(utilityNetwork)
+        mapView.map = map
```
yo1995, line 69:
```suggestion
```

This line seems to be redundant.

yo1995 (commented):
Just one more thing

yo1995 (approved):

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
philium: Thank you for catching that!

philium (approved):

saratchandrakarumuri (approved):

PR #1040, opened 2021-02-24
Fix: blurry basemaps in scene samples

yo1995:
This PR fixes the blurred basemaps in various samples that use the Brest, France coastline area, which became a navy base and pixelated/censored from the imagery basemap. issue at `common-samples/issues/2253`

| Affected sample | Action |
| -- | -- |
| Distance measurement analysis| update to use topographic map |
| Scene layer selection | update to use topographic map |
| Viewshed camera| update to Girona, Spain integrated mesh layer |
| Scene layer (URL)| no action needed, uses a topographic basemap |

image


vquach2404 (approved):
LGTM!

saratchandrakarumuri:
@yo1995 Not directly related to change addressed in this PR. But since we are here can you update the brest scene layer used in `Distance measurement analysis` to the one used in `Scene layer (URL)`. Layer used distance sample has lot of bleeding and showing basemap which doesn't look good 

IMG_78B664648D63-1

philium (approved):

yo1995:
> ...can you update the brest scene layer used in `Distance measurement analysis` to the one used in `Scene layer (URL)`. Layer used distance sample has lot of bleeding and showing basemap which doesn't look good

https://github.com/Esri/arcgis-runtime-samples-ios/blob/6d5ed31c11f03f6962b2c7527e117871fa75f190/arcgis-ios-sdk-samples/Analysis/Distance%20measurement%20analysis/DistanceMeasurementAnalysisViewController.swift#L38-L47

@philium In the `Distance measurement analysis` sample there are 2 elevation sources added. Do you remember why we need both of them, and if we need to remove one, which one should I remove? Thanks.

I tried to remove the Brest elevation source - the flakiness gets better, but not totally resolved. My guess is that the accuracy of elevation caused this issue.

philium:
@yo1995 I do not know/remember why we used two elevation source. Presumably because it was specified in the design (if there is one) or because the reference implementation I consulted used two.

arcgis-ios-sdk-samples/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -31,23 +31,18 @@ class DistanceMeasurementAnalysisViewController: UIViewController, AGSGeoViewTou
...
         // Create the scene.
-        scene = AGSScene(basemap: .imagery())
+        scene = AGSScene(basemap: .topographic())
         
-        /// The url of the Terrain 3D ArcGIS REST Service.
-        let worldElevationServiceURL = URL(string: "https://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer")!
-        /// The url of the image service for elevation in Brest, France.
-        let brestElevationServiceURL = URL(string: "https://scene.arcgis.com/arcgis/rest/services/BREST_DTM_1M/ImageServer")!
+        // The elevation image service URL.
+        let elevationServiceURL = URL(string: "https://tiles.arcgis.com/tiles/d3voDfTFbHOCRwVR/arcgis/rest/services/MNT_IDF/ImageServer")!
```
yo1995, line 37: According to Tanner, we substitute the elevation source with this service. I cannot really tell if the surface is elevated 🤔  - the Brest area is pretty flat as for this elevation source. But it does fixed the flaky-looking buildings in the scene.

saratchandrakarumuri:
With this updated code, elevation is only applied to scene layer but not the scene itself and layer is rendered floating mid air

IMG_76C47025B2AF-1

yo1995:
> With this updated code, elevation is only applied to scene layer but not the scene itself and layer is rendered floating mid air

Confirmed. So I really wonder if it is possible to correctly elevate these buildings. I'll wait for more feedback on this topic in the channel!

arcgis-ios-sdk-samples/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -31,26 +31,23 @@ class DistanceMeasurementAnalysisViewController: UIViewController, AGSGeoViewTou
...
+        surface.elevationSources.append(elevationSource)
         scene.baseSurface = surface
         
-        /// The url of the scene service for buildings in Brest, France.
+        // The URL of the scene service for buildings in Brest, France.
         let brestBuildingsServiceURL = URL(string: "https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/Buildings_Brest/SceneServer/layers/0")!
         // Create the building layer and add it to the scene.
         let buildingsLayer = AGSArcGISSceneLayer(url: brestBuildingsServiceURL)
+        // Offset the altitude to avoid clipping with the elevation source.
+        buildingsLayer.altitudeOffset = 2
```
yo1995, line 50: Last week most people agreed to add an offset to the buildings layer to mitigate the clipping between elevated surface and buildings. Once the design get finalized, I'll request another review from you all.

yo1995:
An 2m elevation offset has been added to the `DistanceMeasurementAnalysisViewController` so the scene layer doesn't clip into the terrain. Please feel free to take another look!

saratchandrakarumuri (approved):
This change looks good

vquach2404 (approved):
Changes look good!

PR #1043, opened 2021-03-05
New: Create load report

yo1995:
This PR adds the new `Create load report` sample in Utility network category. The issue on DT: `common-samples/issues/2594`

Link to the branch: [here](https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/Fix-SceneBasemap-vNext/arcgis-ios-sdk-samples/Utility%20network/Create%20load%20report). Apologies for the wrong branch name - it was too late when I realized that I wasn't on the right branch 😅


arcgis-ios-sdk-samples/Utility network/Create load report/README.md
```diff
@@ -0,0 +1,60 @@
+# Create load report
+
+Create a simple electric distribution report that adds up the count of customers and total load per phase by tracing downstream from a given point.
+
+![Image of create load report](create-load-report.jpg)
```
vquach2404, line 5:
```suggestion
![Image of create load report](create-load-report.png)
```

vquach2404 (commented):
A few points to discuss: 

- Is it possible to individually delete some of the phases? Maybe it can be similar to Configure subnetwork trace.
- What do you think about adding a default phase to act as an example of what the user should expect. If swipe to delete can be enabled, the user could easily delete the default phase.
- I know we discussed the title issue during the review but it seems a bit odd to have no title at all. What are your thoughts about adding the title "Phases, total customers, total load" or "Phase, customers per phase, load per phase"? That way, it provides a bit more context for the user and can help with the issue of displaying the units.

yo1995:
Thanks for the suggestions. I agree with and applied all of them and updated the screenshot. Please take another look!

arcgis-ios-sdk-samples/Utility network/Create load report/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+![Image of create load report](create-load-report.png)
+
+## Use case
+
+You can use a load report to display the customers per phase as well as the load per phase based on a chosen starting point in a utility network. Load reports are used for electric load restoration and balancing.
+
+## How to use the sample
+
+Choose phases to be included in the report. Tap "Run" to initiate a downstream trace on the network and create a load report. Tap "Reset" to clear the phases and start over.
```
vquach2404, line 13: Might be a good idea to mention something about swiping to delete here.

vquach2404 (approved):
Looks good! Just a few more suggestions. Feel free to add Phil!

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+        guard !remainingPhases.isEmpty else { return }
+        let alertController = UIAlertController(
+            title: "Add a phase to get the load report.",
+            message: nil,
+            preferredStyle: .actionSheet
+        )
+        remainingPhases.forEach { phase in
+            alertController.addAction(UIAlertAction(title: phase.name, style: .default) { _ in
+                self.phaseSummaries.append((phase, nil))
+                self.tableView.insertRows(at: [IndexPath(row: self.phaseSummaries.endIndex - 1, section: 0)], with: .automatic)
```
philium, line 170: Have you thought about keeping the table view sorted by name? As far as I know, the Swift standard library doesn't provide a way to get the insertion index into a sorted array, but `Foundation` does:

```swift
let insertionIndex = (selectedPhases as NSArray)
    .index(of: phase, inSortedRange: NSRange(selectedPhases.indices), options: [.insertionIndex], usingComparator: { ($0 as! AGSCodedValue).name.compare(($1 as! AGSCodedValue).name) })
selectedPhases.insert(phase, at: insertionIndex)
self.tableView.insertRows(at: [IndexPath(row: insertionIndex, section: 0)], with: .automatic)
```

It may not look pretty, but it gets the job done. Just a thought.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+        alertController.addAction(cancelAction)
+        alertController.popoverPresentationController?.barButtonItem = sender
+        present(alertController, animated: true)
+    }
+    
+    @IBAction func runBarButtonItemTapped(_ sender: UIBarButtonItem) {
+        guard !phaseSummaries.isEmpty,
+              let expression = initialExpression,
+              let phasesNetworkAttribute = phasesNetworkAttribute,
+              let traceParameters = traceParameters else { return }
```
philium, line 183: These should be non-nil, right? And if they are `nil`, the run button shouldn't be enabled. So could those properties be declared as implicitly unwrapped optionals?

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+        
+        let traceGroup = DispatchGroup()
+        SVProgressHUD.show(withStatus: "Creating load report…")
+        
+        for i in 0..<phaseSummaries.count {
```
philium, line 188: FYI: `count` is not an index. To iterate over the indexes of a collection, always use `Collection.indices`.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+            let phase = phaseSummaries[i].phase
+            guard let phaseCode = phase.code else { continue }
+            // Create a conditional expression.
+            let phasesAttributeComparison = AGSUtilityNetworkAttributeComparison(networkAttribute: phasesNetworkAttribute, comparisonOperator: .doesNotIncludeAny, value: phaseCode)!
+            // Chain it with the base condition using an OR operator.
+            traceParameters.traceConfiguration?.traversability?.barriers = AGSUtilityTraceOrCondition(leftExpression: expression, rightExpression: phasesAttributeComparison)
+            
+            traceGroup.enter()
+            utilityNetwork.trace(with: traceParameters) { [weak self] results, _ in
```
philium, line 197: This is only place that `traceParameters` is actually used, right? Is there an advantage to creating it ahead of time and reusing it? Since it has to be configured each time anyway (with the trace conditions), it seems to me that it might make more sense to create the parameters here and not store them as a property. What do you think?

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+                    traceGroup.leave()
+                }
+                // Return if not result and ignore any trace error.
+                guard let results = results else { return }
+                var totalCustomers = 0
+                var totalLoad = 0
+                results.forEach { result in
+                    if let elementResult = result as? AGSUtilityElementTraceResult {
+                        // Get the unique customers count.
+                        totalCustomers = Set(elementResult.elements.map { $0.objectID }).count
```
philium, line 208:
```suggestion
                        totalCustomers = Set(elementResult.elements.map(\.objectID)).count
```

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+                    } else if let functionResult = result as? AGSUtilityFunctionTraceResult {
+                        // Get the total load with a function output.
+                        totalLoad = functionResult.functionOutputs.first?.result as? Int ?? 0
+                    }
```
philium, line 212: This is totally up to you, but when I have to test the same value multiple times, I prefer to use a `switch` statement:

```swift
switch result {
case let elementResult as AGSUtilityElementTraceResult:
    // Get the unique customers count.
    totalCustomers = Set(elementResult.elements.map(\.objectID)).count
case let functionResult as AGSUtilityFunctionTraceResult:
    // Get the total load with a function output.
    totalLoad = functionResult.functionOutputs.first?.result as? Int ?? 0
default:
    break
}
```

It does require adding a `default` case, but I find it to be more readable. With the `if` statement approach, all clauses have to be read in order to find out that the same value is being tested each time.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+                }
+                self?.phaseSummaries[i].summary = PhaseSummary(phase: phase, totalCustomers: totalCustomers, totalLoad: totalLoad)
```
philium, line 214: If `self` is no longer around, there's no use in performing any work at all. Therefore I would suggest unwrapping `self` at the beginning of the closure.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+    
+    /// A struct for the phase summary, which contains the phase,
+    /// the total customers and total load for the phase.
+    struct PhaseSummary {
+        let phase: AGSCodedValue
+        let totalCustomers: Int
+        let totalLoad: Int
+        
+        var description: String {
+            "Customers: \(totalCustomers)\tLoad: \(totalLoad)"
```
philium, line 49: For user-visible text, it is best not to use string interpolation (unless you're using SwiftUI, but that's a totally different thing). String interpolation doesn't produce a localized result, so it is not a practice to model for our customers who may localize their apps. Instead, I would suggest using a number formatter. Then the number of customers can be displayed with a grouping separator ("1,190" instead of "1190"), which is what I expected. What does `Load` represent? I don't know if it makes sense for that value to use a grouping separator.

Since `description` is only used by the table view, it might make sense to eliminate this property and just construct the string in `tableView(_:cellForRowAt:)`.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReport.storyboard
```diff
@@ -0,0 +1,103 @@
...
+                                        <connections>
+                                            <action selector="addBarButtonItemTapped:" destination="SSS-gO-RzR" id="YML-fw-rnH"/>
+                                        </connections>
+                                    </barButtonItem>
+                                    <barButtonItem style="plain" systemItem="flexibleSpace" id="xUT-hy-Cda"/>
+                                    <barButtonItem title="Run" id="5Du-wq-bZz">
+                                        <connections>
+                                            <action selector="runBarButtonItemTapped:" destination="SSS-gO-RzR" id="2kg-VK-tP2"/>
+                                        </connections>
+                                    </barButtonItem>
```
philium, line 39: Should all the buttons be disabled by default and then enabled as needed?

yo1995: Good point. I didn't change the states of buttons in the first place, because I thought it would introduce ~20 lines of additional code that has nothing to do with the APIs themselves as the sample is already pretty long. 

Since you mentioned it, I'll disable and enable the buttons in the code.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
yo1995: 🤩  wasn't aware of that before... That's pretty neat!

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
yo1995: IMO the order doesn't matter too much

- I tend to keep the list of phases in the order of how we added them (sort by date added 😆 )
- I'm not sure if it makes sense to sort current phases by names... 😅  But I sorted the list of choices/UIActions because it's easier for user to pick
- The original csharp design doesn't sort the list

image

philium: I don't care very much how the list is sorted, but it seems like it would be nice to have the order be sorted somehow, so that the order is always the same. Not to completely redesign this sample for iOS, but another options is to do away with the action sheet and have two sections in the table view, one with the selected phases and one with the available ones and let users order them however they want. See Settings > [Control Center](https://support.apple.com/en-us/HT211812) for a good example of that.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
yo1995: I have a question here: what is the best practice to construct strings with both mutable and immutable parts?

In our case for example, I want to display the string "Phase: \(phase.name)" in the cell's title text label. I know it is the best to use a static label to display the "Phase: " portion, but since I don't want to do too much UI work here, what about sth like `let title = "Phase: " + phase.name`? So that the first string can still be localized.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
yo1995: And it is very interesting to learn the ObjC way of saying binary search 🤣 Never heard of this method before. Thanks for sharing!

Ref: https://github.com/lionheart/openradar-mirror/issues/23

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
yo1995: Yeah the original design is pretty tangled here... So I cannot clearly tell which to hold. The design holds the trace parameters.

And it seems to me that if we don't hold the `traceParameters`, we'll instead need to hold the `traceConfiguration` and `startingLocation` to create the parameters. So I'll leave it unchanged and wait for your next review.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+            // Proceed if the utility network loaded without issue.
+            self.utilityNetworkDidLoad(startingLocation: startingLocation, traceConfiguration: utilityTierConfiguration)
+        }
+    }
+    
+    /// Called in response to the utility network load operation completing.
+    /// - Parameters:
+    ///   - startingLocation: The utility element to start the trace from.
+    ///   - traceConfiguration: The utility tier's trace configuration.
+    func utilityNetworkDidLoad(startingLocation: AGSUtilityElement, traceConfiguration: AGSUtilityTraceConfiguration) {
```
yo1995, line 104: After housekeeping the old monstrously lengthy method, I'm still not quite satisfied with this one. There are a few things going on after the utility network is loaded

- Get a list of phases or `[AGSCodedValue]`
- Create a base `AGSUtilityTraceParameters` with some default configurations
- Hold reference to a few properties

The difficulty is that getting the trace parameters depends on multiple preliminary steps and unwrapped variables.  Looking for more suggestions.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell: UITableViewCell
+        let phase: AGSCodedValue
+        switch Section.allCases[indexPath.section] {
+        case .included:
+            cell = tableView.dequeueReusableCell(withIdentifier: Section.included.cellIdentifier, for: indexPath)
+            phase = includedPhases[indexPath.row]
+            cell.textLabel?.text = "Phase: " + phase.name
+            if let summary = summaries[phase] {
+                cell.detailTextLabel?.text = "C: " + numberFormatter.string(from: NSNumber(value: summary.totalCustomers))! + "    L: " + numberFormatter.string(from: NSNumber(value: summary.totalLoad))!
```
yo1995, line 262: Not very happy with the strings here, but I don't know where to go 🤔 
Also I haven't updated the screenshot, in case further UI updates to make.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+    var phasesNetworkAttribute: AGSUtilityNetworkAttribute!
+    /// A list of possible phases populated from the network's attributes.
+    var excludedPhases = [AGSCodedValue]()
+    /// A list of phases that are included in the load report.
+    var includedPhases = [AGSCodedValue]() {
+        didSet {
+            let isEmpty = includedPhases.isEmpty
+            DispatchQueue.main.async { [weak self] in
+                self?.runBarButtonItem.isEnabled = !isEmpty
+            }
```
philium, line 44: Having to dispatch async in a property observer like this is a bit of a code smell. When the dispatched work is executed, there's no guarantee that `isEmpty == includedPhases.isEmpty`. Almost always, it would be better to move the dispatch to the place where the property is being updated. In this case, dispatching wouldn't be needed if the summaries were accumulated into a local variable and then the property was updated in the `work` closure of the call to `notify(qos:flags:queue:execute:)`, which is called on the main thread.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+        }
+    }
+    /// The phase summaries in the load report.
+    var summaries = [AGSCodedValue: PhaseSummary]() {
+        didSet {
+            if !resetBarButtonItem.isEnabled {
+                DispatchQueue.main.async { [weak self] in
+                    self?.resetBarButtonItem.isEnabled = true
+                }
```
philium, line 53: Having to dispatch async in a property observer like this is a bit of a code smell. Almost always, it would be better to move the dispatch to the place where the property is being updated. In this case, dispatching wouldn't be needed if the summaries were accumulated into a local variable and then the property was updated in the `work` closure of the call to `notify(qos:flags:queue:execute:)`, which is called on the main thread.

Similar with the `includedPhases` observer.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+           let loadAttribute = utilityNetwork.definition.networkAttributes.first(where: { $0.name == "Service Load" }),
+           // The phase attribute for getting total phase current load.
+           let phasesNetworkAttribute = utilityNetwork.definition.networkAttributes.first(where: { $0.name == "Phases Current" }) {
+            self.phasesNetworkAttribute = phasesNetworkAttribute
+            // Get possible coded phase values from the attributes.
+            if let domain = phasesNetworkAttribute.domain as? AGSCodedValueDomain {
+                excludedPhases = domain.codedValues.sorted { $0.name < $1.name }
+                DispatchQueue.main.async { [weak self] in
+                    guard let self = self else { return }
+                    self.tableView.reloadSections([Section.excluded.rawValue], with: .automatic)
```
philium, line 124: You could help yourself out here (and elsewhere) with an extension on `UITableView`:

```swift
private extension UITableView {
    func reloadSection(_ section: CreateLoadReportViewController.Section) {
        reloadSections([section.rawValue], with: .automatic)
    }
}

self.tableView.reloadSection(.excluded)
```

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
philium: To be honest, I don't know that one approach would be better than the others, so just stick with what you have.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+                        // Get the unique customers count.
+                        totalCustomers = Set(elementResult.elements.map(\.objectID)).count
+                    case let functionResult as AGSUtilityFunctionTraceResult:
+                        // Get the total load with a function output.
+                        totalLoad = functionResult.functionOutputs.first?.result as? Int ?? 0
+                    default:
+                        break
+                    }
+                }
+                self.summaries[phase] = PhaseSummary(totalCustomers: totalCustomers, totalLoad: totalLoad)
```
philium, line 212: As I've already mentioned, I recommend have a local `summaries` variable for accumulating the new summaries and then using it to update the `summaries` property when you have all of them.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+            // Binary search to find the insertion index.
+            let insertionIndex = (excludedPhases as NSArray).index(
+                of: phase,
+                inSortedRange: NSRange(excludedPhases.indices),
+                options: [.insertionIndex],
+                usingComparator: { ($0 as! AGSCodedValue).name.compare(($1 as! AGSCodedValue).name) }
+            )
+            excludedPhases.insert(phase, at: insertionIndex)
+            let section = Section.excluded.rawValue
+            tableView.insertRows(at: [IndexPath(row: insertionIndex, section: section)], with: .automatic)
```
philium, line 299: You could help yourself out here (and elsewhere) with an extension of `IndexPath`:

```swift
private extension IndexPath {
    init(row: Int, section: CreateLoadReportViewController.Section) {
        self.init(row: row, section: section.rawValue)
    }
}

tableView.insertRows(at: [IndexPath(row: insertionIndex, section: .excluded)], with: .automatic)
```

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+            let phase = excludedPhases[indexPath.row]
+            excludedPhases.remove(at: indexPath.row)
+            includedPhases.append(phase)
+            let section = Section.included.rawValue
+            let newRow = tableView.numberOfRows(inSection: section)
+            tableView.insertRows(at: [IndexPath(row: newRow, section: section)], with: .automatic)
+        default:
+            break
+        }
+        tableView.endUpdates()
```
philium, line 311: I recommend putting this in a `defer`, right after the call to `beginUpdates()`:

```swift
tableView.beginUpdates()
defer { tableView.endUpdates() }
```

That opens up options for returning in the `default` case and then moving some of the logic after the `switch` statement. For example:

```swift
let phase = includedPhases[indexPath.row]
let indexPathToInsert: IndexPath
switch editingStyle { ... }
tableView.deleteRows(at: [indexPath], with: .automatic)
tableView.insertRows(at: [indexPathToInsert], with: .automatic)
```

philium (commented):
When a phase is removed, should it's summary be removed as well? Just a thought. I noticed that when I removed a phase with a summary and then re-added it, the summary was still there.

yo1995:
From Friday's review

- The upper half of the list doesn't need to be reorder-able. Instead, making it alphabetically ordered make more sense.
- Good to use abbreviations
- Clear dict when removed

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -295,40 +284,42 @@ extension CreateLoadReportViewController: UITableViewDelegate, UITableViewDataSo
...
+            tableView.insertRows(at: [IndexPath(row: insertionIndex, section: .excluded)], with: .automatic)
         case .insert:
-            tableView.deleteRows(at: [indexPath], with: .automatic)
             let phase = excludedPhases[indexPath.row]
             excludedPhases.remove(at: indexPath.row)
-            includedPhases.append(phase)
-            let section = Section.included.rawValue
-            let newRow = tableView.numberOfRows(inSection: section)
-            tableView.insertRows(at: [IndexPath(row: newRow, section: section)], with: .automatic)
+            let insertionIndex = (includedPhases as NSArray).index(
```
yo1995, line 293: According to Friday's peer review, the included phases would be better to be sorted alphabetically.

This part is pretty sad - how I wish I can create a reference to a Swift array! 😿 
Basically it is the same code for 2 arrays which order is decided by the editing style. Looking for some suggestions.

yo1995 (commented):
Addressed a few suggestions and looking for more!

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
philium: This is still string concatenation, which is not a good idea in a localized environment. Again, not a problem for this project because it isn't localized, but I'd hate for someone to copy what we've done in a localized app. The way to go is to use a _format string_. If you do a search for `String(format:`, you'll find plenty of examples in our samples.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
philium: > Basically it is the same code for 2 arrays

So write a helper function! It could just be nested under `tableView(_:commit:forRowAt:)`:

```swift
func tableView(_ tableView: UITableView, commit editingStyle: UITableViewCell.EditingStyle, forRowAt indexPath: IndexPath) {
    func index(forInserting phase: AGSCodedValue, into array: [AGSCodedValue]) -> Int {
        (array as NSArray).index(
            of: phase,
            inSortedRange: NSRange(array.indices),
            options: [.insertionIndex],
            usingComparator: { ($0 as! AGSCodedValue).name.compare(($1 as! AGSCodedValue).name) }
        )
    }
    
    ...
    excludedPhases.insert(phase, at: index(forInserting: phase, into: excludedPhases))
    ...
    includedPhases.insert(phase, at: index(forInserting: phase, into: includedPhases))
    ...
}
```

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,357 @@
...
+    @IBOutlet var tableView: UITableView!
+    /// The button to reset the phase summaries.
+    @IBOutlet var resetBarButtonItem: UIBarButtonItem!
+    /// The button to run the trace and create a load report.
+    @IBOutlet var runBarButtonItem: UIBarButtonItem!
+    
+    // MARK: Properties
+    
+    /// A feature service of an electric utility network in Naperville, Illinois.
+    let utilityNetwork = AGSUtilityNetwork(url: URL(string: "https://sampleserver7.arcgisonline.com/arcgis/rest/services/UtilityNetwork/NapervilleElectric/FeatureServer")!)
```
yo1995, line 31: I'll probably need to use the new `AGSUtilityNetwork(url:map:)` initializer here... 🤔

yo1995:
From Jennifer: need to set `includeBarriers` to false.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
yo1995: No there is no map.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+            summaries.removeValue(forKey: phase)
+            tableView.insertRows(at: [IndexPath(row: insertionIndex, section: .excluded)], with: .automatic)
+        case .insert:
+            let phase = excludedPhases[indexPath.row]
+            excludedPhases.remove(at: indexPath.row)
+            insertionIndex = index(forInserting: phase, into: includedPhases)
+            includedPhases.insert(phase, at: insertionIndex)
+            tableView.insertRows(at: [IndexPath(row: insertionIndex, section: .included)], with: .automatic)
+        default:
+            return
```
philium, line 311: I don't remember when (if ever) this method is called with `none`, but that might be a possibility. If it does happen, the row at the given index path will be deleted--not great. It would be a good idea to guard either that `editingStyle != .none`, or, even better (since the enum is not frozen) `editingStyle == .delete && editingStyle == .insert`. Another option would be to use `performBatchUpdates(_:completion:)` inside each of the cases. And come to think of it, that is exactly what the documentation for `beginUpdates()` recommends:

image

I'm sorry for not remembering that sooner.

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+            .tier(withName: "Medium Voltage Radial")?
+            .traceConfiguration
+    }
+    
+    // MARK: Actions
+    
+    @IBAction func resetBarButtonItemTapped(_ sender: UIBarButtonItem) {
+        summaries.removeAll()
+        tableView.reloadSection(.included)
+        runBarButtonItem.isEnabled = !includedPhases.isEmpty
```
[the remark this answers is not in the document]
yo1995: Nice catch. It seems there is no need to set the button state here in reset, as the state should solely dependent on the emptiness of the included phases (which can only be changed within the `CreateLoadReportViewController.tableView(_:commit:forRowAt:)` method)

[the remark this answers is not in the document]
philium: Then it sounds like it should just be there (in `tableView(_:commit:forRowAt:)`).

arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
```diff
@@ -0,0 +1,357 @@
...
+                inSortedRange: NSRange(array.indices),
+                options: [.insertionIndex],
+                usingComparator: { ($0 as! AGSCodedValue).name.compare(($1 as! AGSCodedValue).name) }
+            )
+        }
+        
+        switch editingStyle {
+        case .delete:
+            includedPhases.remove(at: indexPath.row)
+            let phase = includedPhases[indexPath.row]
```
philium, line 288: There a now some bugs in this sample. Rows do not have the right description when moving between sections and removing the last row in a section results in a crash. I think the issue is that the phase is being removed before the phase is retrieved from the array, both here and in the `insert` case.

yo1995: Oh my...  🤦 I was trying to be clever without even thinking... Fixed!

philium (approved):
🎉

saratchandrakarumuri (commented):
* Reset button is just resetting the values in report but not changing the list of selected phases. Is this expected?
* If the above mentioned behavior is expected, what is the purpose of even resetting those values, are they tend to change on different runs?

yo1995:
> * Reset button is just resetting the values in report but not changing the list of selected phases. Is this expected?

Yep it is expected. Reset button only resets the load report summaries but not the phase table.

> * If the above mentioned behavior is expected, what is the purpose of even resetting those values, are they tend to change on different runs?

In a real world use case, the value will change on each run, since the electric utility network updates in real-time. But our demo doesn't change because the demo network is static.

I didn't find a doc that describes such use case in specific, but here is a similar one [with the water network](https://www.esri.com/en-us/landing-page/industry/water/2019/understanding-utility-network#main-content:~:text=happening%20in%20real%20time).

saratchandrakarumuri (approved):
Sample works as expected

PR #1044, opened 2021-03-12
Fix: Empty balloon for placemarks in Identify KML features

yo1995:
This PR proposes a fix for `common-samples/issues/2542`. 

## Background

In recent updates, core introduced a condition to return an empty string for the balloons/KML nodes that doesn't have description or extended data. The data source in the `Identify KML features` doesn't provide a description for each significant weather condition, thus empty balloons are displayed in the callout.

## Fix

If description is absent for a placemark, add a placeholder so that the content of the balloon can be displayed.

The fix is for U11, but adding the attribute doesn't conflict with U10, so it can go into `v.next`.

## Test

Please note: the sample can only be tested if the data source contains significant weather in the US. Also, the issue only appears in the recent U11 daily builds.

- Before: the popup for any feature in the map is empty
- After: the popup should show HTML content

### DIFF
--- a/arcgis-ios-sdk-samples/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
+++ b/arcgis-ios-sdk-samples/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
@@ -66,9 +66,15 @@ extension IdentifyKMLFeaturesViewController: AGSGeoViewTouchDelegate {
         geoView.callout.dismiss()
         geoView.identifyLayer(forecastLayer, screenPoint: screenPoint, tolerance: 15, returnPopupsOnly: false) { [weak self] (result) in
             if let error = result.error {
-                print("Error identifying layer: \(error)")
+                self?.presentAlert(error: error)
             } else if let placemarkIndex = result.geoElements.firstIndex(where: { $0 is AGSKMLPlacemark }) {
                 let placemark = result.geoElements[placemarkIndex] as! AGSKMLPlacemark
+                // Google Earth only displays the placemarks with description
+                // or extended data. To match its behavior, add a description
+                // placeholder if it doesn't exist in the data source.
+                if placemark.attributes["description"] == nil {
+                    placemark.attributes["description"] = "Weather condition"
+                }
                 self?.showCallout(for: placemark, at: mapPoint)
             }
         }
--- a/arcgis-ios-sdk-samples/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
+++ b/arcgis-ios-sdk-samples/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
@@ -17,33 +17,34 @@
 import UIKit
 import ArcGIS
 
-/// A view controller that manages the interface of the Identify KML Features
+/// A view controller that manages the interface of the Identify KML features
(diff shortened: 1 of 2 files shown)


vquach2404 (commented):
I'm a bit confused. Is it supposed to say "weather condition" inside the callout? 

Simulator Screen Shot - iPhone 12 Pro Max - 2021-03-12 at 15 46 25

yo1995:
> I'm a bit confused. Is it supposed to say "weather condition" inside the callout?

Nope. The popup should show content with the changes, whereas it is blank without these changes. Need to be built with the U11 SDK to test.

saratchandrakarumuri:
Tried with latest U11 build and still see the callout empty on identifying

yo1995:
I see... Let me try to figure it out! 😥

yo1995:
@vquach2404 and @saratchandrakarumuri Sorry for the confusion... I don't know why it worked on my end even if my original approach was wrong - maybe due to build artifacts or cache. 🤦‍♂️ 

Please take a look at the new changes! Thanks.

saratchandrakarumuri (approved):
Now it works as expected with the latest build

vquach2404 (approved):
Works as expected now! Since you're editing this sample, you might want to consider updating the coding style – like using `didSet` to instantiate the mapView, etc.

arcgis-ios-sdk-samples/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
```diff
@@ -66,9 +66,14 @@ extension IdentifyKMLFeaturesViewController: AGSGeoViewTouchDelegate {
         geoView.callout.dismiss()
         geoView.identifyLayer(forecastLayer, screenPoint: screenPoint, tolerance: 15, returnPopupsOnly: false) { [weak self] (result) in
             if let error = result.error {
-                print("Error identifying layer: \(error)")
-            } else if let placemarkIndex = result.geoElements.firstIndex(where: { $0 is AGSKMLPlacemark }) {
-                let placemark = result.geoElements[placemarkIndex] as! AGSKMLPlacemark
+                self?.presentAlert(error: error)
+            } else if let placemark = result.geoElements.first(where: { $0 is AGSKMLPlacemark }) as? AGSKMLPlacemark {
```
philium, line 70: If `first(where:)` returns a non-nil value, the second cast is always going to succeed, which is why it was originally written the way it was.

philium (approved):

arcgis-ios-sdk-samples/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
yo1995: In this case, I'll change to `if let placemark = result.geoElements.first as? AGSKMLPlacemark { ... }`, since the `AGSGeoView.identifyLayer(_:screenPoint:tolerance:returnPopupsOnly:completion:)` method only returns the top-most identified feature in the layer. Directly checking the first should be sufficient.

topmost

vquach2404 (approved):
Looks good, works as expected!

philium (approved):

saratchandrakarumuri (approved):

PR #1045, opened 2021-03-17
Fix: Add error checking for related feature sample

yo1995:
This PR adds an error checking for the  `Features` -> `Add delete related features` sample. If an error occurs during edit, it will present the error.

It is part of the efforts in `common-samples/issues/2602` where we want to fix the delete feature request for the feature service. While Ting is working with AGOL's Khaled to figure out the data issue, we need to first inform the user that there is an error, rather than fail silently.

|Before|After|
|-|-|
|demo-before|deletion-demo|


arcgis-ios-sdk-samples/Features/Add delete related features/RelatedFeaturesViewController.swift
```diff
@@ -130,15 +130,19 @@ class RelatedFeaturesViewController: UITableViewController {
         // show progress hud
         SVProgressHUD.show(withStatus: "Applying edits")
         
-        relatedTable.applyEdits { [weak self] (_, error) in
+        relatedTable.applyEdits { [weak self] (results, error) in
             SVProgressHUD.dismiss()
             
-            if let error = error {
-                // show error
+            if let firstResult = results?.first {
```
yo1995, line 136: I'm not pretty sure here as the original sample isn't quite well structured.

The sample's UI only allows applying 1 edit each time, so it is safe to assume there is only 1 result. But in real case there could be many local edits and all of them get applied at once. Looking forward to suggestions.

arcgis-ios-sdk-samples/Features/Add delete related features/RelatedFeaturesViewController.swift
```diff
@@ -130,15 +130,19 @@ class RelatedFeaturesViewController: UITableViewController {
...
+                if firstResult.completedWithErrors {
```
yo1995, line 137: Again I'm checking the flag rather than doing conditional unwrapping here because it's a better pattern if there are multiple results.

arcgis-ios-sdk-samples/Features/Add delete related features/RelatedFeaturesViewController.swift
vquach2404: Hm, if you want to make changes to the sample to support multiple edits at once, you could try only calling the query method once "Done" is tapped.

vquach2404 (approved):
Looks good so far. What do others think of updating changes only when "Done" is tapped?

philium (commented):
There is a slight behavior change here where `queryRelatedFeatures()` will no longer be called now if `results` is non-nil but empty. Is that okay?

arcgis-ios-sdk-samples/Features/Add delete related features/RelatedFeaturesViewController.swift
```diff
@@ -130,15 +130,18 @@ class RelatedFeaturesViewController: UITableViewController {
...
-            } else {
-                // query to update features
+            if let results = results {
+                if let firstResult = results.first,
+                   firstResult.completedWithErrors {
+                    // The edit fails with error.
+                    self?.presentAlert(error: firstResult.error!)
+                }
+                // Query to update features in the table.
                 self?.queryRelatedFeatures()
```
philium, line 142: `self` will now be potentially unwrapped multiple times, so it would be best to unwrap it at the beginning of the closure.

yo1995: 🤦 🤦  Lemme fix that...

philium (dismissed):

saratchandrakarumuri:
I now see the error message shown but for the first deletion feature is removed from the list and is not populated even when we re-identify the same feature again. But when we reopen the sample, deleted related feature is populated again

deleteRelated

yo1995:
> I now see the error message shown but for the first deletion feature is removed from the list and is not populated even when we re-identify the same feature again. But when we reopen the sample, deleted related feature is populated again

I also saw this issue, and it seems to be related with the service itself.

I checked the count of features in the query result, and it did give 1 less feature after the first deletion, even if the feature is not successfully deleted. So the behavior in UI is actually expected. That's been said, there are 3 potential ways to deal with it

1. If an error occurs, do not run `queryRelatedFeatures` to update the table.

    This makes the UI consistent when the deletion fails. However, it changes the behavior of the original sample, where it always updates the table no matter what happens in the `applyEdits` callback.

2. Leave it as-is and wait for the feature service to be fixed
3. Close this PR and don't make change to the sample.

    Since the sample used to be working well, maybe we just need to fix the service, or replace the data source.

@saratchandrakarumuri @philium Would you share your opinions on this issue? Thanks.

Meanwhile I'm working with AGOL's Khaled to try to fix the service. But I cannot accurately estimate how long it would take, as we are targeting an AppStore release next Wednesday.

arcgis-ios-sdk-samples/Features/Add delete related features/RelatedFeaturesViewController.swift
```diff
@@ -56,7 +56,7 @@ class RelatedFeaturesViewController: UITableViewController {
         parameters.orderByFields = [AGSOrderBy(fieldName: "OBJECTID", sortOrder: .descending)]
         
         // query for species related to the selected park
-        self.originFeatureTable.queryRelatedFeatures(for: self.originFeature, parameters: parameters) { [weak self] (results: [AGSRelatedFeatureQueryResult]?, error: Error?) in
+        self.originFeatureTable.queryRelatedFeatures(for: self.originFeature, parameters: parameters, queryFeatureFields: .loadAll) { [weak self] (results: [AGSRelatedFeatureQueryResult]?, error: Error?) in
```
yo1995, line 59: When I was testing with the dev feature table, it doesn't automatically populate all the fields from the table in a query. Even if the original code works fine with our current data source, I decide to add this parameter to be more defensive.

yo1995:
Hi all, feel free to ignore the comment above.

Working with AGOL folks, the problem was fixed from the backend. Please follow the following steps to test

1. Build the sample viewer on this branch **without** API key (API key doesn't allow us to visit the internal testing server)
2. Replace the feature table URL with the one I provided in Slack
3. Go to the sample. When credential challenge pops up, type in any account that has basemap access
4. Try to add and delete a feature

saratchandrakarumuri (approved):
Tried with the provided service and this sample is working as expected... I am ok with merging this now and updating the service later when fix goes to AGOL

philium (approved):

PR #1046, opened 2021-03-18
New: Display content of utility network container

vquach2404:
This PR implements the Display content of utility network container sample. See `common-samples/issues/2235` for more info.


arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,296 @@
...
+    func getLegends() {
+        // Create feature tables from URLs.
+        let electricDistributionTable = AGSServiceFeatureTable(url: URL(string: "https://sampleserver7.arcgisonline.com/arcgis/rest/services/UtilityNetwork/NapervilleElectric/FeatureServer/105")!)
+        let structureJunctionTable = AGSServiceFeatureTable(url: URL(string: "https://sampleserver7.arcgisonline.com/arcgis/rest/services/UtilityNetwork/NapervilleElectric/FeatureServer/900")!)
+        // Create feature layers using the feature tables.
+        featureLayers = [electricDistributionTable, structureJunctionTable].map(AGSFeatureLayer.init)
+        featureLayers.forEach { layer in
+            // Get the legend information of each layer.
+            layer.fetchLegendInfos { [weak self] (legendInfos, error) in
+                guard let self = self, let legendInfos = legendInfos else { return }
```
yo1995, line 89: We can first disable the legend button in storyboard, and enable the legend button here. 

Per my test, if there is no network connection, the legend table will only contain the 3 images for the attachment and box symbols.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,296 @@
...
+                    var symbol = AGSSymbol()
+                    if association.associationType == .attachment {
+                        symbol = self.attachmentSymbol
+                    } else {
+                        symbol = self.connectivitySymbol
+                    }
+                    overlay.graphics.add(AGSGraphic(geometry: association.geometry, symbol: symbol))
+                }
+                // Enable the bar button item to exit the container view.
+                self.exitBarButtonItem.isEnabled = true
```
yo1995, line 211: This should be moved up front to the beginning of the tap event. Otherwise there would be some timing bugs if you try to pan the map while get association call is still running.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainer.storyboard
```diff
@@ -0,0 +1,100 @@
...
+        <scene sceneID="bvO-0D-Blr">
+            <objects>
+                <viewController id="AJ2-vJ-kOC" customClass="DisplayContentUtilityNetworkContainerViewController" customModule="ArcGIS_Runtime_SDK_Samples" customModuleProvider="target" sceneMemberID="viewController">
+                    <view key="view" contentMode="scaleToFill" id="UYX-lE-URm">
+                        <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <subviews>
+                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="fgJ-4V-EwJ" userLabel="Map View" customClass="AGSMapView">
+                                <rect key="frame" x="0.0" y="44" width="414" height="769"/>
+                                <color key="backgroundColor" systemColor="systemBackgroundColor"/>
```
yo1995, line 22: map view bgc can be default, but it doesn't really matter.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainer.storyboard
```diff
@@ -0,0 +1,100 @@
...
+                                        <connections>
+                                            <action selector="exitContainerView" destination="AJ2-vJ-kOC" id="soo-wX-CO5"/>
+                                        </connections>
+                                    </barButtonItem>
+                                    <barButtonItem style="plain" systemItem="flexibleSpace" id="y1a-Dr-ncT"/>
+                                </items>
+                            </toolbar>
+                        </subviews>
+                        <viewLayoutGuide key="safeArea" id="OZH-zL-pnv"/>
+                        <color key="backgroundColor" systemColor="systemBackgroundColor"/>
```
yo1995, line 46: As we did in U10, the root view's bgc is tertiary sys bgc... 🕶️

yo1995 (commented):
To avoid an overwhelmingly huge review, I'll stop here and review the remaining parts later.

One more thing - per Phil here: cocoa/pull/10460#discussion_r585923

> - Rarely is `get` used as a prefix of method names in Swift.

It's up to you to choose another verb or just leave them as is. I believe in the samples we already have a bunch of method with get prefix. 😥

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
vquach2404: There is no tap event within this function. Do you mean move it out of this function and into geoView or beginning of `setViewpointGeometry`?

yo1995: Sorry for being vague. I commented on the wrong line... 😅  I mean to move the line below, `mapView.isUserInteractionEnabled = false` to the beginning of geoView -> didTapAtScreenPoint. So that once the user tap on a feature, they cannot pan or zoom the map, to get rid of timing bugs.

yo1995 (commented):
Another general issue I'd like to mention - Currently the usage of `guard` is not very consistent. They are...

- On the first line of an async callback
- On the first line of a method
- In the middle of a code block
- In a loop
- Mixed with `if let` conditional unwrapping

From your code I guess you are trying to use `guard` *just in time*. I think it is OK, but worth a bit of housekeeping to make them consistent.

I know we don't have a coding style guide for using `guard`, and my approach might be wrong. But I typically use it *aggressively* - always put it at where the ambiguity first appears. Whenever there is an optional that has to be non-nil for the following steps, I add a `guard`, so that we don't need to check its emptiness in the following steps.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,292 @@
...
+        let map = AGSMap(url: webMapURL)!
+        return map
+    }
+    
+    /// Create and load the utility network using the feature service URL.
+    func loadUtilityNetwork() {
+        utilityNetwork = AGSUtilityNetwork(url: featureServiceURL, map: mapView.map!)
+        utilityNetwork?.load { [weak self] error in
+            // Add self as the touch delegate for the map view.
+            self?.mapView.touchDelegate = self
```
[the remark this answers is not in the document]
vquach2404: I didn't put it in the else condition since someone pointed out that we should avoid pushing to the right due to nesting conditions. Not sure which is preferred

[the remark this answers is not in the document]
yo1995: Well I mean if there is an error when loading the UN, we probably shouldn't even allow the map view to detect tap events, to avoid unexpected outcomes.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,292 @@
...
+        // Disable the bar button item since container view will be exited.
+        exitBarButtonItem.isEnabled = false
+        // Remove all the objects that were added onto the graphics overlay.
+        graphicsOverlay.graphics.removeAllObjects()
+        (mapView.map?.operationalLayers as? [AGSLayer])?.forEach { layer in
+            // Make each operational layer visible.
+            layer.isVisible = true
+            guard let previousViewpoint = previousViewpoint else { return }
+            // Return to the viewpoint before container view was entered.
+            mapView.setViewpointGeometry(previousViewpoint.targetGeometry) { [weak self] _ in
```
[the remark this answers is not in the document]
vquach2404: 1. As far as I know, yes, we can assume `previousViewpoint` is always non-nil when the exit container view button is enabled.
2. `setViewpoint` does not provide animation so I opted for `setViewpointGeometry`.
3. When the container view only has one element, `setViewpointCenter` is needed because `setViewpointGeometry` can't take care of point geometry without providing a scale.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,292 @@
...
+                self.presentAlert(error: error)
+            }
+        }
+    }
+    
+    /// Create a swatch from the provided symbol.
+    ///
+    /// - Parameter symbol: The symbol to make a swatch from.
+    /// - Returns The resulting image that represents the swatch.
+    func makeSwatch(symbol: AGSSymbol) -> UIImage {
```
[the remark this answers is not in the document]
vquach2404: I hope you don't mind, I modeled the solution after Create symbol styles!

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,292 @@
...
+        utilityNetwork?.features(for: contentElements) { [weak self] (contentFeatures, error) in
+            guard let self = self else { return }
+            if let contentFeatures = contentFeatures {
+                if contentFeatures.count == 1 {
+                    self.presentAlert(title: nil, message: "This feature contains no associations.")
+                }
+                // Create and add a symbol for each of the features.
+                contentFeatures.forEach { content in
+                    let symbol = (content.featureTable as? AGSArcGISFeatureTable)?.layerInfo?.drawingInfo?.renderer?.symbol(for: content)
+                    self.graphicsOverlay.graphics.add(AGSGraphic(geometry: content.geometry, symbol: symbol))
```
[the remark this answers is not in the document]
vquach2404: I think the for loop is more appropriate here since each graphic is dependent on each content's geometry.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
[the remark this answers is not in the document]
philium: > `setViewpoint` does not provide animation so I opted for `setViewpointGeometry`.

Did you try using one of the `setViewpoint` methods that takes a completion handler? They should animate. If they don't, that's a bug.

vquach2404:
> Another general issue I'd like to mention - Currently the usage of `guard` is not very consistent. They are...
> From your code I guess you are trying to use `guard` _just in time_. I think it is OK, but worth a bit of housekeeping to make them consistent.

Hm, I tried to use it at the top level closures – the ones that would guard values for the rest of the method. Can you point out which instances you think should be changed?

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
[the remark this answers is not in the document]
vquach2404: Oops, sorry please ignore my comment above. I was referring to the wrong section of code.

yo1995 (commented):
I'll pause here to avoid making the review too lengthy which results in this 👇 
image

Also, could you check your local directory name if it is capitalized to `Utility Network` rather than `Utility network`? I'm getting some git error and cannot preview the changes in Xcode, which made me confused... 🤔

yo1995:
Oh 🤦  So the hidden discussion threshold is 10. Next time I'll notice that!!

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,366 @@
...
+    func adaptivePresentationStyle(for controller: UIPresentationController, traitCollection: UITraitCollection) -> UIModalPresentationStyle {
+        return .none
+    }
+}
+
+// MARK: - AGSAuthenticationManagerDelegate
+extension DisplayContentUtilityNetworkContainerViewController: AGSAuthenticationManagerDelegate {
+    func authenticationManager(_ authenticationManager: AGSAuthenticationManager, didReceive challenge: AGSAuthenticationChallenge) {
+        // NOTE: Never hardcode login information in a production application. This is done solely for the sake of the sample.
+        let credentials = AGSCredential(user: "editor01", password: "S7#i2LWmYH75")
```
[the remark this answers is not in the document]
vquach2404: Good idea, I've changed the credentials to use viewer01. In this case, the map is loaded from the web map which requires authentication because it's hosted on the sample server 7 portal. The authentication from loading the web map doesn't transfer when loading the UN. Generally, we'll only use the authentication manager in situations like this where we'll be prompted multiple separate times for credentials.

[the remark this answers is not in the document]
yo1995: Nice! Thanks.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,366 @@
...
+                }
+            }
+        }
+        legendGroup.notify(queue: .main) { [weak self] in
+            guard let self = self else { return }
+            let featureLayerSymbols = self.legendInfos.reduce(into: [(String, AGSSymbol)]() ) { specifiedSymbols, legendInfo in
+                if !legendInfo.name.isEmpty, let symbol = legendInfo.symbol {
+                    specifiedSymbols.append((legendInfo.name, symbol))
+                }
+            }
```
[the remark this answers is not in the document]
vquach2404: >  but it is also OK to fetch legend info and create symbol swatches all at once (so that we can use [LegendItem] instead of the array of tuples).

What do you mean? I only created the array of tuples so that I could make swatches for the feature layer legend items and the added line symbols all at once. Do you think there's a more efficient solution?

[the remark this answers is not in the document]
yo1995: I'm suggesting sth (drafted in the editor, might not compile) like 👇 that doesn't need the temp tuple.
```swift
            featureLayers.forEach { layer in
                legendGroup.enter()
                // Get the legend information of each layer.
                layer.fetchLegendInfos { [weak self] legendInfos, _ in
                    defer { legendGroup.leave() }
                    guard let legendInfos = legendInfos else { return }
                    legendInfos.forEach { info in
                        info.symbol.createSwatch { image, _ in
                            defer { getSymbolsGroup.leave() }
                            guard let image = image else { return }
                            legendItems.append(LegendItem(name: info.name, image: image))
                        }
                    }
                }
            }
```

But I like the idea to split into 2 methods when it's too lengthy, so maybe just keep what you have here.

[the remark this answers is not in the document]
vquach2404: Oh I see. Yeah I agree – I like having 2 methods so that I can utilize it for the line symbols as well.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,366 @@
...
+        }
+    }
+    
+    // MARK: - AGSGeoViewTouchDelegate
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        // Turn off user interaction to avoid straying away from the container view.
+        self.mapView.isUserInteractionEnabled = false
+        // Identify the top most feature that corresponds to the tapped point.
+        mapView.identifyLayers(atScreenPoint: screenPoint, tolerance: 5, returnPopupsOnly: false) { [weak self] (layerResults, error) in
```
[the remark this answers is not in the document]
vquach2404: I'm not sure I understand. The doc says a completion with `[AGSIdentifyLayerResult]?` is returned.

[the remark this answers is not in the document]
yo1995: The completion block will give either an error, or 1 feature in the array - the **top-most feature**. Therefore, if there is no error, you may conditionally unwrap the first element in the array, and I think it is guaranteed it exists.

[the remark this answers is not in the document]
vquach2404: I see, no problem! I think it has to stay as an array though because I have experienced instances where there are 2 values in the `layerResults` array. It occurs when 2 features are super close to one another like these 
Screen Shot 2021-04-15 at 1 03 07 PM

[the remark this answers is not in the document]
vquach2404: Ralf G. confirmed that this is expected behavior. This is what he said on Slack:

> if this is the function that you are using then you would get 2 results because it should return the top most feature from each layer

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,358 @@
...
+        // Hide the layers to display the container view.
+        (mapView.map?.operationalLayers as? [AGSLayer])?.forEach { layer in
+            layer.isVisible = false
+        }
+        // Get the corresponding features for the array of content elements.
+        utilityNetwork?.features(for: contentElements) { [weak self] (contentFeatures, error) in
+            guard let self = self else { return }
+            if let contentFeatures = contentFeatures {
+                // Create and add a symbol for each of the features.
+                contentFeatures.forEach { content in
```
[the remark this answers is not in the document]
vquach2404: Hm, I gave that workflow a try but it was redundant in my opinion. 2-3 lines (making the symbol, adding the graphics) would have to be repeated twice –  if the count is 1 and if it isn't. I've shortened the `let point =...` line which I think is a good compromise.

yo1995 (approved):

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+                    let symbol = (content.featureTable as? AGSArcGISFeatureTable)?.layerInfo?.drawingInfo?.renderer?.symbol(for: content)
+                    self.graphicsOverlay.graphics.add(AGSGraphic(geometry: content.geometry, symbol: symbol))
+                }
```
philium, line 217: Better to create the graphics and add them all at once than to add them one at a time.
```suggestion
                let graphics: [AGSGraphic] = contentFeatures.compactMap { (feature) in
                    guard let featureTable = feature.featureTable as? AGSArcGISFeatureTable,
                          let symbol = featureTable.layerInfo?.drawingInfo?.renderer?.symbol(for: feature) else {
                        return nil
                    }
                    return AGSGraphic(geometry: feature.geometry, symbol: symbol)
                }
                self.graphicsOverlay.graphics.addObjects(from: graphics)
```

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+                // Set the bounding box which defines the container view may be computed using the extent of the features it contains or centered around its geometry at the container's view scale.
+                if contentFeatures.count == 1,
+                   let point = contentFeatures.first?.geometry as? AGSPoint {
+                    self.mapView.setViewpointCenter(point, scale: containerElement.assetType.containerViewScale) { _ in
+                        self.presentAlert(title: nil, message: "This feature contains no associations.")
+                        guard let boundingBox = self.mapView.currentViewpoint(with: .boundingGeometry)?.targetGeometry else { return }
+                        self.identifyAssociationsWithExtent(boundingBox: boundingBox)
+                    }
+                } else {
+                    let boundingBox = AGSGeometryEngine.bufferGeometry(self.graphicsOverlay.extent, byDistance: 0.05)
```
philium, line 227: This value is buffered again in `identifyAssociationsWithExtent(boundingBox:)`. Does it need to be buffered twice? It seems to me that the single point extent (in the other branch of the `if` statement) doesn't need to be buffered, so the extent could just be buffered here.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+    ///
+    /// - Parameter boundingBox: The gemeotry which represents the boundaries of the extent.
+    func identifyAssociationsWithExtent(boundingBox: AGSGeometry) {
+        // Add the bounding box symbol.
+        graphicsOverlay.graphics.add(AGSGraphic(geometry: boundingBox, symbol: boundingBoxSymbol.symbol))
+        let geometry = AGSGeometryEngine.bufferGeometry(graphicsOverlay.extent, byDistance: 0.05)!
+        mapView.setViewpointGeometry(geometry) { [weak self] _ in
+            guard let self = self else { return }
+            // Get the associations for this extent to display how content features are attached or connected.
+            self.utilityNetwork?.associations(withExtent: self.graphicsOverlay.extent) { containmentAssociations, _ in
```
philium, line 246: This completion handler is strongly capturing `self`.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+                guard let containmentAssociations = containmentAssociations else { return }
```
philium, line 247: If there are no associations (if there was an error), should the exit button item still become enabled?

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+                containmentAssociations.forEach { association in
+                    var symbol = AGSSymbol()
+                    if association.associationType == .attachment {
+                        symbol = self.attachmentSymbol.symbol
+                    } else {
+                        symbol = self.connectivitySymbol.symbol
+                    }
+                    self.graphicsOverlay.graphics.add(AGSGraphic(geometry: association.geometry, symbol: symbol))
+                }
```
philium, line 256:
```suggestion
                let graphics: [AGSGraphic] = containmentAssociations.map { (association) in
                    let symbol = association.associationType == .attachment
                        ? self.attachmentSymbol.symbol
                        : self.connectivitySymbol.symbol
                    return AGSGraphic(geometry: geometry, symbol: symbol)
                }
                self.graphicsOverlay.graphics.addObjects(from: graphics)
```

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+    
+    // MARK: - UIViewController
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        AGSAuthenticationManager.shared().delegate = self
+        // Load the utility network.
+        loadUtilityNetwork()
+        // Add a graphics overlay.
+        mapView.graphicsOverlays.add(graphicsOverlay)
```
philium, line 289: It seems like this should be done in the `mapView` property observer.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+        // Get the legends from the feature service.
+        fetchLegendInfo()
+        // Add the source code button item to the right of navigation bar.
+        (self.navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["DisplayContentUtilityNetworkContainerViewController"]
+    }
+    
+    // MARK: - Navigation
```
philium, line 296:
```suggestion
    // MARK: - Navigation
    
```

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        if segue.identifier == "DisplayContentLegendSegue",
+            let controller = segue.destination as? UITableViewController {
+                controller.presentationController?.delegate = self
+                controller.tableView.dataSource = symbolsDataSource
+            }
+    }
+    
+    // MARK: - UIAdaptivePresentationControllerDelegate
```
philium, line 305:
```suggestion
    // MARK: - UIAdaptivePresentationControllerDelegate
    
```

However, usually this type of conformance is done in an extension.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+    func adaptivePresentationStyle(for controller: UIPresentationController, traitCollection: UITraitCollection) -> UIModalPresentationStyle {
+        return .none
+    }
+}
+
+// MARK: - AGSAuthenticationManagerDelegate
```
philium, line 311:
```suggestion
// MARK: - AGSAuthenticationManagerDelegate

```

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+extension DisplayContentUtilityNetworkContainerViewController: AGSAuthenticationManagerDelegate {
+    func authenticationManager(_ authenticationManager: AGSAuthenticationManager, didReceive challenge: AGSAuthenticationChallenge) {
+        // NOTE: Never hardcode login information in a production application. This is done solely for the sake of the sample.
+        let credentials = AGSCredential(user: "viewer01", password: "I68VGU^nMurF")
+        challenge.continue(with: credentials)
```
philium, line 316:
```suggestion
        let credential = AGSCredential(user: "viewer01", password: "I68VGU^nMurF")
        challenge.continue(with: credential)
```

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+    func numberOfSections(in tableView: UITableView) -> Int {
+        1
+    }
+
+    func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
+        "Legend"
+    }
+}
+
+// MARK: - LegendItem
```
philium, line 350:
```suggestion
// MARK: - LegendItem

```

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+private struct LegendItem {
+    let name: String
+    let image: UIImage
+}
+
+// MARK: - ContainerViewSymbol
```
philium, line 356:
```suggestion
// MARK: - ContainerViewSymbol

```

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+struct ContainerViewSymbol {
```
philium, line 357: This type should be `private`.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+    }
+    
+    /// Create a map.
+    ///
+    /// - Returns: An `AGSMap` object.
+    func makeMap() {
+        let webMapURL = URL(string: "https://sampleserver7.arcgisonline.com/portal/home/item.html?id=813eda749a9444e4a9d833a4db19e1c8")!
+        let map = AGSMap(url: webMapURL)
+        mapView.map = map
+    }
```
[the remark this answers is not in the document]
vquach2404: I wanted to avoid force unwrapping `map`. I like the consistency of having the `makeMap()` function but I agree, it can be shortened – especially with a sample this long. I'll condense it to remove the function unless you think I should resolve it differently.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+            // Get the legend information of each layer.
+            layer.fetchLegendInfos { [weak self] legendInfos, _ in
+                defer { legendGroup.leave() }
+                guard let legendInfos = legendInfos else { return }
+                self?.legendInfos.append(contentsOf: legendInfos)
+            }
+        }
+        legendGroup.notify(queue: .main) { [weak self] in
+            guard let self = self else { return }
+            let featureLayerSymbols: [(String, AGSSymbol)] = self.legendInfos.compactMap { legendInfo in
```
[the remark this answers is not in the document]
vquach2404: I gave it a shot but it seems like there are some impediments that make doing so not worth it. Each feature layer contains an `AGSLegendInfo` with the name "Unknown". Having duplicate keys causes an error in making the dictionary. I realize having duplicate `AGSLegendInfo`s also isn't ideal but there doesn't seem to be an easy way to remove just one duplicate without hardcoding the string "Unknown". I was okay with having 2 "Unknown"s since they technically come from different legends. What're your thoughts?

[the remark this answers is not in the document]
philium: How are you creating the dictionary?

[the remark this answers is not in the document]
vquach2404:
```swift
let featureLayerSymbols = [String: AGSSymbol](uniqueKeysWithValues: accumulatedLegendInfos.compactMap { legendInfo in
     if !legendInfo.name.isEmpty,
        let symbol = legendInfo.symbol {
           return (legendInfo.name, symbol)
         } else {
             return nil
          }
})
```
This is how I tried to create the dictionary. I got an error that said "Duplicate keys" due to having 2 `AGSLegendInfo`s with the name "Unknown" populated from each feature layer.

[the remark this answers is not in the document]
philium: Could you use [`Dictionary.init(_:uniquingKeysWith:)`](https://developer.apple.com/documentation/swift/dictionary/3127161-init) instead?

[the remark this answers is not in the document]
vquach2404:
``` swift
let symbols: [(String, AGSSymbol)] = accumulatedLegendInfos.compactMap { legendInfo in
                if !legendInfo.name.isEmpty,
                   let symbol = legendInfo.symbol {
                    return (legendInfo.name, symbol)
                } else {
                    return nil
                }
            }
            let featureLayerSymbols = Dictionary(symbols, uniquingKeysWith:  { (first, _) in first})
```
I could use `Dictionary.init(_:uniquingKeysWith:)` but it would still require making an array of tuples first.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
vquach2404: The first buffers within the bounding box (yellow dashed line) and the second buffers outside of it. Without one or the other, symbols might be on the bounding box line or the bounding box line will be on the edge of the screen. IMO, both are necessary but there may be a more efficient solution. Do you have any further suggestions?

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
vquach2404: I think there was a previous discussion about removing the error variable because `containmentAssociations` is guarded. So even if `containmentAssociations` has 0 values, I think it would still enter the container view and the user would need an option to exit.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
vquach2404: I tried this but it didn't add the symbols correctly. This was the output: 
Simulator Screen Shot - iPhone 12 Pro Max - 2021-05-03 at 12 31 13

vquach2404:
@philium Hi, this is ready for another review. Please view the unresolved comments as I provided more info in the replies!

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
philium: That's because there was a bug in my suggestion (`geometry` instead of `association.geometry`). I've updated the suggestion.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,360 @@
...
+        swatchGroup.notify(queue: .main) {
+            completion(legendItems)
+        }
+    }
+    
+    /// Get the container feature that was tapped on.
+    ///
+    /// - Parameters:
+    ///   - layerResults: The layer results identified by the touch delegate.
+    ///   - completion: A closure to pass back the feature that was tapped on.
```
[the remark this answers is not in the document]
philium: This method still does not have a parameter named `completion`.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
philium: I think I understand. Thanks!

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// MARK: - UIAdaptivePresentationControllerDelegate
+
+extension DisplayContentUtilityNetworkContainerViewController: UIAdaptivePresentationControllerDelegate {
+    func adaptivePresentationStyle(for controller: UIPresentationController, traitCollection: UITraitCollection) -> UIModalPresentationStyle {
+        return .none
+    }
+}
+// MARK: - AGSAuthenticationManagerDelegate
```
philium, line 313:
```suggestion

// MARK: - AGSAuthenticationManagerDelegate
```

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+![Image of display content of utility network container](display-content-utility-network.png)
+
+## Use case
+
+Offering a container view for features aids in the review for valid structural attachment and containment relationships and helps determine if a dataset has an association role set. Container views often model a cluster of electrical devices on a pole top or inside a cabinet or vault.
+
+## How to use the sample
+
+Tap on a container feature to show all features inside the container.  The container is shown as a polygon graphic with the content features contained within. The viewpoint and scale of the map are also changed to the container's extent. Connectivity and attachment associations inside the container are shown as dotted lines.
```
philium, line 13:
```suggestion
Tap on a container feature to show all features inside the container. The container is shown as a polygon graphic with the content features contained within. The viewpoint and scale of the map are also changed to the container's extent. Connectivity and attachment associations inside the container are shown as dotted lines.
```

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/README.md
```diff
@@ -0,0 +1,46 @@
...
+* AGSSubtypeFeatureLayer
+* AGSUtilityAssociation
+* AGSUtilityAssociationType
+* AGSUtilityElement
+* AGSUtilityNetwork
+* AGSUtilityNetworkDefinition  
+
+## About the data
+
+The [Naperville electric network feature service](https://sampleserver7.arcgisonline.com/arcgis/rest/services/UtilityNetwork/NapervilleElectric/FeatureServer), hosted on ArcGIS Online, contains a utility network used to find associations shown in this sample and a web map portal item, [Naperville electric containers](https://ss7portal.arcgisonline.com/arcgis/home/item.html?id=5b64cf7a89ca4f98b5ed3da545d334ef), that use the same feature service endpoint which displays only container features.
```
philium, line 42: Something is a bit off here. Should it be "that uses" instead of "that us"?

```suggestion
The [Naperville electric network feature service](https://sampleserver7.arcgisonline.com/arcgis/rest/services/UtilityNetwork/NapervilleElectric/FeatureServer), hosted on ArcGIS Online, contains a utility network used to find associations shown in this sample and a web map portal item, [Naperville electric containers](https://ss7portal.arcgisonline.com/arcgis/home/item.html?id=5b64cf7a89ca4f98b5ed3da545d334ef), that uses the same feature service endpoint, but displays only container features.
```

philium (dismissed):
Looks good!

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/README.md
```diff
@@ -0,0 +1,46 @@
...
+* AGSSubtypeFeatureLayer
+* AGSUtilityAssociation
+* AGSUtilityAssociationType
+* AGSUtilityElement
+* AGSUtilityNetwork
+* AGSUtilityNetworkDefinition  
+
+## About the data
+
+The [Naperville electric network feature service](https://sampleserver7.arcgisonline.com/arcgis/rest/services/UtilityNetwork/NapervilleElectric/FeatureServer), hosted on ArcGIS Online, contains a utility network used to find associations shown in this sample and a web map portal item, [Naperville electric containers](https://ss7portal.arcgisonline.com/arcgis/home/item.html?id=5b64cf7a89ca4f98b5ed3da545d334ef), that uses the same feature service endpoint, but displays only container features.
```
saratchandrakarumuri, line 42: Portal item url mentioned here for containers service is not same as the one used in the sample code and this is an invalid link

saratchandrakarumuri (changes requested):

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/README.md
vquach2404: Oops, please try again I just fixed it!

saratchandrakarumuri (approved):
new sample looks good

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,365 @@
...
+    /// The action that is prompted when exiting the container view.
+    @IBAction func exitContainerView() {
+        // Disable the bar button item since container view will be exited.
+        exitBarButtonItem.isEnabled = false
+        // Remove all the objects that were added onto the graphics overlay.
+        graphicsOverlay.graphics.removeAllObjects()
+        (mapView.map?.operationalLayers as? [AGSLayer])?.forEach { layer in
+            // Make each operational layer visible.
+            layer.isVisible = true
+        }
```
yo1995, line 68: I think this can be replaced by the `setOperationalLayersVisibility` method.

arcgis-ios-sdk-samples/Utility network/Display content of utility network container/DisplayContentUtilityNetworkContainerViewController.swift
```diff
@@ -0,0 +1,365 @@
...
+                    return AGSGraphic(geometry: association.geometry, symbol: symbol)
+                }
+                self.graphicsOverlay.graphics.addObjects(from: graphics)
+                // Enable the bar button item to exit the container view.
+                self.exitBarButtonItem.isEnabled = true
+            }
+        }
+    }
+    
+    // MARK: - AGSGeoViewTouchDelegate
```
yo1995, line 262:
```suggestion
    // MARK: - AGSGeoViewTouchDelegate
    
```

yo1995 (approved):
👍  🥳

yo1995 (approved):

saratchandrakarumuri (approved):

PR #1051, opened 2021-04-01
Fix: Various bugs in Edit and sync features

yo1995:
This PR is a quick fix to address the AppStore feedback on the crash in `Edit and sync features` sample

- Add `SanFrancisco.tpk` on-demand resource into plist dependency
- Fix the features not showing issue when the feature layers are loaded
- Clear away some smelling code

The sample would need further refactor, but as we are catching up with a release, let's punt it into the future...

Astonishingly, the On-Demand Resource bug has been there since 2019, but we never caught it as we've never run ad-hoc test on the AppStore released version. 😨 Really hard to catch.

### DIFF
--- a/arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
+++ b/arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
@@ -80,13 +80,14 @@ class EditAndSyncFeaturesViewController: UIViewController {
                 self.presentAlert(error: error)
             } else {
                 let featureServiceInfo = self.geodatabaseSyncTask.featureServiceInfo!
-                for index in featureServiceInfo.layerInfos.indices.reversed() {
-                    // For each layer in the serice, add a layer to the map.
-                    let layerURL = self.featureServiceURL.appendingPathComponent(String(index))
+                let featureLayers = featureServiceInfo.layerInfos.compactMap { (layerInfo) -> AGSFeatureLayer? in
+                    let layerID = layerInfo.id
+                    guard layerID >= 0 else { return nil }
+                    let layerURL = self.featureServiceURL.appendingPathComponent(String(layerID))
                     let featureTable = AGSServiceFeatureTable(url: layerURL)
-                    let featureLayer = AGSFeatureLayer(featureTable: featureTable)
-                    self.mapView.map?.operationalLayers.add(featureLayer)
+                    return AGSFeatureLayer(featureTable: featureTable)
                 }
+                self.mapView.map?.operationalLayers.addObjects(from: featureLayers)
                 self.barButtonItem.isEnabled = true
             }
         }
--- a/arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
+++ b/arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
@@ -26,10 +26,6 @@ class EditAndSyncFeaturesViewController: UIViewController {
             let map = AGSMap(basemap: AGSBasemap(baseLayer: tiledLayer))
             // Assign the map to the map view.
             mapView.map = map
-            
-            // Create a geodatabase sync task using the feature service URL.
-            geodatabaseSyncTask = AGSGeodatabaseSyncTask(url: featureServiceURL)
-            self.addFeatureLayers()
         }
     }
     
@@ -41,7 +37,6 @@ class EditAndSyncFeaturesViewController: UIViewController {
         }
     }
     
-    @IBOutlet private var toolBar: UIToolbar!
     @IBOutlet private var barButtonItem: UIBarButtonItem!
     @IBOutlet private var instructionsLabel: UILabel!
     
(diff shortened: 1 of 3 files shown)


arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -41,7 +37,6 @@ class EditAndSyncFeaturesViewController: UIViewController {
         }
     }
     
-    @IBOutlet private var toolBar: UIToolbar!
```
yo1995, line 44: This property isn't used in the code, so I removed it from both storyboard and source code.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -86,21 +80,14 @@ class EditAndSyncFeaturesViewController: UIViewController {
                 self.presentAlert(error: error)
             } else {
                 let featureServiceInfo = self.geodatabaseSyncTask.featureServiceInfo!
-                let map = self.mapView.map!
-                for index in featureServiceInfo.layerInfos.indices {
+                for index in featureServiceInfo.layerInfos.indices.reversed() {
                     // For each layer in the serice, add a layer to the map.
                     let layerURL = self.featureServiceURL.appendingPathComponent(String(index))
                     let featureTable = AGSServiceFeatureTable(url: layerURL)
-                    featureTable.load { [weak self, unowned featureTable] error in
```
yo1995, line 94: No need to explicitly load the feature layers, and resolved the wrong reference count issue.

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -86,21 +80,14 @@ class EditAndSyncFeaturesViewController: UIViewController {
                 self.presentAlert(error: error)
             } else {
                 let featureServiceInfo = self.geodatabaseSyncTask.featureServiceInfo!
-                let map = self.mapView.map!
-                for index in featureServiceInfo.layerInfos.indices {
+                for index in featureServiceInfo.layerInfos.indices.reversed() {
```
philium, line 83: Why was `reversed()` added?

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -86,21 +80,14 @@ class EditAndSyncFeaturesViewController: UIViewController {
...
                     // For each layer in the serice, add a layer to the map.
                     let layerURL = self.featureServiceURL.appendingPathComponent(String(index))
```
philium, line 85: Looking at this now, I'm wondering if the layer ID should be used instead of an arbitrary index?

```swift
let featureLayers = featureServiceInfo.layerInfos.compactMap { (layerInfo) -> AGSFeatureLayer? in
    let layerID = layerInfo.id
    guard layerID >= 0 else { return nil }
    let layerURL = self.featureServiceURL.appendingPathComponent(String(layerID))
    let featureTable = AGSServiceFeatureTable(url: layerURL)
    return AGSFeatureLayer(featureTable: featureTable)
}
self.mapView.map?.operationalLayers.addObjects(from: featureLayers)
```

yo1995: Applied!

arcgis-ios-sdk-samples/Edit data/Edit and sync features/EditAndSyncFeaturesViewController.swift
```diff
@@ -52,19 +47,18 @@ class EditAndSyncFeaturesViewController: UIViewController {
...
-
+    
     private var selectedFeature: AGSFeature? {
         didSet {
-            if let feature = selectedFeature {
-                if let featureLayer = feature.featureTable?.layer as? AGSFeatureLayer {
-                    featureLayer.select(feature)
-                }
+            if let feature = selectedFeature,
+               let featureLayer = feature.featureTable?.layer as? AGSFeatureLayer {
```
philium, line 54: This change means that if the selected feature's table's layer is not an AGSFeatureLayer, then the selection will not be cleared. Is that possible?

yo1995: My understanding might be wrong, but...

- Without making this change, if the `selectedFeature` is non-nil but its `featureTable.layer` is not an `AGSFeatureLayer`, nothing will happen
- With this change, if any of the 2 conditions is not met, then it would clear the selection.

It shouldn't make a difference in this sample as all the operational layers added are `AGSFeatureLayer`s. But I think it makes more sense to only allow 2 states:

1. Either a feature in an `AGSFeatureLayer` is selected
2. Or nothing selected

and avoid the state where `selectedFeature` is non-nil, but nothing happens in the UI.

philium: That makes sense. I think I just got confused.

philium (approved):

vquach2404 (approved):

saratchandrakarumuri (approved):
Samples works better with these new changes

PR #1052, opened 2021-04-01
Update: Create load report, Utility network URL for sample server 7

yo1995:
There has been 1 additional UN sample `Create load report` added since we created #1026 . With the rolled out SS7, the `Create load report` sample doesn't work with the invalid URL anymore.

This PR fixes the issue with the new URL and credentials.


philium (approved):

vquach2404 (approved):

saratchandrakarumuri (approved):
New service works good

PR #1053, opened 2021-04-05
Update: Bump version number and deployment target

yo1995:
This PR makes the following change to move towards update 11.

- Bump deployment target to iOS 13
- Bump version number to 100.11.0
- Update root README
- Merge back #1054 into `v.next`
- Address deprecation warnings: 
    - cocoa/issues/9876, cocoa/issues/9759, both for the `AGSApplicationDelegate`
    - The rerouting deprecated API
    - Other iOS 13 warnings
- SVProgressHUD deprecation warnings #1059 

This PR **DOES NOT** make the following changes

- ~~Update the toolkit dependency to latest `Update11` branch~~ Change will be made in a separate PR due to the complexity of testing out SPM


vquach2404 (approved):

philium:
As part of increasing the deployment target, please also remove any code that is now obsolete. For example, [these lines](https://github.com/Esri/arcgis-runtime-samples-ios/blob/f863edd6507f07eae2ccb2cdb9b4530936afd6f9/arcgis-ios-sdk-samples/AppDelegate.swift#L118-L120) will never be executed now.

philium:
> This PR DOES NOT [address]…SVProgressHUD deprecation warnings

Why not? PRs to the development branch should not introduce any new warnings, which this PR currently does.

yo1995:
> As part of increasing the deployment target, please also remove any code that is now obsolete. For example, [these lines](https://github.com/Esri/arcgis-runtime-samples-ios/blob/f863edd6507f07eae2ccb2cdb9b4530936afd6f9/arcgis-ios-sdk-samples/AppDelegate.swift#L118-L120) will never be executed now.

Sorry 🤕 I forgot to do that... It was in my checklist but I just somehow missed it.

philium:
It looks like at least some of the SVProgressHUD code could be easily change to not use deprecated APIs. If there are uses of deprecated APIs that can't be easily avoided, then temporarily [suppress them](https://stackoverflow.com/a/14065939) and create an issue (if there isn't one already) to make the necessary changes. I repeat, suppressing a deprecation warning should only be _temporary_.

yo1995:
> Why not? PRs to the development branch should not introduce any new warnings, which this PR currently does.

Because these deprecation warnings are coming from upgrading the deployment target to iOS 13, even though we didn't touch the source code of the HUD.

As it was proposed in #983 we want to replace it with a view controller based solution. So I don't know if we should still update `SVProgressHUD`... 🤔  given that it doesn't have a new release for almost 2 years.

philium:
A library becoming stale is part of the risk of adding a third-party dependency. Instead of updating the code, another option would be to wrap it in a local Swift package and add that as a dependency. As a dependency, any build warnings will not be shown.

yo1995:
OK. Let me put this PR back to draft and discuss with others to figure out how to tackle the HUD. I'll request your review when it is ready.

arcgis-ios-sdk-samples/Cloud and portal/Search for webmap by keyword/SearchForWebmapByKeywordViewController.swift
```diff
@@ -38,16 +38,6 @@ class SearchForWebmapByKeywordViewController: UICollectionViewController {
...
-            // for iOS 13 this must be added in viewDidLoad
-            // for iOS 12, it gets added in viewDidAppear, because for iOS 12 it doesn't
-            // show up if added here.
-            addSearchController()
-        }
-    }
-    
-    override func viewDidAppear(_ animated: Bool) {
-        super.viewDidAppear(animated)
         if navigationItem.searchController == nil {
```
philium, line 41: This check is no longer needed at `navigationItem.searchController` will always be `nil` when this method (`viewDidLoad()`) is called.

arcgis-ios-sdk-samples/Maps/Apply scheduled updates to preplanned map area/ApplyScheduledUpdatesToPreplannedMapAreaViewController.swift
```diff
@@ -98,15 +98,11 @@ class ApplyScheduledUpdatesToPreplannedMapAreaViewController: UIViewController {
             if updatesInfo.downloadAvailability == .available {
                 let downloadSize = updatesInfo.scheduledUpdatesDownloadSize
                 let downloadSizeString: String
```
philium, line 100: There's no need to declare this ahead of time now. Please move the declaration to the line where it is now being initialized.

yo1995:
With the newly merged changes, this branch should only contain warnings from the toolkit.
And these warnings will be addressed in a separate PR that adopts SPM in Sample Viewer.

vquach2404 (approved):

philium (approved):

PR #1056, opened 2021-04-08
Update: callout custom view text color

yo1995:
In U10 when the callout isn't dark mode adaptive, we hardcoded the text color to black in 2 custom callout views. Now it is time to revert them as callout supports Dark Mode!

- [x] Identify raster cell
- [x] project


philium (approved):

saratchandrakarumuri (approved):
Callouts in both the samples are now good in dark mode

PR #1057, opened 2021-04-09
Fix: Explicitly specify the result types for Create load report

yo1995:
After the Friday review meeting, Tanner and I collaborated to find out the issue within the code. And later we realized that it would be better to explicitly specify the result types for the trace, instead of relying on the default populated elements type.

image

Pros and cons for this change:

- Pros: the user can have a clearer expectation for the output result types
- Cons: it seems a bit redundant to override the default value, as it was described in the docstring

I find this change helpful because I didn't realize/took for granted that the element output type was implicitly included, without thinking about the reason.

Please let me know your thoughts!

---

This PR doesn't change the behavior of the sample.

### DIFF
--- a/arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
+++ b/arcgis-ios-sdk-samples/Utility network/Create load report/CreateLoadReportViewController.swift
@@ -97,9 +97,9 @@ class CreateLoadReportViewController: UIViewController {
         // Set the default expression.
         initialExpression = traceConfiguration.traversability?.barriers as? AGSUtilityTraceConditionalExpression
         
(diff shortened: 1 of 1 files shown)


tanneryould:
I find this change helpful because I was having a lot of trouble debugging this sample. I was only able to get one type of output/result (either a `UtilityElementTraceResult` or a `UtilityFunctionTraceResult`) but not both. I wasn't sure where the element result was coming from until I found that it was pre-defined by default.

I think we should make this change because it makes the code more explicit and may help the user understand what's going on, rather than expecting them to know that element result is predefined by default.

philium (approved):

vquach2404 (approved):
Sounds like an improvement!

PR #1058, opened 2021-04-14
Update: NMEA sample with GNSS device API

yo1995:
## Goals

To showcase the new NMEA LDS constructor in U11 that connect to a real device easily without pushing NMEA data manually.

## Changes and Updates

- talked to MFi device manufacturers to whitelist the samples app for next release, so that Apple can approve Samples app to connect to external accessories.
- update source code to support both mock data and U11 constructor, and to update UI and logics
- update README

## Demo

Please see the internally shared video - don't post here to avoid leaking my accurate location info 😅 

## How to Test

- If you don't have a GNSS surveyor, simply test the mock data functionalities, and look for any logic bugs in the code.
- If you have a GNSS surveyor that is one of these 3 brands: Bad Elf, EOS, SXBlue, then try to connect the device and switch back and forth to see if the sample works as expected

## Todolist

- [ ] after the changes are approved, have Divesh take a look
- [ ] verify the sample with 3 GNSS devices that are currently possessed by different teams
    - [x] Ting: BadElf
    - ~~Nick: EOS~~ Nick has sent back the GNSS accessory. 😢 `cocoa/issues/10553#issuecomment-2824557`
    - [ ] Apps?: SXBlue
- [ ] before merging PR, will add an iOS specific design doc to internal repo
- [ ] before merging PR, follow up with manufacturer representative to see if we get our bundle id whitelisted

## Extra Information

The BadElf CTO Brett offers to help to review the verbiage for adding GNSS receiver support to an app. So after the code is good and people are happy with the README, we can send him a link to this PR to get some insights!


arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -6,18 +6,22 @@ Parse NMEA sentences and use the results to show device location on the map.
 
 ## Use case
 
-NMEA sentences can be retrieved from GPS receivers and parsed into a series of coordinates with additional information. Devices without a built-in GPS receiver can retrieve NMEA sentences by using a separate GPS dongle, commonly connected bluetooth or through a serial port.
+NMEA sentences can be retrieved from GNSS/GPS receivers and parsed into a series of coordinates with additional information. Devices without a built-in GNSS receiver can retrieve NMEA sentences by using a separate GNSS dongle, commonly connected Bluetooth or through a serial port.
```
vquach2404, line 9:
```suggestion
NMEA sentences can be retrieved from GNSS/GPS receivers and parsed into a series of coordinates with additional information. Devices without a built-in GNSS receiver can retrieve NMEA sentences by using a separate GNSS dongle, commonly connected through Bluetooth or a serial port.
```

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -6,18 +6,22 @@ Parse NMEA sentences and use the results to show device location on the map.
...
 ## How it works
 
-1. Load NMEA sentences from a local file.
-2. Parse the NMEA sentence strings, and push data into `AGSNMEALocationDataSource`.
+1. Load NMEA sentences.
+    * If a supported GNSS receiver is connected, the sample can get NMEA updates from it.
+    * Otherwise, the sample will read mock data from a local file.
+2. Create an `AGSNMEALocationDataSource`. There are 2 ways to provide updates to the data source.
+    * When updates are received from a GNSS device or the mock data provider, push the data into `AGSNMEALocationDataSource`.
+    * In the Runtime SDK 100.11.0 release, it is also supported to create an `AGSNMEALocationDataSource` with the GNSS device connected via `EAAccessory` and its protocol specified in `UISupportedExternalAccessoryProtocols`. The data source created this way will automatically get updates from the device instead of requiring to push data explicitly.
```
vquach2404, line 24:
```suggestion
    * With the Runtime SDK 100.11.0 release, you can create an `AGSNMEALocationDataSource` with the GNSS device connected via `EAAccessory` and its protocol specified in `UISupportedExternalAccessoryProtocols`. The data source created this way will automatically get updates from the device instead of requiring to push data explicitly.
```
I know using "you" isn't ideal here but I feel like this phrasing reads better.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -30,10 +34,23 @@ Tap "Start" to parse the NMEA sentences into a simulated location data source, a
 
 ## About the data
 
-A list of NMEA sentences is used to initialize a `SimulatedNMEADataSource` object. This simulated data source provides NMEA data periodically, and allows the sample to be used on devices without a GPS dongle that produces NMEA data.
+A list of NMEA sentences is used to initialize a `SimulatedNMEADataSource` object. This simulated data source provides NMEA data periodically, and allows the sample to be used on devices without a GNSS receiver that produces NMEA data.
```
vquach2404, line 37:
```suggestion
A list of NMEA sentences is used to initialize a `SimulatedNMEADataSource` object. This simulated data source provides NMEA data periodically and allows the sample to be used on devices that produce NMEA data without a GNSS receiver.
```
I'm not as familiar with NMEA so my suggestion may be incorrect.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -30,10 +34,23 @@ Tap "Start" to parse the NMEA sentences into a simulated location data source, a
...
 
 The route taken in this sample features a [2-minute driving trip around Redlands, CA](https://arcgis.com/home/item.html?id=d5bad9f4fee9483791e405880fb466da).
 
+## Additional information
+
+To support GNSS device connection in an app, here are a few steps:
+
+* Enable Bluetooth connection in the Settings, or connect via Lightning connector.
```
vquach2404, line 45:
```suggestion
* Enable Bluetooth connection in the device settings or connect via Lightning connector.
```
Is Lightning the correct term here? I think some iOS devices only support Thunderbolt.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -30,10 +34,23 @@ Tap "Start" to parse the NMEA sentences into a simulated location data source, a
...
+* Refer to the device manufacturer's documentation to get its protocol string, and add the protocol to the app’s `Info.plist` under the key `UISupportedExternalAccessoryProtocols`.
```
vquach2404, line 46:
```suggestion
* Refer to the device manufacturer's documentation to get its protocol string and add the protocol to the app’s `Info.plist` under the `UISupportedExternalAccessoryProtocols` key.
```

vquach2404 (commented):
Looking good! I made some wording suggestions.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
yo1995: Hmm good point. I'll change it to cable.

vquach2404 (approved):
Looks good to me!

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -26,29 +26,90 @@ class DisplayDeviceLocationWithNMEADataSourcesViewController: UIViewController {
...
     
     // MARK: Actions
     
-    @IBAction func start() {
+    @IBAction func chooseDataSource(_ sender: UIBarButtonItem) {
+        let alertController = UIAlertController(
+            title: "Choose an NMEA data source.",
+            message: nil,
+            preferredStyle: .actionSheet
+        )
```
philium, line 72: Is there a reason why the sample can't use the bluetooth accessory picker ([`EAAccessoryManager.showBluetoothAccessoryPicker(withNameFilter:completion:)`](https://developer.apple.com/documentation/externalaccessory/eaaccessorymanager/1613913-showbluetoothaccessorypicker))? According to the documentation for `EAAccessoryManager.connectedAccessories`, "Because the contents of this property can change dynamically based on the connection and disconnection of accessories, you should not cache the value of this property.", which is kind of what this sample is doing by creating a static alert view. Presumably the picker stays up-to-date as accessories come and go. Also, the documentation for `EAAccessory.protocolStrings` says "When encountering a new accessory, always check the list of supported protocols for the one you expect to use before creating a session. " So according to the documentation, the protocol check should occur after connection but before creating a session, not while discovering accessories.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -26,29 +26,90 @@ class DisplayDeviceLocationWithNMEADataSourcesViewController: UIViewController {
...
+            }
+            alertController.addAction(action)
+        }
+        // Add mock data source to the options.
+        let mockDataSourceAction = UIAlertAction(title: "Mock Data", style: .default) { [self] _ in
+            nmeaLocationDataSource = AGSNMEALocationDataSource(receiverSpatialReference: .wgs84())
+            nmeaLocationDataSource!.locationChangeHandlerDelegate = self
+            mockNMEADataSource.delegate = self
+            start()
+        }
```
philium, line 95: Instead of trying to make the mock data appear as an accessory, it might be better to first prompt the user for what _kind_ of data source they want to use: accessory or mock data.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -107,14 +168,32 @@ class DisplayDeviceLocationWithNMEADataSourcesViewController: UIViewController {
...
+        let verticalAccuracy = Measurement(
+            value: nmeaLocation.verticalAccuracy,
+            unit: UnitLength.meters
+        )
+        let accuracyText = String(
+            format: "Accuracy - Horizontal: %@; Vertical: %@",
+            distanceFormatter.string(from: horizontalAccuracy),
+            distanceFormatter.string(from: verticalAccuracy)
+        )
+        statusLabel.text = accuracyText + "\n" + satelliteInfoText
```
philium, line 194:
```suggestion
        statusLabel.text = """
        \(accuracyText)
        \(satelliteInfoText)
        """
```

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -123,13 +202,12 @@ extension DisplayDeviceLocationWithNMEADataSourcesViewController: AGSNMEALocatio
...
             .joined(separator: ", ")
-        let statusText =
-            """
-            \(satellites.count) satellites in view
-            System(s): \(satelliteSystemsText)
-            IDs: \(idText)
-            """
-        statusLabel.text = statusText
+        satelliteInfoText = String(
+            format: "%d satellites in view\nSystem(s): %@\nIDs: %@",
```
philium, line 206:
```suggestion
            format: """
            %d satellites in view
            System(s): %@
            IDs: %@
            """,
```

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -26,29 +26,90 @@ class DisplayDeviceLocationWithNMEADataSourcesViewController: UIViewController {
...
+        return protocols
+    }()
+    
     // MARK: Instance properties
     
     /// An NMEA location data source, to parse NMEA data.
-    let nmeaLocationDataSource = AGSNMEALocationDataSource(receiverSpatialReference: .wgs84())
+    var nmeaLocationDataSource: AGSNMEALocationDataSource?
+    /// A string to hold the latest satellite info.
+    var satelliteInfoText = "Satellites info will be shown here."
```
philium, line 52: In general it is best not to pre-compute strings displayed to the user, particularly if the string is locale-dependent (ie. was created from a formatter) as information in the locale may have changed before the string is actually displayed onscreen. For example, a user crossing into a different time zone or region. That may not be an issue if the string is displayed right away, but then why have a property for it? Another thing to consider is that when a label's text is updated, the entire label needs to be redrawn, even if only a small part of that text changed. Instead of trying to put multiple pieces of information in a single label, it would be better to have a separate label for each piece. That way, when on one piece of information changes (like the horizontal accuracy, for example), only that label needs to be re-drawn. Does that make sense?

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -6,18 +6,22 @@ Parse NMEA sentences and use the results to show device location on the map.
 
 ## Use case
 
-NMEA sentences can be retrieved from GPS receivers and parsed into a series of coordinates with additional information. Devices without a built-in GPS receiver can retrieve NMEA sentences by using a separate GPS dongle, commonly connected bluetooth or through a serial port.
+NMEA sentences can be retrieved from GNSS/GPS receivers and parsed into a series of coordinates with additional information. Devices without a built-in GNSS receiver can retrieve NMEA sentences by using a separate GNSS dongle, commonly connected through Bluetooth or a serial port.
```
philium, line 9: Is this section supposed to be generic to any platform? It just seems odd that it references "a built-in GNSS receiver" and "a serial port" when iOS devices have neither of those things.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -30,10 +34,23 @@ Tap "Start" to parse the NMEA sentences into a simulated location data source, a
...
 
 The route taken in this sample features a [2-minute driving trip around Redlands, CA](https://arcgis.com/home/item.html?id=d5bad9f4fee9483791e405880fb466da).
 
+## Additional information
+
+To support GNSS device connection in an app, here are a few steps:
+
+* Enable Bluetooth connection in the device settings or connect via cable connection.
+* Refer to the device manufacturer's documentation to get its protocol string and add the protocol to the app’s `Info.plist` under the `UISupportedExternalAccessoryProtocols` key.
+* When working with any MFi accessory, the end user must register their iOS app with the accessory manufacturer first to whitelist their app before submitting it to the AppStore for approval. This is a requirement by Apple and stated in the iOS Developer Program License Agreement.
```
philium, line 47: App Store is two words.

```suggestion
* When working with any MFi accessory, the end user must register their iOS app with the accessory manufacturer first to whitelist their app before submitting it to the App Store for approval. This is a requirement by Apple and stated in the iOS Developer Program License Agreement.
```

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
yo1995: Oh 🤦  I vaguely remember there is a system built-in UI for picking Bluetooth device, but an earlier search yields no result for me (maybe I'm just not using the right keywords). So thanks for pointing this out!

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
yo1995: I slacked off here 😅  will use 2 labels for different text!

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -26,29 +26,90 @@ class DisplayDeviceLocationWithNMEADataSourcesViewController: UIViewController {
...
+    // MARK: Constants
+    
+    /// The protocols specified in the `Info.plist` that the app uses to
+    /// communicate with external accessory hardware.
+    let supportedProtocolStrings: [String] = {
+        guard let protocols = Bundle.main.object(forInfoDictionaryKey: "UISupportedExternalAccessoryProtocols") as? [String] else {
+            return []
+        }
+        return protocols
+    }()
```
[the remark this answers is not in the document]
yo1995: Hmm good point. In this case, do you suggest that I should also hardcode a list of NMEA protocols in this sample, and use them to check the connected BT devices?

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
yo1995: > Another thing to consider is that when a label's text is updated, the entire label needs to be redrawn, even if only a small part of that text changed. Instead of trying to put multiple pieces of information in a single label, it would be better to have a separate label for each piece.

In terms of this issue - I totally agree it is the right approach, but am a bit reluctant to do so. It may improve the performance of the sample, but would make it harder to read the source code as the users need to find out the mappings between labels and static/mutating text pieces. In contrast, putting all the prompt text in a single string and only assign to one label is clearer IMO.

It's pushing me to think of Swift UI's `Text`... 🤐

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -24,31 +24,129 @@ class DisplayDeviceLocationWithNMEADataSourcesViewController: UIViewController {
...
+                            presentAlert(message: "Selecting an accessory failed for an unknown reason.")
+                        }
+                    } else if let (accessory, protocolString) = firstSupportedAccessoryWithProtocol() {
+                        // Proceed with supported and connected accessory, and
+                        // ignore other accessories that aren't supported.
+                        accessoryDidConnect(connectedAccessory: accessory, protocolString: protocolString)
+                    }
+                }
+            }
+        }
```
yo1995, line 124: A few things to notice (and feedback needed)

- The 2 helper method names: `firstSupportedAccessoryWithProtocol` and `accessoryDidConnect` might not be the best naming pattern, but are the best that I can came up with 😅 
- Since there is no explicit disconnect operation from the iOS side, we have to check if the phone is already connected to some BT device. There are several states:
    - Not connected. Then we can prompt the BT picker
    - Connected, but to some other BT device such as headphone. Will need to present picker here.
    - Connected to a GNSS surveyor. Then we can directly instantiate the NMEA LDS from that accessory.
- Do we need the error code handling? I found the generic error message not helpful, but there aren't error description associated with these codes, so I copy and pasted from [here](https://developer.apple.com/documentation/externalaccessory/eabluetoothaccessorypickererror/code)
- Please revisit the previous comment on hardcoding protocol strings: https://github.com/Esri/arcgis-runtime-samples-ios/pull/1058#discussion_r616010708

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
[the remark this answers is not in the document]
philium: I think what you have is fine for now. Just something to be aware of.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
philium: > It may improve the performance of the sample, but would make it harder to read the source code

The suggestion ("Instead of trying to put multiple pieces of information in a single label, it would be better to have a separate label for each piece.") implies that each piece of information is updated independently. When all pieces are updated at the same time, then yes, it does make sense to put them in the same label. I think what you've done with having a separate label for the accuracy is a good improvement and putting everything else in the other label makes sense.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -6,18 +6,22 @@ Parse NMEA sentences and use the results to show device location on the map.
...
 
-Tap "Start" to parse the NMEA sentences into a simulated location data source, and initiate the location display. Tap "Recenter" to recenter the location display. Tap "Reset" to reset the location display.
+Tap "Source" to choose between a simulated location data source or any data source created from a connected GNSS device, and initiate the location display. Tap "Recenter" to recenter the location display. Tap "Reset" to reset the location display and location data source.
 
 ## How it works
 
-1. Load NMEA sentences from a local file.
-2. Parse the NMEA sentence strings, and push data into `AGSNMEALocationDataSource`.
+1. Load NMEA sentences.
+    * If a supported GNSS receiver is connected, the sample can get NMEA updates from it.
```
philium, line 20:
```suggestion
    * If a supported GNSS accessory is connected, the sample can get NMEA updates from it.
```

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -6,18 +6,22 @@ Parse NMEA sentences and use the results to show device location on the map.
...
+    * Otherwise, the sample will read mock data from a local file.
+2. Create an `AGSNMEALocationDataSource`. There are 2 ways to provide updates to the data source.
+    * When updates are received from a GNSS device or the mock data provider, push the data into `AGSNMEALocationDataSource`.
```
philium, line 23:
```suggestion
    * When updates are received from a GNSS accessory or the mock data provider, push the data into `AGSNMEALocationDataSource`.
```

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -6,18 +6,22 @@ Parse NMEA sentences and use the results to show device location on the map.
...
+    * With the Runtime SDK 100.11.0 release, you can create an `AGSNMEALocationDataSource` with the GNSS device connected via `EAAccessory` and its protocol specified in `UISupportedExternalAccessoryProtocols`. The data source created this way will automatically get updates from the device instead of requiring to push data explicitly.
```
philium, line 24:
```suggestion
    * Starting with Runtime SDK 100.11, you can initialize `AGSNMEALocationDataSource` with a GNSS accessory. The data source created this way will automatically get updates from the device instead of requiring to push data explicitly.
```

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -30,10 +34,23 @@ Tap "Start" to parse the NMEA sentences into a simulated location data source, a
 
 ## About the data
 
-A list of NMEA sentences is used to initialize a `SimulatedNMEADataSource` object. This simulated data source provides NMEA data periodically, and allows the sample to be used on devices without a GPS dongle that produces NMEA data.
+A list of NMEA sentences is used to initialize a `SimulatedNMEADataSource` object. This simulated data source provides NMEA data periodically and allows the sample to be used on devices without a GNSS receiver.
```
philium, line 37:
```suggestion
A list of NMEA sentences is used to initialize a `SimulatedNMEADataSource` object. This simulated data source provides NMEA data periodically and allows the sample to be used without a GNSS accessory.
```

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -30,10 +34,23 @@ Tap "Start" to parse the NMEA sentences into a simulated location data source, a
...
 
 The route taken in this sample features a [2-minute driving trip around Redlands, CA](https://arcgis.com/home/item.html?id=d5bad9f4fee9483791e405880fb466da).
 
+## Additional information
+
+To support GNSS device connection in an app, here are a few steps:
```
philium, line 43:
```suggestion
To support GNSS accessory connection in an app, here are a few steps:
```

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/README.md
```diff
@@ -30,10 +34,23 @@ Tap "Start" to parse the NMEA sentences into a simulated location data source, a
...
+
+Please read Apple's documentation below for further details.
+
+* [`EAAccessory`](https://developer.apple.com/documentation/externalaccessory)
+* [`UISupportedExternalAccessoryProtocols`](https://developer.apple.com/documentation/bundleresources/information_property_list/uisupportedexternalaccessoryprotocols)
+
 ## Tags
 
-dongle, GPS, history, navigation, NMEA, real-time, trace
+Bluetooth, dongle, GNSS, GPS, history, navigation, NMEA, real-time, trace
```
philium, line 56: I would suggest replacing `dongle` with `accessory`. You might also add `MFi`.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSources.storyboard
```diff
@@ -22,33 +22,42 @@
...
                                         <fontDescription key="fontDescription" type="system" pointSize="17"/>
                                         <nil key="textColor"/>
                                         <nil key="highlightedColor"/>
                                     </label>
+                                    <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Satellites info will be shown here." lineBreakMode="wordWrap" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="18n-Ym-Ezn">
+                                        <rect key="frame" x="20" y="20.5" width="374" height="20.5"/>
+                                        <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                        <nil key="textColor"/>
+                                        <nil key="highlightedColor"/>
+                                    </label>
```
philium, line 38: You could put these labels in a stack view if you wanted to. That would simplify the constraints a bit.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -24,31 +24,129 @@ class DisplayDeviceLocationWithNMEADataSourcesViewController: UIViewController {
...
+        )
+        // Add real data source to the options.
+        let realDataSourceAction = UIAlertAction(title: "Device", style: .default) { [self] _ in
+            if let (accessory, protocolString) = firstSupportedAccessoryWithProtocol() {
+                // Use the supported accessory directly if it's already connected.
+                accessoryDidConnect(connectedAccessory: accessory, protocolString: protocolString)
+            } else {
+                // Display an alert to pair the device with a Bluetooth accessory.
+                EAAccessoryManager.shared().showBluetoothAccessoryPicker(withNameFilter: nil) { error in
+                    if let error = error as? EABluetoothAccessoryPickerError {
```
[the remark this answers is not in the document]
yo1995: > ...the only kind of error returned is EABluetoothAccessoryPickerError

Yes, per the [`EABluetoothAccessoryPickerError`](https://developer.apple.com/documentation/externalaccessory/eabluetoothaccessorypickererror) doc.

> If the error code is alreadyConnected, we should just be able to use the connected accessory, correct?

Yes. But instead of trying to init the NMEA LDS after an `alreadyConnected` error code is returned, I attempted to init the LDS from potential connected accessory before showing the picker. Not sure if it is the best practice...

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -24,31 +24,129 @@ class DisplayDeviceLocationWithNMEADataSourcesViewController: UIViewController {
...
+                // Skip any device which protocol is not included in the plist.
+                // This typically shouldn't happen, unless the device requires
+                // additional configuration.
+                continue
+            }
+            // Only return the first connected and supported accessory.
+            return (accessory, protocolString)
+        }
+        return nil
+    }
```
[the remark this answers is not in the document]
yo1995: Chatted with Eric Ito:

> Yes it is possible. We only use the one that emits NMEA but some devices support multiple and you could, in theory, read from them.

It goes back to https://github.com/Esri/arcgis-runtime-samples-ios/pull/1058#discussion_r616010708 and I think the future-proof approach would be to hardcode them in the sample, so that we only have 1 NMEA protocol for each device.

Was your comment cut at the end? I'm curious for what after "if not..." 😳

yo1995:
Code isn't ready for another review yet; will hardcode the protocol strings in the sample if no objection, and simplify the logic for the BT picker & error handling.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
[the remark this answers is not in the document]
philium: I started to say say something, but then changed my mind and apparently forgot to retract what I started to say! I was going to say that if not, we may need a more deterministic method of picking protocols as I don't know whether the order of `protocolStrings` is deterministic.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
```diff
@@ -24,31 +24,130 @@ class DisplayDeviceLocationWithNMEADataSourcesViewController: UIViewController {
...
-        startBarButtonItem.isEnabled = false
-        resetBarButtonItem.isEnabled = true
+    /// Get the first connected and supported Bluetooth accessory with its
+    /// protocol string.
+    /// - Returns: A tuple of the accessory and its protocol,
+    ///            or nil if no supported accessory exists.
+    func firstSupportedAccessoryWithProtocol() -> (EAAccessory, String)? {
+        for accessory in EAAccessoryManager.shared().connectedAccessories {
+            // The protocol string to establish the EASession.
+            guard let protocolString = accessory.protocolStrings.first(where: { supportedProtocolStrings.contains($0) }) else {
```
philium, line 74: Assuming that `supportedProtocolString` is ordered from best to worst:

```suggestion
            guard let protocolString = supportedProtocolStrings.first(where: { accessory.protocolStrings.contains($0) }) else {
```

yo1995: Actually I don't know what is the best order here, so I'll probably apply your suggestion. 🤔  It doesn't matter for most of the time, as we only expected 1 supported accessory connected each time.

Since I moved to hardcode the protocol strings in the sample, it is guaranteed that each type of NMEA accessory will only have 1 protocol for this sample. So it doesn't involve the issue in Apple doc for `EAAccessory.protocolStrings`

> If your application supports multiple protocols for a single accessory, your code should always choose the highest-fidelity protocol that you support.

- Using `accessory.protocolStrings.first(where: { supportedProtocolStrings.contains($0) })`

Will use the first available connected accessory, ordered by the order of `EAAccessoryManager.shared().connectedAccessories` (probably least recently connected? I don't know)

- Using `supportedProtocolStrings.first(where: { accessory.protocolStrings.contains($0) })`

Will use the first accessory ordered by brand: BadElf, EOS, SXBlue.

philium (approved):

yo1995:
Hi Divesh, @dg0yal please take a look at the README and share your thoughts! 

I'll also send a copy of the README to Brett from BadElf to get some feedback.

arcgis-ios-sdk-samples/Maps/Display device location with NMEA data sources/DisplayDeviceLocationWithNMEADataSourcesViewController.swift
yo1995: @philium Today when I was trying the BT picker with a newly created Xcode 12.5 project, I found it not showing when `SceneDelegate` methods are present. Searched online and someone [mentioned](https://developer.apple.com/forums/thread/128044?answerId=652027022#652027022) `UIAlertView`, which I don't quite understand 🤔 ...

Anyway, it isn't an problem for the Samples app for now, but might worth a revisit in the future.

saratchandrakarumuri (approved):

yo1995:
I'll merge this PR now as Brett has limited availability to review the PR. But he said he'll jot down a few suggestions when available, and we may make an update then.

PR #1061, opened 2021-04-16
Logistics: Add gitattributes to exclude paths from linguist detection

yo1995:
image

[Linguist](https://github.com/github/linguist) is the GitHub tool for summarizing repo language stats.

By adding a `.gitattributes` configuration file and merge it onto `main` branch in U11, we'll have a fully Swift repo, 100% orange bar.
IMO it would make our repo look more professional and up-to-date.

Doc here for excluded items: https://github.com/github/linguist/blob/master/docs/overrides.md#summary

Do you think this is necessary/acceptable?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+Scripts linguist-detectable=false
+"arcgis-ios-sdk-samples/Content Display Logic/Js" linguist-vendored


philium (approved):

vquach2404:
Would making this change mean that the language stats are inaccurate?

yo1995:
> Would making this change mean that the language stats are inaccurate?

Good point. I've conflicting thoughts on making this change:

- Pros: 
    - As described above, a more "Swift-y" looking repo, since we are primarily demonstrating Swift samples.
    - When the additional scripts grow larger, it won't take up more percentage in the stats
- Cons:
    - Not accurate reflection of the content in the repo. Because we have these DevOps scripts in the repo, they should show as a small portion of the source code. (though they have nothing to do with our users, who won't set up the repo exactly the same way as us.)
    - Added complexity and file count for future repo maintenance. The `.gitattribute` file will show up in the root dir of the repo, which makes it looks busy
    - Too trivial. I have to admit this is a trivial change, so if it doesn't make sense I can just close it 😅

vquach2404:
Thank you for the detailed response! Although I agree it would be a better depiction of the sample viewer itself, I'm concerned that it would misrepresent the repo or confuse users when they come across non-Swift files. Is this common practice for professional repos? If so, I'm on board; otherwise, I would vote no on this.

yo1995:
> Thank you for the detailed response! Although I agree it would be a better depiction of the sample viewer itself, I'm concerned that it would misrepresent the repo or confuse users when they come across non-Swift files. Is this common practice for professional repos? If so, I'm on board; otherwise, I would vote no on this.

There are a few examples, and I think this [PR](https://github.com/apple/swift-evolution/pull/1074) (Matt, NSHipster, yay!) explains it well. 

Did a search in Google and Apple repos

- https://github.com/search?q=org%3Agoogle+linguist&type=code
- https://github.com/search?q=org%3Aapple+linguist&type=code

and the only substantial usage is in https://github.com/apple/swift-nio . Given that it is not a popular practice, I feel it is better not to make this change.

yo1995:
After merging to `main`, it actually doesn't look too bad!

image

I think this proportion exactly depicted my (our) daily development language usage... 😆

PR #1067, opened 2021-05-07
Update: Use basemap API key instead of global key for Create and save map

yo1995:
## Description

This PR address the issue in `common-samples/issues/2613`, to resolve the potential API key and Named User conflict on creating a portal item, by

- First unassign the environment API key, and keep it in a property (or should I do that in `init()` instead of `viewDidLoad`?)
- Whenever a basemap needs to be accessed, use the API key on the basemap
- When it requires to login to save the map to a user's account, don't use the API key and instead use Named User
- When the sample exits, restore the API key on the environment

## Known Issue

The current `Create and save map` sample works on iOS even if the same sample doesn't work on other platforms. Trevor's comment: `2613#issuecomment-2836473` sums up the the behavior and situation very well. To be more defensive, we tend to make this change on all platforms.

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
@@ -47,7 +47,8 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
             "CreateOptionsViewController",
             "SaveAsViewController"
         ]
-        
+        // Temporarily remove the API key for this sample.
+        // Please see the additional information in the README.
         apiKey = AGSArcGISRuntimeEnvironment.apiKey
         AGSArcGISRuntimeEnvironment.apiKey = ""
         
--- a/arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
@@ -36,7 +36,6 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
     @IBOutlet private weak var mapView: AGSMapView!
     
     private var portal: AGSPortal?
-    private var apiKey: String!
     
     override func viewDidLoad() {
         super.viewDidLoad()
@@ -47,10 +46,6 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
             "CreateOptionsViewController",
             "SaveAsViewController"
         ]
-        // Temporarily remove the API key for this sample.
-        // Please see the additional information in the README.
-        apiKey = AGSArcGISRuntimeEnvironment.apiKey
-        AGSArcGISRuntimeEnvironment.apiKey = ""
         
         // Auth Manager settings
         let config = AGSOAuthConfiguration(portalURL: nil, clientID: "xHx4Nj7q1g19Wh6P", redirectURL: "iOSSamples://auth")
@@ -59,10 +54,6 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
         
         // Initially show the map creation UI.
         performSegue(withIdentifier: "CreateNewSegue", sender: self)
-    }
-    
-    deinit {
-        AGSArcGISRuntimeEnvironment.apiKey = apiKey
     }
     
     private func showSuccess() {
@@ -115,7 +106,6 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
     
     func createOptionsViewController(_ createOptionsViewController: CreateOptionsViewController, didSelectBasemap basemap: AGSBasemap, layers: [AGSLayer]) {
         // Create a map with the selected basemap.
-        basemap.apiKey = apiKey
         let map = AGSMap(basemap: basemap)
         
         // Add the selected operational layers.
@@ -144,6 +134,11 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
             // Also to cut on the size.
             let croppedImage: UIImage? = image?.croppedImage(CGSize(width: 200, height: 200))
             
+            // Temporarily unset the API key for this sample.
+            // Please see the additional information in the README.
+            let apiKey = AGSArcGISRuntimeEnvironment.apiKey
+            AGSArcGISRuntimeEnvironment.apiKey = ""
+            
             self.mapView.map?.save(as: title, portal: self.portal!, tags: tags, folder: nil, itemDescription: itemDescription, thumbnail: croppedImage, forceSaveToSupportedVersion: true) { [weak self] (error) in
                 // Dismiss progress hud.
                 UIApplication.shared.hideProgressHUD()
@@ -155,6 +150,8 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
                     }
                 }
             }
+            
+            AGSArcGISRuntimeEnvironment.apiKey = apiKey
         }
     }
 }
--- a/arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
@@ -36,6 +36,7 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
     @IBOutlet private weak var mapView: AGSMapView!
     
     private var portal: AGSPortal?
+    private var apiKey: String!
     
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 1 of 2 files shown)


vquach2404 (commented):
Looks good. Should the readme be updated to include the API changes?

yo1995:
> Looks good. Should the readme be updated to include the API changes?

I took a look at the README and it didn't specifically mention the authentication patterns, so I didn't make change there. Also in dotnet PR: https://github.com/Esri/arcgis-runtime-samples-dotnet/pull/840 they didn't make such change.

But I'm open to suggestions for README!

vquach2404:
Hm, maybe in "Additional information" you can add something like "Note: The app allows users to view the basemaps due to API keys but may require additional permissions to save the map on AGOL."

arcgis-ios-sdk-samples/Maps/Create and save a map/README.md
```diff
@@ -27,6 +27,10 @@ Select the basemap and layers you'd like to add to your map. Tap the "Save" butt
 * AGSMap.save
 * AGSPortal
 
+## Additional information
+
+The sample allows users to view the basemaps with the pre-baked API keys. To save the map to ArcGIS Online, please login with an ArcGIS identity/named user account.
```
yo1995, line 32:
```suggestion
The sample allows users to view the basemaps with the pre-baked API key. To save the map to ArcGIS Online, please login with an ArcGIS identity/named user account.
```

yo1995: I'd like to express the meaning of "app-wide key", "the key set on `ArcGISRuntimeEnvironment`", or "the default key" here. I don't think automated is the word that I'm looking for 😕

vquach2404: I like "default key'!

yo1995:
Please hold off reviewing this PR - there are new changes in the samples repo, which I'll apply here when finalized.

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
```diff
@@ -47,6 +48,9 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
             "SaveAsViewController"
         ]
         
+        apiKey = AGSArcGISRuntimeEnvironment.apiKey
+        AGSArcGISRuntimeEnvironment.apiKey = ""
```
philium, line 52: Is clearing the environment API key really necessary? The key set on the basemap should override the environment key anyway, right?

yo1995: To my understanding: it is to prevent the map getting saved to the account which created the API key.

According to Trevor: the expected behavior with API key being set on `AGSArcGISRuntimeEnvironment` is that the `Create and save map` sample will save the map to the key-issuing account (runtime SDK account) without asking for OAuth credentials. But that is not what we are seeing, and we don't know why 😕 , given that the API key should [behave similar to an OAuth token](https://developers.arcgis.com/documentation/mapping-apis-and-services/security/api-keys/).

Please see more discussion in the Epic issue 2613: `common-samples/issues/2613`

arcgis-ios-sdk-samples/Maps/Create and save a map/README.md
```diff
@@ -27,6 +27,10 @@ Select the basemap and layers you'd like to add to your map. Tap the "Save" butt
 * AGSMap.save
 * AGSPortal
 
+## Additional information
+
+In this sample, an API key is set directly on `AGSBasemap` objects rather than on the whole app using the `AGSArcGISRuntimeEnvironment` class. This is useful in a scenario where an individual developer is part of an organization within ArcGIS Online that uses an API key to access a range of `AGSBasemapStyle`s. In the case that an individual member of the organization wants to save a map locally to their account, and not that of the organization, they can set the organization's API key on the basemap, and log in to their own account when challenged.  The individual can then save the final map to their own ArcGIS Online account.
```
yo1995, line 32:
```suggestion
In this sample, an API key is set directly on `AGSBasemap` objects rather than on the whole app using the `AGSArcGISRuntimeEnvironment` class. This is useful in a scenario where an individual developer is part of an organization within ArcGIS Online that uses an API key to access a range of `AGSBasemapStyle`s. In the case that an individual member of the organization wants to save a map locally to their account, and not that of the organization, they can set the organization's API key on the basemap, and log in to their own account when challenged. The individual can then save the final map to their own ArcGIS Online account.
```
extra space

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
yo1995: Also @TADraeseke to chime in here.

TADraeseke: We didn't clear the key like this on Android or Java--it shouldn't be necessary unless there's something I'm missing.

yo1995: Discussed this during Friday review, and it seems Android/Java have different Sample Viewer architecture. To my understanding Android doesn't set an environment key as each sample is sort of standalone, thus they don't clear the key in init. Meanwhile, dotnet does clear the key https://github.com/Esri/arcgis-runtime-samples-dotnet/pull/840#discussion_r562973021 on init, and reload it on destroy.

philium: Thank you @yo1995 and @TADraeseke; that makes sense. However, since this change is only needed when saving the map to the portal, could it be done before the call to `save(as:portal:tags:folder:itemDescription:thumbnail:forceSaveToSupportedVersion:completion:)` and undone in the completion handler? That way no `apiKey` property is needed and `viewDidLoad()` isn't doing hokey stuff. While quite common, it generally isn't a good idea to perform actions in `viewDidLoad()` and undo them in `deinit`. For one thing, `viewDidLoad()` isn't guaranteed to be called. It is possible (though rare) for a view controller to be instantiated but never presented, in which case deinit would be called without `viewDidLoad()` being called. Also, in most cases, the work being performed in `viewDidLoad()` has nothing to do with the view and therefore would be better off in the initializer itself.

yo1995: > could it be done before the call to `save...` and undone in the completion handler?

Actually, can I even recover the key immediately after `save` is called (https://github.com/Esri/arcgis-runtime-samples-ios/pull/1067/commits/f371652d928b28c20aa1d94e98031ef1870a557b), as there is a chance that it never calls back due to connection issue or so? Also, please see my comment below.

> ... deinit would be called without viewDidLoad() being called.

Yeah that is my fear too 🤕

yo1995:
So this afternoon I've been doing research on this topic using [`mitmproxy`](https://github.com/mitmproxy/mitmproxy), and had some interesting findings. I will try to keep it succinct here but feel free to ask me to elaborate!

**Conclusion**: with the current REST API endpoint and iOS SDK, it seems there is no need to make this change - no conflict between API key and Named user OAuth sign-in

**Reason**: during the POST request to `addItem` endpoint, no token is included in the request body or the form payload, hence no conflict between API key and OAuth token.

**Ambiguity**: The [API key documentation page](https://developers.arcgis.com/documentation/mapping-apis-and-services/security/api-keys/#using-your-api-key) writes:

> API endpoints on ArcGIS Platform support a token parameter that will accept either an API key or an OAuth 2.0 token, and to which you can pass your API key.

According to my experiments, only the GET requests pass the API key as the token parameter; saving map with POST doesn't use token. (Maybe there is special handling within the SDK?)

**Possible next steps**:

- ✅  Review and merge the PR with what Phil suggested: https://github.com/Esri/arcgis-runtime-samples-ios/pull/1067#discussion_r633618703 , as it does no harm to the validity of this sample
- Revert any changes to the code, but rephrase and keep the README updates
- Revert all changes and close this PR

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
```diff
@@ -145,6 +150,8 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
                     }
                 }
             }
+            
+            AGSArcGISRuntimeEnvironment.apiKey = apiKey
```
philium, line 154: It seems to me like this should go inside the completion handler. The completion handler should always get called as in the event of connection issues, the request should time out.

philium (approved):

yo1995:
Hi Sarat, feel free to give it a test! It should not change the behavior of this sample.

saratchandrakarumuri (approved):

yo1995:
Discussed with other platforms, Mike and Divesh, and it seems to need a rework on this update. Will put back into draft. 🤕

PR #1068, opened 2021-05-11
Update: Add filter barrier to Perform valve isolation trace

yo1995:
## Description

This PR updates the `Perform valve isolation trace` sample for `common-samples/issues/2354`, that adds the capability to tap on map and add filter barriers in the network, in addition to specifying a category comparison to run the trace.

## How to test

- The original workflow using category comparison should still work
- The new workflow of adding filter barrier in the network will give different results
- The 2 workflows cannot exist at the same time: you can either specify a category comparison, or add points to the map. This is achieved by changing UI states and checking nil for `selectedCategory` property.

Since there are lots of changes in UI states and functionalities, please also do extensive holistic ad-hoc testing to see if any UI logic is wrong or it crashes.


yo1995:
Please don't review yet - still some bug 😅

yo1995:
Please hold off the review - the output is not consistent with what Jennifer provided in the design... though I feel it is a bug in the original implementation. 🤔

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -260,25 +351,126 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
+}
+
+// MARK: - AGSGeoViewTouchDelegate
+
+extension PerformValveIsolationTraceViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        if let identifyAction = identifyAction {
+            identifyAction.cancel()
+        }
+        identifyAction = mapView.identifyLayers(atScreenPoint: screenPoint, tolerance: 10, returnPopupsOnly: false) { [weak self] result, error in
```
yo1995, line 386: 🤔  Not sure if we should block user interaction here, as we are identifying features in online layers (which could be not responsive when the connection is bad)

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -180,75 +268,78 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
+            clearLayersSelection()
+            traceParameters.filterBarriers.removeAll()
+            parametersOverlay.graphics.removeAllObjects()
+            traceCompleted = false
+            selectedCategory = nil
+            // Add back the starting location.
+            addGraphic(for: startingLocationPoint, traceLocationType: "starting point")
+            mapView.setViewpointCenter(startingLocationPoint, scale: 3000, completion: nil)
+            // Set UI state.
+            setStatus(message: "Tap on the map to add filter barriers, or run the trace directly without filter barriers.")
```
vquach2404, line 328:
```suggestion
            setStatus(message: "Tap on the map to add filter barriers or run the trace directly without filter barriers.")
```
In the README you mentioned that the filter barriers are configured when a category is selected. I think that should be expressed here as well.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
vquach2404: That sounds like a good idea.

vquach2404 (commented):
Looks good overall! What're your thoughts about adding an undo button in case the user wants to remove a filter barrier or go back to the category selection workflow? The toolbar seems full already but the only other way to exit that workflow is to trace and reset or restart the sample.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
yo1995: I hope today's demo helped to clear stuff up! Please let me know if further edit is needed in these status messages. 🙂 

> Looks good overall! What're your thoughts about adding an undo button in case the user wants to remove a filter barrier or go back to the category selection workflow? The toolbar seems full already but the only other way to exit that workflow is to trace and reset or restart the sample.

I was thinking about adding a reset button, but don't want to overcrowd the toolbar. I tend to keep the current UI TBH 😆

vquach2404 (commented):
Currently, the user is able to add additional barriers after the trace has been completed but cannot trace it. Their only option is to reset and re-add the barriers. Is this expected? Maybe the map should not allow additional filter barriers after the trace has been completed.

performValve

yo1995:
> Maybe the map should not allow additional filter barriers after the trace has been completed.

I added a condition to abort identify when the trace is completed. Please give it a try!

vquach2404 (approved):
Looks good! Ready for Phil's review.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -18,140 +18,228 @@ import ArcGIS
...
+    
+    /// The base trace parameters.
+    let traceParameters = AGSUtilityTraceParameters(traceType: .isolation, startingLocations: [])
+    /// The utility category selected for running the trace.
+    var selectedCategory: AGSUtilityCategory?
+    /// An array of available utility categories with current network definition.
+    var filterBarrierCategories: [AGSUtilityCategory]!
+    /// An array to hold the gas line and gas device feature layers created from
+    /// the service geodatabase.
+    var layers: [AGSFeatureLayer]!
```
philium, line 67: These don't need to be optional.

```suggestion
    /// An array of available utility categories with current network definition.
    var filterBarrierCategories: [AGSUtilityCategory] = []
    /// An array to hold the gas line and gas device feature layers created from
    /// the service geodatabase.
    var layers: [AGSFeatureLayer] = []
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -18,140 +18,228 @@ import ArcGIS
...
+                return
+            }
+            // Create a default starting location.
+            guard let startingLocation = self.makeStartingLocation() else {
+                UIApplication.shared.hideProgressHUD()
+                self.presentAlert(message: "Failed to create starting location.")
+                self.setStatus(message: errorMessage)
+                return
+            }
+            self.utilityNetworkDidLoad(startingLocation: startingLocation)
```
philium, line 148: The spirit behind `guard` is that it checks for a condition that must be met in order for forward progress to be made (hence the `else` clause must exit the current scope). When work is done in an `else` clause, forward progress is being made either way, making the guard statement less intuitive and thereby making the code harder to reason about, not easier. In almost all cases, the `else` clause of `guard` statement should be a single line. If it isn't, that's a pretty good indication to use an `if` statement instead.

```suggestion
            if let error = error {
                UIApplication.shared.hideProgressHUD()
                self.presentAlert(error: error)
                self.setStatus(message: errorMessage)
            } else if let startingLocation = self.makeStartingLocation() {
                self.utilityNetworkDidLoad(startingLocation: startingLocation)
            } else {
                UIApplication.shared.hideProgressHUD()
                self.presentAlert(message: "Failed to create starting location.")
                self.setStatus(message: errorMessage)
            }
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -180,75 +268,78 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
             } else {
-                self.setStatus(message: "Trace with \(categoryName) category completed with no output.")
+                self.setStatus(message: "Trace completed with no output.")
+            }
+            completion()
+        }
+    }
+    
+    @IBAction func traceResetButtonTapped(_ button: UIBarButtonItem) {
+        switch traceCompleted {
```
philium, line 316: Please use and `if` statement instead of a `switch` statement.

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -18,140 +18,228 @@ import ArcGIS
...
+            UIApplication.shared.hideProgressHUD()
+            guard let self = self else { return }
+            if let features = features,
+               let feature = features.first,
+               let startingLocationPoint = feature.geometry as? AGSPoint {
+                // Get the geometry of the starting location as a point.
+                // Then draw the starting location on the map.
+                self.startingLocationPoint = startingLocationPoint
+                self.addGraphic(for: startingLocationPoint, traceLocationType: "starting point")
+                self.mapView.setViewpointCenter(startingLocationPoint, scale: 3000, completion: nil)
```
philium, line 168:
```suggestion
            mapView.setViewpointCenter(startingLocationPoint, scale: 3_000)
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -180,75 +268,78 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
+        case true:
+            // Reset the trace if it is already completed
+            clearLayersSelection()
+            traceParameters.filterBarriers.removeAll()
+            parametersOverlay.graphics.removeAllObjects()
+            traceCompleted = false
+            selectedCategory = nil
+            // Add back the starting location.
+            addGraphic(for: startingLocationPoint, traceLocationType: "starting point")
+            mapView.setViewpointCenter(startingLocationPoint, scale: 3000, completion: nil)
```
philium, line 326:
```suggestion
            mapView.setViewpointCenter(startingLocationPoint, scale: 3_000)
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -260,25 +351,131 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
+        loadServiceGeodatabase()
+    }
+}
+
+// MARK: - AGSGeoViewTouchDelegate
+
+extension PerformValveIsolationTraceViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        // Don't identify taps if trace has completed.
+        if traceCompleted { return }
```
philium, line 384: This would be a good place to use a `guard` statement.

```suggestion
        guard !traceCompleted else { return }
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -260,25 +351,131 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
+        if let identifyAction = identifyAction {
+            identifyAction.cancel()
+        }
```
philium, line 387:
```suggestion
        identifyAction?.cancel()
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -260,25 +351,131 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
+            let cancelAction = UIAlertAction(title: "Cancel", style: .cancel)
+            terminalPicker.addAction(cancelAction)
+            if let popoverController = terminalPicker.popoverPresentationController {
+                // If presenting in a split view controller (e.g. on an iPad),
+                // provide positioning information for the alert controller.
+                popoverController.sourceView = mapView
+                let tapPoint = mapView.location(toScreen: mapPoint)
+                popoverController.sourceRect = CGRect(origin: tapPoint, size: .zero)
+            }
+            present(terminalPicker, animated: true, completion: nil)
```
philium, line 476:
```suggestion
            present(terminalPicker, animated: true)
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/README.md
```diff
@@ -2,38 +2,46 @@
...
 2. Create and load an `AGSServiceGeodatabase` with a feature service URL and get tables with their layer IDs.
 3. Create an `AGSMap` object that contains `AGSFeatureLayer`(s) created from the service geodatabase's tables.
-4. Create and load an `AGSUtilityNetwork` with the same feature service URL and map.
-5. Create a default starting location from a given asset type and global ID.
-6. Add an `AGSGraphicsOverlay` with an `AGSGraphic` that represents this starting location.
-7. Populate the choice list for the filter barriers from the `categories` property of `AGSUtilityNetworkDefinition`.
-8. Get a default `AGSUtilityTraceConfiguration` from a given tier in a domain network. Set its `filter` property with an `AGSUtilityTraceFilter` object.
-9. When "Trace" is tapped,
+4. Create and load an `AGSUtilityNetwork` with the same feature service URL and map. Use `AGSGeoViewTouchDelegate.geoView(_:didTapAtScreenPoint:mapPoint:)` to get the `mapPoint` where a user tapped on the map.
+5. Create `AGSUtilityTraceParameters` with `isolation` trace type and aa default starting location from a given asset type and global ID.
```
philium, line 22:
```suggestion
5. Create `AGSUtilityTraceParameters` with `isolation` trace type and a default starting location from a given asset type and global ID.
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/PerformValveIsolationTraceViewController.swift
```diff
@@ -260,25 +351,131 @@ class PerformValveIsolationTraceViewController: UIViewController {
...
+            if self.categoryBarButtonItem.isEnabled {
+                self.categoryBarButtonItem.isEnabled = false
+                self.selectedCategory = nil
+            }
+            if !self.traceResetBarButtonItem.isEnabled {
+                self.traceResetBarButtonItem.isEnabled = true
+            }
+            self.traceParameters.filterBarriers.append(element)
+            let point = geometry as? AGSPoint ?? location
+            self.addGraphic(for: point, traceLocationType: PerformValveIsolationTraceViewController.filterBarrierIdentifier)
```
philium, line 422:
```suggestion
            self.addGraphic(for: point, traceLocationType: Self.filterBarrierIdentifier)
```

arcgis-ios-sdk-samples/Utility network/Perform valve isolation trace/README.md
yo1995: 🤦  🅰️ 🅰️

philium (approved):

yo1995:
There will be a patch to the UN service for this sample tomorrow, and it may change the trace output for filter barriers. So please hold off further review 😬

yo1995:
After discussing with Tony Wakim, he pointed out that we can add 3 filter barriers (1 additional barrier near the subnetwork controller), rather than 2, to show a more substantial result. But we didn't reach a conclusion on why the filter barriers behave the same as traditional barriers in terms of isolation trace vs connected trace.

Anyway, I'll update the second screenshot and request Sarat's review.

saratchandrakarumuri (approved):
looks good

vquach2404 (approved):

yo1995:
Since Phil [approved](https://github.com/Esri/arcgis-runtime-samples-ios/pull/1068#pullrequestreview-663574852) and the only change afterwards was the screenshot, I'll merge. 🎉

PR #1069, opened 2021-05-13
Update: 100.11.1 version numbers

yo1995:
This PR updates the version numbers, and get ready for the 100.11.1 App Store release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The ```main``` branch of this repository contains samples configured for the lat
 
 ## Requirements
 
-* [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/ios/) 100.11.1 (or newer)
+* [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/ios/) 100.11.0 (or newer)
 * [ArcGIS Runtime Toolkit for iOS](https://github.com/Esri/arcgis-runtime-toolkit-ios) 100.11.0 (or newer)
 * Xcode 12.0 (or newer)
 
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The ```main``` branch of this repository contains samples configured for the lat
 
 ## Requirements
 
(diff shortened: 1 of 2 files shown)


vquach2404 (approved):

README.md
```diff
@@ -25,7 +25,7 @@ The ```main``` branch of this repository contains samples configured for the lat
 
 ## Requirements
 
-* [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/ios/) 100.11.0 (or newer)
+* [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/ios/) 100.11.1 (or newer)
```
philium, line 28: I'm not sure this would need to change as the app still builds with 100.11.0 and we haven't bumped the dependency minimum version.

yo1995: Reasons I made this change:

- 100.11.0 SDK has the OAuth issue, that might cause the app to crash in debug scheme
- Toolkit's manifest is configured as `upToNextMinorVersion`, so it picks up the SDK SP minor version number update automatically. So while the toolkit SP stays on 100.11.0, the underlying SDK SP is 100.11.1.
- I plan to tag the next PR-to-`main` commit as `100.11.1` for this repo and upload to App Store after that, so I thought it would be good to at least have some version bump in the README. Otherwise it makes me feel like the minor version bump is due to some samples bug, rather than the dependencies.

Or do you prefer that I still upload the sample viewer to App Store with `100.11.0` version? If so then I don't think these changes are necessary.

philium: I think the other README changes make sense. Just this one about 100.11.1 being a _requirement_ doesn't make sense. Users could still use the app with the installed version of 100.11.0. If it is 100.11.1 is a requirement, we should include the dependency version number.

yo1995, line 28:
```suggestion
* [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/ios/) 100.11.0 (or newer)
```

yo1995: Suggestions applied.

philium (approved):

vquach2404 (approved):

Issue #980, opened 2020-10-07
Delegates not being called


aliawais:
AGSRouteTrackerDelegate methods are not getting called, following is SDK and other related information
Using ArcGIS-Runtime-SDK-iOS (100.7)
Xcode 11.7
iOS Simulator with iOS 13.7
iPhone X 13.4.1

Also, do we need to set data sources for real navigation? If yes, how to set instead of "AGSSimulatedLocationDataSource".

yo1995:
In terms of setting real location, consider using [AGSCLLocationDataSource](https://developers.arcgis.com/ios/latest/api-reference/interface_a_g_s_c_l_location_data_source.html)

aliawais:
@yo1995 Any example to how to set AGSCLLocationDataSource with other parameters as for AGSSimulatedLocationDataSource in the example.

Also the demo is not calling the delegates.

yo1995:
> @yo1995 Any example to how to set AGSCLLocationDataSource with other parameters as for AGSSimulatedLocationDataSource in the example.

[`Navigate in AR`](https://github.com/Esri/arcgis-runtime-samples-ios/blob/master/arcgis-ios-sdk-samples/Augmented%20reality/Navigate%20in%20AR/NavigateARNavigatorViewController.swift) is a similar sample that uses real world location (from Apple, `AGSCLLocationDataSource`) as data source.

While it doesn't need a list of simulated location points, it works pretty much the same as the `AGSSimulatedLocationDataSource` in terms of other parameters.

> Also the demo is not calling the delegates.

Which sample are you referring to? The simulated location data source API has been changed, so to build with current `master` and see how it works, you'll need 100.9 SDK.

@nixta for more help.

aliawais:
Thanks @yo1995 
We were using the sample code for the from ArcGIS Demo example that was using AGSSimulatedLocationDataSource, the example was working fine few weeks back, but now the same code is not calling delegates. I am not sure which ArcGIS version was initially used but current one is 100.7 in which the delegates are not called.

To implement for real world example, i would change the network analysis service, update stops to use real location and update data source in that example. But the example is not calling delegates so i am stuck. 

What is the minimum iOS version required for ArcGIS 100.9 as our app currently support iOS 11 and ArcGIS 100.9 does not support it.

aliawais:
@nixta 
Can you help in why the delegates are not being called with the specifications given in first message. If thats is fixed i can update the data source and and stop and it should work.

[...]

PR #1070, opened 2021-05-13
Fix: Add alert for route tracker licensing error

yo1995:
## Background

`common-samples/issues/2288`

User may use the route tracker sample code in the sample viewer. There are a few scenarios

- User is in developer mode without a license: the basemap will show the watermark, but all functionalities would work
- User uses a Advanced license string for deployment: all good
- User uses a Lite license string: in this case, since the route tracking is not supported by Lite license, the `AGSRouteTracker.trackLocation(_:completion:)` call would fail and give error. 
    However in our (my 🤕 ) previous implementations, we didn't handle that error and let the sample fail silently. That is due to my negligence of others using Lite license to build the app. See more in #980 

Better depict with a table

||`v.next`|this branch|
|-|-|-|
|Dev mode|✅ no error|✅ no error|
|Lite|❌ error not shown, silently fail|✅ Error handled with prompt|
|Advanced|✅ no error|✅ no error|

## Fix

Add error handling and comment in the code: By nilifying the route tracker, we can stop the route tracking.

This is not an elegant way to stop route tracker from giving a series of errors while location updates keep firing, but it works. Possible ways include:
    
- ✅  Nilify the route tracker object
- Nililfy the `locationChangeHandlerDelegate` on the data source, hence stop location updates.
- When error encountered, call the `reset` method in the VC, so that no more location updates are given. All 3 samples can be reset.

Also, we can show an alert instead of updating the status label to show this error. What do you think? (like this 👇 )

error

## How to test

- Swap the Advanced license key in the dev `.secrets` file with a Lite one (which can be acquired [here](https://developers.arcgis.com/ios/license-and-deployment/)), and compare `v.next` with this branch. The former will fail silently while the latter show the error in status label.

### DIFF
--- a/arcgis-ios-sdk-samples/Augmented reality/Navigate in AR/NavigateARNavigatorViewController.swift
+++ b/arcgis-ios-sdk-samples/Augmented reality/Navigate in AR/NavigateARNavigatorViewController.swift
@@ -249,7 +249,15 @@ extension NavigateARNavigatorViewController: AGSLocationChangeHandlerDelegate {
     func locationDataSource(_ locationDataSource: AGSLocationDataSource, locationDidChange location: AGSLocation) {
         startButtonItem.isEnabled = true
         // Send location updates to route tracker if actively navigating.
-        routeTracker?.trackLocation(location)
+        routeTracker?.trackLocation(location)  { [weak self] error in
+            guard let self = self else { return }
+            if let error = error {
+                // Display error message and stop further route tracking if it
+                // fails due to formatting or licensing issue.
+                self.setStatus(message: error.localizedDescription)
+                locationDataSource.locationChangeHandlerDelegate = nil
+            }
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


arcgis-ios-sdk-samples/Augmented reality/Navigate in AR/NavigateARNavigatorViewController.swift
```diff
@@ -249,7 +249,15 @@ extension NavigateARNavigatorViewController: AGSLocationChangeHandlerDelegate {
     func locationDataSource(_ locationDataSource: AGSLocationDataSource, locationDidChange location: AGSLocation) {
         startButtonItem.isEnabled = true
         // Send location updates to route tracker if actively navigating.
-        routeTracker?.trackLocation(location)
+        routeTracker?.trackLocation(location)  { [weak self] error in
```
philium, line 252:
```suggestion
        routeTracker?.trackLocation(location) { [weak self] error in
```

philium (changes requested):
Just one little tiny change requested.

arcgis-ios-sdk-samples/Augmented reality/Navigate in AR/NavigateARNavigatorViewController.swift
yo1995: My bad here 😅 . What do you think for nilifying `locationChangeHandlerDelegate` vs `routeTracker` in this case?

philium: I don't really know which is the better approach.

philium (approved):

vquach2404 (approved):

PR #1073, opened 2021-05-20
New: Display OGC API collection (OGC1)

yo1995:
This is the first of a series of 3 OGC samples. It shows an OGC feature layer

Please refer to `common-samples/issues/2059` for details.


yo1995:
Since Viv is out today, I'll directly go to the next phase of review... 😆

arcgis-ios-sdk-samples/Layers/Display OGC API collection/DisplayOGCAPICollectionViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+        populateFeaturesFromQuery()
+        // Query for features whenever map's viewpoint change has finished.
+        navigatingObservation = mapView.observe(\.isNavigating, options: .new) { _, change in
+            if change.newValue == false {
+                DispatchQueue.main.async {
+                    self.populateFeaturesFromQuery()
+                }
+            }
+        }
+    }
```
philium, line 106: A few things:

- Boolean values are, well, Boolean values. They should not be tested for equality using `== [true/false]`.
- The first call to `populateFeaturesFromQuery()` can be avoided (or rather, made implicit) by passing `initial` as one of the options.
- The change dictionary isn't needed as the new value can be read directly from the map view, as supplied by the closure. Doing so prevents the change dictionary from being created, which is a good thing if it isn't necessary.
- `self` could potentially go away before `populateFeaturesFromQuery()`, so the closure should weakly capture `self` so as not to keep it around longer than necessary.
- 
```suggestion
        // Query for features whenever the user stops interacting with the map
        // view.
        navigatingObservation = mapView.observe(\.isNavigating, options: .initial) { mapView, _ in
            guard !mapView.isNavigating else { return }
            DispatchQueue.main.async { [weak self] in
                self?.populateFeaturesFromQuery()
            }
        }
```

philium (changes requested):

arcgis-ios-sdk-samples/Layers/Display OGC API collection/DisplayOGCAPICollectionViewController.swift
yo1995: Great suggestions. 👍

philium (approved):

vquach2404 (approved):

saratchandrakarumuri (approved):
new sample looks good

PR #1074, opened 2021-05-25
New: Browse OGC API feature service (OGC2)

yo1995:
This is the second of a series of 3 OGC samples. It allows the user to load an OGC service from a URL and browse its layers.

Please refer to `common-samples/issues/2681` for more details.

Note: in addition to the default Daraa OGC API service, you can also use the other ones under https://demo.ldproxy.net/ to test, such as
- https://demo.ldproxy.net/strassen

I'll fix the Xcode project ordering (if there is any) after all 3 OGC samples have been approved!


yo1995 (commented):
Please see these comments below.

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIServicesViewController.swift
```diff
@@ -0,0 +1,224 @@
...
+        optionsViewController.popoverPresentationController?.barButtonItem = sender
+        present(optionsViewController, animated: true)
+    }
+    
+    // MARK: UIViewController
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Add the source code button item to the right of navigation bar.
+        (navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["BrowseOGCAPIFeatureServiceViewController"]
```
vquach2404, line 185: I cannot view the source code in the sample viewer. I think the file name is off here.

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+![Image of Browse OGC API feature service](browse-OGC-API-feature-service.png)
+
+## Use case
+
+OGC API standards are used for sharing geospatial data on the web. As an open standard, the OGC API aims to improve access to geospatial or location information and could be a good choice for sharing data internationally or between organizations. That data could be of any type, including, for example transportation layers shared between government organizations and private businesses.
+
+## How to use the sample
+
+Select a layer to display from the list of layers shown in an OGC API service.
```
vquach2404, line 13: I would suggest adding information about entering a valid URL here.

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## How it works
+
+1. Create an `AGSOGCFeatureService` object with a URL to an OGC API feature service.
```
vquach2404, line 17:
```suggestion
1. Create an `AGSOGCFeatureService` object with a URL of an OGC API feature service.
```

or 

```suggestion
1. Create an `AGSOGCFeatureService` object with an OGC API feature service URL.
```

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIServicesViewController.swift
```diff
@@ -0,0 +1,224 @@
...
+            // Set the viewpoint to Daraa, Syria.
+            mapView.setViewpoint(AGSViewpoint(latitude: 32.62, longitude: 36.10, scale: 20_000))
+        }
+    }
+    /// The bar button to browse feature layers.
+    @IBOutlet var browseBarButtonItem: UIBarButtonItem!
+    
+    // MARK: Properties
+    
+    let map = AGSMap(basemapStyle: .arcGISTopographic)
```
vquach2404, line 34: Is it possible to make this a local variable? Seems like `map` is only used in `loadLayers`

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIServicesViewController.swift
```diff
@@ -0,0 +1,224 @@
...
+                let featureLayer = AGSFeatureLayer(featureTable: table)
+                featureLayer.renderer = self.makeRenderer(for: table.geometryType)
+                self.map.operationalLayers.add(featureLayer)
+                self.mapView.setViewpointGeometry(extent, padding: 50, completion: nil)
+            }
+        }
+    }
+    
+    /// Create an appropriate renderer for a type of geometry.
+    func makeRenderer(for geometryType: AGSGeometryType) -> AGSSimpleRenderer? {
```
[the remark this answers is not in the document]
vquach2404: I vote to stick to the design. The blue renderer also helps identify layers with small highlighted areas and are more zoomed out

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIServicesViewController.swift
```diff
@@ -0,0 +1,224 @@
...
+        let cancelAction = UIAlertAction(title: "Cancel", style: .cancel) { _ in
+            NotificationCenter.default.removeObserver(textFieldObserver!)
+            completion(.failure(InputError.userCancelled))
+        }
+        // Validate and create URL from user input and remove observer.
+        let loadAction = UIAlertAction(title: "Load", style: .default) { _ in
+            NotificationCenter.default.removeObserver(textFieldObserver!)
+            guard let text = alertController.textFields![0].text?.trimmingCharacters(in: .whitespacesAndNewlines),
+                  let serviceURL = URL(string: text) else {
+                completion(.failure(InputError.invalidURL))
```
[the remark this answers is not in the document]
vquach2404: When an invalid URL is inputted, I think it'd be nice to prompt the user with another text field so they can try another URL. This would be useful for typos so the user wouldn't have to exit the sample to try a new URL.

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIServicesViewController.swift
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// MARK: - Custom errors
+
+extension BrowseOGCAPIFeatureServiceViewController {
+    private enum InputError: LocalizedError {
+        var errorDescription: String? {
+            switch self {
+            case .invalidURL:
+                return "Bad URL."
```
vquach2404, line 213:
```suggestion
                return "Invalid URL."
```

yo1995: I'll leave this to Phil.

- The SDK seems to give error with description "Bad URL" if the URL is 404
- I'm not sure about the best phrasing for a string that cannot be constructed into a URL 😅

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIServicesViewController.swift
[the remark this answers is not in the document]
yo1995: Please take a look at the new changes and share your thoughts. 🙂 

The code is a bit ugly because it's a recursive closure in this case. Also, do you think having a 2-step cancel makes sense?

yo1995:
Thanks for the suggestions, please take another look! 🤗

vquach2404 (dismissed):
Looks good! I'm in favor of the new changes!

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIFeatureService.storyboard
```diff
@@ -0,0 +1,63 @@
...
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <subviews>
+                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="u6U-yX-17o" userLabel="Map View" customClass="AGSMapView">
+                                <rect key="frame" x="0.0" y="44" width="414" height="769"/>
+                            </view>
+                            <toolbar opaque="NO" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="jXp-jp-Vzg">
+                                <rect key="frame" x="0.0" y="813" width="414" height="49"/>
+                                <items>
+                                    <barButtonItem style="plain" systemItem="flexibleSpace" id="KnP-M3-jrs"/>
+                                    <barButtonItem enabled="NO" title="Browse Layers" id="7Dh-s3-xHh">
```
philium, line 27: Should this button just be called `Layers`? It presents a UI more for _picking_ a layer, so more than just browsing (looking at) layers.

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIFeatureServiceViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+        // Set the feature request mode to manual (only manual is currently
+        // supported). In this mode, you must manually populate the table -
+        // panning and zooming won't request features automatically.
+        table.featureRequestMode = .manualCache
+        
+        lastQuery = table.populateFromService(
+            with: queryParameters,
+            clearCache: false,
+            outfields: nil
+        ) { [weak self, table] _, error in
```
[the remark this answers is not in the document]
yo1995: Searching with `.cancel()` within the project, I don't see other samples nilify the `AGSJob` (or `AGSCancelable`) when they complete. However, there are a few samples set the job to nil at the top, together with the `.cancel()` call.

Do you want me to
- Set the job to nil at the top of the method
- Set the job to nil in the completion handler
- Don't make a change

I'm not sure the impact of not setting it to nil here. Can you explain a bit more?
Thanks!

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIFeatureServiceViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+                    self.askUserForServiceURL(completion: completion)
+                }
+                let cancelAction = UIAlertAction(title: "Cancel", style: .cancel)
+                alertController.addAction(tryAgainAction)
+                alertController.addAction(cancelAction)
+                alertController.preferredAction = tryAgainAction
+                self.present(alertController, animated: true)
+            }
+        }
+        askUserForServiceURL(completion: completion)
```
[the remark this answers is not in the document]
yo1995: Ohh good point... I absolutely forgot that I shouldn't present in viewDidLoad 🤦  Let me redo the logic here...

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIFeatureServiceViewController.swift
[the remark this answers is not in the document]
yo1995: > I'm wondering if featureCollectionInfos should maybe be featureCollectionTables?

Not sure hmm. To me, holding an array of tables doesn't make a huge difference to holding infos. The design held the infos for the drop down list, so I followed.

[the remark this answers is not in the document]
philium: > I'm not sure the impact of not setting it to nil here.

It is simply a matter of being a good steward of the memory given to the app. At this point the cancellable is no longer needed, so it should be set to `nil` so that 1) It can be deallocated, and 2) so that `cancel()` isn't called on it the next time.

> there are a few samples set the job to nil at the top, together with the .cancel() call.

The important thing is to call `cancel()` on the current cancellable (if it exists) to stop the operation right away. Setting the associated property to `nil` is just extra work as the property's value is going to be updated with the cancellable for the new operation.

Does that make sense?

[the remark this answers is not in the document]
yo1995: Suggestion applied.

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIFeatureServiceViewController.swift
```diff
@@ -0,0 +1,246 @@
...
+            } else if let infos = self.service.serviceInfo?.featureCollectionInfos {
+                self.featureCollectionInfos = infos
+                self.layersBarButtonItem.isEnabled = true
+            }
+        }
+    }
+    
+    /// Load and display a feature layer from the OGC feature collection table.
+    /// - Parameter table: The `AGSOGCFeatureCollectionTable` selected by user.
+    func displayLayer(with table: AGSOGCFeatureCollectionTable) {
```
philium, line 74: Didn't this class take the info before? You can go back to that as there's really no difference between it taking the info or the table. I did do some testing and apparently _something_ (the service?) is holding onto the table, as it doesn't go away when the completion handler holds a weak reference to it. However, it would be good for the completion handler to hold an _unowned_ reference to the table.

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIFeatureServiceViewController.swift
```diff
@@ -0,0 +1,246 @@
...
+                    self.askUserForServiceURL(completion: completion)
+                }
+                let cancelAction = UIAlertAction(title: "Cancel", style: .cancel)
+                alertController.addAction(tryAgainAction)
+                alertController.addAction(cancelAction)
+                alertController.preferredAction = tryAgainAction
+                self.present(alertController, animated: true)
+            }
+        }
+        askUserForServiceURL(completion: completion)
```
philium, line 216: I think I understand this a little bit better now, but I'm still confused by a couple things. Shouldn't the retry mechanism be part of asking the user for the service URL (part of `askUserForServiceURL(completion:)`)? Then the completion handler could just take a `URL?` instead of a result. However, I'm not sure that retrying is even necessary. Instead of allowing the user to enter an invalid URL, when not just make that part of the text validation process? If a instance of `URL` can't be created from the text, don't enable the `Load` button. Would that work?

yo1995: I like the idea of using URL constructor to validate! Will make the change.

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIServicesViewController.swift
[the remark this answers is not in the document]
yo1995: Per Phil I'll remove the retry logic. I'll re-request your review when it's updated!

arcgis-ios-sdk-samples/Layers/Browse OGC API feature service/BrowseOGCAPIFeatureServiceViewController.swift
yo1995: Changing it back to take `info` as parameter, IMO the local `table` variable needs to be held strongly. 

I tried adding `unowned` and it seems to crash with completion handler. But I cannot be too sure if there are other objects that hold the table. Please let me know your thoughts!

yo1995:
Ready for review!

philium (approved):

vquach2404 (approved):

saratchandrakarumuri (approved):
sample looks good

PR #1076, opened 2021-05-27
New: Query OGC API with CQL filters (OGC3)

yo1995:
This is the last of a series of 3 OGC samples. It creates a CQL filter query to get features from the OGC service.

Please refer to `common-samples/issues/2682` for details.

Branch: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/New-OGC3-vNext/arcgis-ios-sdk-samples/Layers/Query%20with%20CQL%20filters

Please see the comments below. Also, on Friday 28th Ting will discuss the date picker conflict with CQL json stuff.


arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+    let sampleWhereClauses = [
+        // Sample Query 1, cql-text, cql-json.
+        "F_CODE = 'AP010'",
+        "{\"eq\":{\"property\":\"F_CODE\",\"value\":\"AP010\"}}",
+        // Sample Query 2.
+        "F_CODE LIKE 'AQ%'",
+        "{\"like\":{\"property\":\"F_CODE\",\"value\":\"AQ%\"}}",
+        // Sample Query 3: use cql-json to combine "before" and "after" temporal
+        // operators with the logcial "and" operator
+        "{\"and\":[{\"after\":{\"property\":\"ZI001_SDV\", \"value\":\"2011-12-31\"}},{\"before\":{\"property\":\"ZI001_SDV\",\"value\":\"2013-01-01\"}}]}"
```
yo1995, line 51: Note: The json query strings here are known to return 400 Bad Request. Zack and Hisham are looking into it. I'll keep this thread posted when I have updates.

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+    ]
+    var selectedWhereClause: String?
+    var maxFeatures: Int?
+    
+    // MARK: Methods and Actions
+    
+    func makeQueryParameters() -> AGSQueryParameters {
+        let queryParameters = AGSQueryParameters()
+        queryParameters.whereClause = selectedWhereClause ?? ""
+        queryParameters.maxFeatures = maxFeatures ?? 1_000
```
yo1995, line 61: These default values are following the design doc by dotnet

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
yo1995: Actually, I'm thinking if we should remove the jsons here for now, and make another PR to add them when they get fixed. So that this PR won't be blocked. 🤔  What do you think?

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersViewController.swift
```diff
@@ -0,0 +1,122 @@
...
+            "QueryWithCQLFiltersSettingsViewController"
+        ]
+        ogcFeatureLayer.load { [weak self] error in
+            guard let self = self else { return }
+            if let error = error {
+                self.presentAlert(error: error)
+            } else {
+                // Show a default query result.
+                let defaultQueryParameters = AGSQueryParameters()
+                defaultQueryParameters.maxFeatures = 1_000
```
vquach2404, line 101: Since the initial query is set to 1000 max features, I think the table view should display the default number upon loading as well. A user might not know to tap on "Max features" and having the default value displayed there may help.

vquach2404 (commented):
I think it's a little odd that there's a query upon loading the sample. I think it would make more sense if the user sees the default parameters and applies the query themselves. Doing so, provides context to the user as to what is being queried.

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+    @IBOutlet var startDatePicker: UIDatePicker!
+    /// The end date picker.
+    @IBOutlet var endDatePicker: UIDatePicker!
+    /// The switch to control if date pickers are enabled.
+    @IBOutlet var dateFilterSwitch: UISwitch!
+    
+    // MARK: Properties
+    
+    /// A delegate to notify other view controllers.
+    weak var delegate: QueryWithCQLFiltersSettingsViewControllerDelegate?
```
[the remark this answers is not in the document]
yo1995: It is also used here: https://github.com/Esri/arcgis-runtime-samples-ios/pull/1076/files#diff-841d0624c54921cc217fd76f73dc2e255a78813dedffcf75df80af4fbb181e40R111 so it cannot be a local variable.

yo1995:
> I think it's a little odd that there's a query upon loading the sample. I think it would make more sense if the user sees the default parameters and applies the query themselves. Doing so, provides context to the user as to what is being queried.

I agree with you. The design does load a default query on init though. 

What do you think if we just display an empty map on init, and let the user to provide the CQL filters themselves? Also I'll add the default query values as placeholders to the table rows.

vquach2404:
> What do you think if we just display an empty map on init, and let the user to provide the CQL filters themselves? Also I'll add the default query values as placeholders to the table rows.

Sounds good!

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
yo1995: Removed for now, and will add back once `completion(Zack's new CQL jsons)` 🤣

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
```diff
@@ -0,0 +1,176 @@
...
+        // Sample Query 1: Features with an F_CODE property of "AP010".
+        // A cql-text query.
+        "F_CODE = 'AP010'",
+        // A cql-json query.
+        "{ \"eq\" : [ { \"property\" : \"F_CODE\" }, \"AP010\" ] }",
+        // Sample Query 2: Features with an F_CODE property matching the pattern.
+        "F_CODE LIKE 'AQ%'",
+        // Sample Query 3: Use cql-json to combine "before" and "eq" operators
+        // with the logical "and" operator.
+        "{\"and\":[{\"eq\":[{ \"property\" : \"F_CODE\" }, \"AP010\"]},{ \"before\":[{ \"property\" : \"ZI001_SDV\"},\"2013-01-01\"]}]}"
```
vquach2404, line 56: I thought Zack said in the review to leave the json out. Did I misunderstand?

yo1995: He said to leave the *second* json out, which was the one with `like` that I've removed.

vquach2404 (dismissed):
Looks good!

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
```diff
@@ -0,0 +1,176 @@
...
+            self.selectedWhereClause = whereClause
+            self.whereClauseCell.detailTextLabel?.text = whereClause
+            self.navigationController?.popViewController(animated: true)
+        }
+        optionsViewController.title = "Where Clause"
+        show(optionsViewController, sender: self)
+    }
+    
+    /// Prompt an alert to allow the user to input a numeric value.
+    func showValueInputField(completion: @escaping (NSNumber) -> Void) {
```
philium, line 86: An alert really shouldn't be used for this. Instead, the cell should just have a text field that the user can edit directly. See `StretchRendererSettingsViewController` for example.

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFilters.storyboard
```diff
@@ -0,0 +1,264 @@
...
+                            <outlet property="delegate" destination="GDg-hx-dhZ" id="d6V-74-n5T"/>
+                        </connections>
+                    </tableView>
+                    <navigationItem key="navigationItem" title="CQL Filters" id="iQj-FT-f4r">
+                        <barButtonItem key="leftBarButtonItem" title="Cancel" id="AK8-Ni-KMZ">
+                            <connections>
+                                <action selector="cancelAction:" destination="GDg-hx-dhZ" id="ttA-SM-YIa"/>
+                            </connections>
+                        </barButtonItem>
+                        <barButtonItem key="rightBarButtonItem" title="Apply" id="OVi-4g-9SG">
```
philium, line 214: The style of this button should be Done:

image

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
```diff
@@ -0,0 +1,176 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+protocol QueryWithCQLFiltersSettingsViewControllerDelegate: AnyObject {
+    func settingsViewController(_ controller: QueryWithCQLFiltersSettingsViewController, queryParameters: AGSQueryParameters)
```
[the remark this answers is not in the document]
yo1995: Yup it makes perfect sense. I was hesitating on the choice of verb here and eventually forgot to use any label. 🤦

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
yo1995: I've made some changes, please let me know what you think!

- I didn't draw the text field in a custom cell, but added it as the accessory view to a `Basic` cell to avoid the complexity of setting up constraints.
- (as always 🤦 ) I couldn't find an elegant way to dismiss the `.numberPad` keyboard in a table view, so I cheated a bit to dismiss it with scroll. See in the `viewDidLoad` method: `tableView.keyboardDismissMode = .onDrag`
- The text will be silently reset to default value if the input is invalid.

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
```diff
@@ -0,0 +1,144 @@
...
+        endDatePicker.isEnabled = sender.isOn
+    }
+    
+    // MARK: UITableViewController
+    
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+        let cell = tableView.cellForRow(at: indexPath)
+        switch cell {
+        case whereClauseCell:
```
yo1995, line 133: Not sure if I should simply change it to `if` here, because we only have 1 select-able row in the static table.

philium: Normally I would suggest that the selection action should be configured in the storyboard, but that would be a significant change at this point (moving away from using `OptionsTableViewController` and designing the where clause picker in Interface Builder). What you have is fine.

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
```diff
@@ -0,0 +1,144 @@
...
+            showWhereClausePicker()
+        default:
+            return
+        }
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        tableView.keyboardDismissMode = .onDrag
```
philium, line 142: This can be set through Interface Builder.

image

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFilters.storyboard
```diff
@@ -0,0 +1,269 @@
...
+                                </cells>
+                            </tableViewSection>
+                        </sections>
+                        <connections>
+                            <outlet property="dataSource" destination="GDg-hx-dhZ" id="2n9-IR-0fL"/>
+                            <outlet property="delegate" destination="GDg-hx-dhZ" id="d6V-74-n5T"/>
+                        </connections>
+                    </tableView>
+                    <navigationItem key="navigationItem" title="CQL Filters" id="iQj-FT-f4r">
+                        <barButtonItem key="leftBarButtonItem" title="Cancel" id="AK8-Ni-KMZ">
```
philium, line 205: This should be a system Cancel button.

image

philium (commented):
> I didn't draw the text field in a custom cell, but added it as the accessory view to a `Basic` cell to avoid the complexity of setting up constraints.

I don't know that I've ever done that with a text field before, but that seems to work. Nice!

> I couldn't find an elegant way to dismiss the `numberPad` keyboard in a table view

Dismissing on drag probably makes sense, but it doesn't solve the problem of the return key (on a hardware keyboard) not dismissing the (software) keyboard. That problem can be solved by setting a delegate of the text field and implementing the appropriate delegate method. As for dismissing the software keyboard, that issue can be addressed by providing a Done button. That can be accomplished by creating a toolbar (in Interface Builder, just like the switch or the text field) with a done button (usually on the right) and then assigning that as the `inputAccessoryView` of the text field. That will make the toolbar appear above the software keyboard. Wire up the Done button to resign the text field's first responder status.

> The text will be silently reset to default value if the input is invalid.

This behavior was a bit unexpected when using the sample. As I suggested, you might consider either allowing empty as a means of specifying no max or revert to the previously entered max value.

yo1995:
Thanks for the suggestions! I've made changes to improve the UI.

> ... by providing a Done button. That can be accomplished by creating a toolbar (in Interface Builder ...

At last I decided to add the `Done` toolbar in the code to make things easier. I can still change to IB if it is preferred.

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersSettingsViewController.swift
```diff
@@ -0,0 +1,168 @@
...
+    /// The text field for max features input.
+    @IBOutlet var maxFeaturesTextField: UITextField! {
+        didSet {
+            // Add a toolbar to dismiss keyboard.
+            let toolbar = UIToolbar()
+            let items: [UIBarButtonItem] = [
+                .init(barButtonSystemItem: .flexibleSpace, target: nil, action: nil),
+                .init(barButtonSystemItem: .done, target: self, action: #selector(inputAccessoryViewDoneButtonTapped(_:)))
+            ]
+            toolbar.setItems(items, animated: true)
```
philium, line 40: No need for this to be animated as it isn't even on screen yet.

```suggestion
            toolbar.setItems(items, animated: false)
```

arcgis-ios-sdk-samples/Layers/Query with CQL filters/QueryWithCQLFiltersViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+            guard let self = self else { return }
+            self.lastQuery = nil
+            if let error = error,
+               // Do not display error if user cancelled the request.
+               (error as NSError).code != NSUserCancelledError {
+                self.presentAlert(error: error)
+            } else if let result = result, let extent = self.ogcFeatureLayer.featureTable?.extent {
+                // Zoom to the extent of the selected collection.
+                self.mapView.setViewpointGeometry(extent, padding: 50)
+                self.presentAlert(title: "Query Result", message: "Query returned \(result.featureEnumerator().allObjects.count) features.")
```
philium, line 81: Consider using number formatter so that count is formatted properly, such as using a group separator when greater than `999`.

```suggestion
                self.presentAlert(title: "Query Result", message: "Query returned \(NumberFormatter.localizedString(from: result.featureEnumerator().allObjects.count as NSNumber, number: .decimal)) features.")
```

philium (approved):

yo1995:
> yo1995 requested review from philium and saratchandrakarumuri now

GitHub glitched 🤬 ! I'm not intended to re-request Phil's approval 🤦

saratchandrakarumuri (approved):
sample looks good

PR #1080, opened 2021-06-21
Update: Replace screenshot with GIF in `Animate images with image overlay`

yo1995:
Please refer to `common-samples/issues/2787` for details.

James from Qt team mentioned that it might be more visually appealing to use a GIF instead of static image for this sample. Java [is already using GIF](https://github.com/Esri/arcgis-runtime-samples-java/tree/main/scene/animate-images-with-image-overlay) and Qt, .NET and iOS are planning to use GIF. Android opts out because they use the screenshot in the app and GIF isn't supported.

The URL to the branch: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/Update-GIF-vNext/arcgis-ios-sdk-samples/Scenes/Animate%20images%20with%20image%20overlay

This is how GIF looks like on the developer website: https://developers.arcgis.com/java/sample-code/animate-images-with-image-overlay/


philium:
Didn't we have gifs before and removed them? I seem to remember something about the developer site not supporting them. Am I remembering that incorrectly? Or has that changed?

yo1995:
> Didn't we have gifs before and removed them?

This is the only other commit I can find that has "gif" in its message https://github.com/Esri/arcgis-runtime-samples-ios/commit/f43a01cd999d785f3e0e0153fcd377fac0494502 using `git log --oneline | grep gif`, so I'm not sure about that.

Since I started to maintain this repo, all the screenshots are unified to use PNG. Yesterday I asked Mike and Trevor if it's appropriate to follow the suggestion in `common-samples/issues/2787` to introduce GIF, and they said yes. So I made the change in #1081 to allow GIF.

> ... the developer site not supporting them. ... has that changed?

Please see in the description atop. Java [used GIF](https://developers.arcgis.com/java/sample-code/animate-images-with-image-overlay/) and it seems to render fine on the new developer website. I think we only use the screenshot in the sample README and on the developer site, so it should be fine?

Android doesn't support GIF because their sample viewer pull in the screenshot as the collection cell background image and GIF cannot be displayed.

philium:
The gif looks great, but it is rather large (3.44 MB), especially compared to the previous screenshot. Gifs will understandably be larger, but I'm wondering if there's a way to reduce the size at all. Maybe it could it be shortened?

yo1995:
> The gif looks great, but it is rather large (3.44 MB), especially compared to the previous screenshot. Gifs will understandably be larger, but I'm wondering if there's a way to reduce the size at all. Maybe it could it be shortened?

Good point. It is pretty large because it is 250 frames @ 50fps with a 64-color palette. I'll try with ezgif.com and gifsicle to shrink its size.

philium (approved):

vquach2404 (approved):

yo1995:
It's now working correctly on the 100.11.2 build of the doc website: https://developers.arcgis.com/ios/swift/sample-code/animate-images-with-image-overlay/

PR #1082, opened 2021-06-21
Update tile packages

vquach2404:
This PR resolves `common-samples/pull/2779`. Note that iOS does not contain samples that use the raster hillshade data. 

Also, I found a bug in Generate offline map (basemap by reference). I'll open an issue and we can resolve that in a separate PR.


arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -1793,6 +1790,8 @@
 		F140E1EC23F35EA6008AC44E /* DisplayAnnotation.storyboard */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; path = DisplayAnnotation.storyboard; sourceTree = "<group>"; };
 		F142F769267C12E400BD608A /* SanFrancisco.mmpk */ = {isa = PBXFileReference; lastKnownFileType = file; path = SanFrancisco.mmpk; sourceTree = "<group>"; };
 		F142F76D267C12F800BD608A /* Yellowstone.mmpk */ = {isa = PBXFileReference; lastKnownFileType = file; path = Yellowstone.mmpk; sourceTree = "<group>"; };
+		F142F79D267D755700BD608A /* SanFrancisco.tpkx */ = {isa = PBXFileReference; lastKnownFileType = file; name = SanFrancisco.tpkx; path = ../../../Downloads/SanFrancisco.tpkx; sourceTree = "<group>"; };
+		F142F79E267D755700BD608A /* streetmap_SD.tpkx */ = {isa = PBXFileReference; lastKnownFileType = file; name = streetmap_SD.tpkx; path = ../../../Downloads/streetmap_SD.tpkx; sourceTree = "<group>"; };
```
yo1995, line 1794: image


The file paths here aren't correct. They should be relative paths in the `Portal Data` folder. Maybe you've dragged these files into the file navigator from `Downloads` folder? 😏

yo1995 (changes requested):
The changes look OK, but there is an error that stops me from testing.

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
vquach2404: Hmm I see, please try again!

arcgis-ios-sdk-samples/Route and directions/Offline routing/README.md
```diff
@@ -32,7 +32,7 @@ Tap near a road to start adding a stop to the route, tap again to place it on th
 
 ## Offline data
 
-The data used by this sample is available on [ArcGIS Online](https://arcgisruntime.maps.arcgis.com/home/item.html?id=567e14f3420d40c5a206e5c0284cf8fc).
+The data used by this sample is available on [ArcGIS Online](https://arcgis.com/home/item.html?id=567e14f3420d40c5a206e5c0284cf8fc).
```
yo1995, line 35: It is interesting to see that both `Offline routing` and `Offline geocode` use the same `streetmap_SD` tile cache/tpkx, but their actual AGOL portal items are different. I'm a bit confused here but assume it is expected. Can you explain a bit here? Thanks.

yo1995 (approved):

arcgis-ios-sdk-samples/Route and directions/Offline routing/README.md
vquach2404: This portal item is a zip folder containing the `streetmap_SD.tpkx` and 2 other resources that are specific to Offline routing. The portal item in Offline geocode is only the TPKX since the sample doesn't require the other data.

yo1995: I see. I'll add some note here for the Offline routing sample

- The item https://arcgis.com/home/item.html?id=567e14f3420d40c5a206e5c0284cf8fc isn't on-demand downloaded by the app.
- `sandiego.geodatabase` and `sandiego.tn`, which by description are included in this item, are actually included in the app bundle
- Therefore, this sample depends on both the tpkx and the 2 files above. Only the tpkx is downloaded from AGOL while the other 2 files in the bundle are from Apple's server via on-demand resource tag.

yo1995:
OK sorry I've a few more comments 😅 

1. Can you help to verify if Edit data -> Edit features connected depend on the San Diego tile package? It is in the ContentPlist dependencies but I don't think the sample actually uses this resource

image

2. There are some READMEs missing Offline data section or don't have a link. For example

- Features -> Generate geodatabase: it uses the SF tpkx but doesn't say about it
- Edit data -> Edit and sync features: it mentions the data but doesn't have a link
- Edit data -> Offline edit and sync: same as above

Whereas the ArcGIS tiled layer (tile cache) sample does include the SF tpkx link in its README.

image

Are they intended to demonstrate different patterns of loading tile cache? If not, we can surely unify them to `AGSTileCache(name: "SanFrancisco")` (as what you have now)

It's OK not to fix these minor issues, because I know it is a headache to try to keep everything in sync - design, iOS specific README, AGOL portal ID, etc. 😥

vquach2404:
Some of the READMEs didn't get updated in the common design repo so I followed that lead to avoid confusion but I've updated the READMEs now because I think it is more beneficial than confusing.

yo1995 (dismissed):

philium (dismissed):

saratchandrakarumuri (changes requested):
Tiled layer (tile cache) sample code is not modified to look for `tpkx` files, so packages are not populated in the list

Other than that rest of the sample changes looks good

vquach2404:
@saratchandrakarumuri I've updated the sample – please review again, thanks!

saratchandrakarumuri (approved):
Changes looks good

philium (approved):

yo1995 (approved):

PR #1055, opened 2021-04-08
Update: Replace route API and remove OAuth from Navigate in AR

yo1995:
This PR resolves the issue in

- common-samples/issues/2678
- common-samples/issues/2677

that uses API key instead of OAuth to authenticate the user for the navigation task.

PR #1085, opened 2021-06-28
Update: Replace/Remove OAuth with API key

yo1995:
Please refer to the Epic issue `common-samples/issues/2823`.

Per Trevor

> Several samples use a named user authentication pattern for simplicity, but it is now simpler to use an API key - especially with the API being set ubiquitously across most samples with the sample viewer.

Samples we should update:

- [x] Generate offline map
- [x] Generate offline map overrides
- [ ] Generate offline map with local basemap - No need to update, not using OAuth currently
- [x] Navigate in AR - Already been done in #1055

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/GenerateOfflineMapOverridesViewController.swift
+++ b/arcgis-ios-sdk-samples/Maps/Generate offline map (overrides)/GenerateOfflineMapOverridesViewController.swift
@@ -17,7 +17,7 @@
 import UIKit
 import ArcGIS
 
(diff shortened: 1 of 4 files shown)


vquach2404 (approved):
Changes look good! This may also be a good time to update the commenting style.

Edit: Forgot to mention, please update the READMEs since they mention signing into AGOL.

philium (approved):

yo1995:
Hi Sarat, please help to double check the samples are behaving the same as before except without signing in. Thanks!

saratchandrakarumuri (approved):
Updated samples looks good

PR #1086, opened 2021-06-29
Update Create and save map to use named user 

vquach2404:
This PR resolves `common-samples/issues/2769`. The user will be prompted with an AGOL login challenge at the start of the sample and a UIPickerView was added to allow the user to choose a folder to save their map to.


arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
```diff
@@ -52,8 +54,31 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
...
         AGSAuthenticationManager.shared().credentialCache.removeAllCredentials()
         
-        // Initially show the map creation UI.
-        performSegue(withIdentifier: "CreateNewSegue", sender: self)
+        // Temporarily unset the API key for this sample.
+        // Please see the additional information in the README.
+        let apiKey = AGSArcGISRuntimeEnvironment.apiKey
+        AGSArcGISRuntimeEnvironment.apiKey = ""
+        
+        let portal = AGSPortal(url: URL(string: "https://www.arcgis.com")!, loginRequired: true)
```
yo1995, line 62:
```suggestion
        let portal = AGSPortal.arcGISOnline(withLoginRequired: true)
```
Not required, but would be better looking.

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
```diff
@@ -52,8 +54,31 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
...
+        self.portal = portal
+        portal.load { [weak self] (error) in
+            guard let self = self else { return }
+            if let error = error {
+                print(error)
```
yo1995, line 67: I'm a bit unsure here. We shouldn't print anything to console. What about presentAlert or just silently fail here?

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
```diff
@@ -52,8 +54,31 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
         AGSAuthenticationManager.shared().oAuthConfigurations.add(config)
         AGSAuthenticationManager.shared().credentialCache.removeAllCredentials()
         
-        // Initially show the map creation UI.
-        performSegue(withIdentifier: "CreateNewSegue", sender: self)
+        // Temporarily unset the API key for this sample.
+        // Please see the additional information in the README.
+        let apiKey = AGSArcGISRuntimeEnvironment.apiKey
+        AGSArcGISRuntimeEnvironment.apiKey = ""
```
yo1995, line 60: It might be better to put these in the initializer, i.e. `required init?(coder aDecoder: NSCoder) {  // ... }` rather than in the `viewDidLoad`. See Phil's previous comment here: https://github.com/Esri/arcgis-runtime-samples-ios/pull/1067#discussion_r633618703.

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
```diff
@@ -52,8 +54,31 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
...
+                // Get the user's array of portal folders.
+                portal.user?.fetchContent { _, folders, _ in
+                    if let portalFolders = folders {
+                        self.portalFolders = portalFolders
+                    }
+                }
+                // Initially show the map creation UI.
+                self.performSegue(withIdentifier: "CreateNewSegue", sender: self)
+                // Reset the API key after successful login.
+                AGSArcGISRuntimeEnvironment.apiKey = apiKey
```
yo1995, line 78: Also this. I'll leave it to Phil for the best practices.

arcgis-ios-sdk-samples/Maps/Create and save a map/README.md
```diff
@@ -13,13 +13,15 @@ Maps can be created programmatically in code and then serialized and saved as an
...
 
-Select the basemap and layers you'd like to add to your map. Tap the "Save" button. Sign into an ArcGIS Online account. Provide a title, tags, and description. Save the map.
+Enter a username and password for an ArcGIS Online named user account (such as your ArcGIS for Developers account). Select the basemap and layers to add to the map. Tap the "Save" button. Provide a title, tags, description, and folder. Save the map.
 
 ## How it works
 
-1. An `AGSMap` is created with an `AGSBasemap` and a few operational layers.
-2. An `AGSPortal` object is created and loaded. This will issue an authentication challenge, prompting the user to provide credentials.
-3. Once the user is authenticated, call `AGSMap.save(as:portal:tags:folder:itemDescription:thumbnail:forceSaveToSupportedVersion:completion:)` and a new map is saved with the specified title, tags, and folder.
+1. Add an `AGSOAuthConfiguration` to the `AGSAuthenticationManager`'s  `oAuthConfigurations` array and remove all credentials from the `credentialCache`.
```
yo1995, line 20: I'd suggest to follow the OAuth sample here to clear cache on deinit

https://github.com/Esri/arcgis-runtime-samples-ios/blob/61c82de7f8d620cf7edb54e509f98c2f0606a713/arcgis-ios-sdk-samples/Cloud%20and%20portal/Authenticate%20with%20OAuth/AuthenticateWithOAuthViewController.swift#L61-L64

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
```diff
@@ -34,8 +34,10 @@ private extension UIImage {
 
 class CreateSaveMapViewController: UIViewController, CreateOptionsViewControllerDelegate, SaveAsViewControllerDelegate {
     @IBOutlet private weak var mapView: AGSMapView!
+    @IBOutlet weak var saveButton: UIBarButtonItem!
```
yo1995, line 37:
```suggestion
    @IBOutlet private weak var saveButton: UIBarButtonItem!
```

I saw you unified all storyboard views to private weak in the other changes, so let's keep them consistent. Ideally they would be `@IBOutlet var`

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -13,16 +13,27 @@
...
 class SaveAsViewController: UITableViewController {
     @IBOutlet private weak var titleTextField: UITextField!
     @IBOutlet private weak var tagsTextField: UITextField!
     @IBOutlet private weak var descriptionTextField: UITextField!
+    @IBOutlet private weak var folderLabel: UILabel!
     
+    /// Indicates whether the reference scale picker is currently hidden.
+    var folderPickerHidden = true
+    /// The index path of the folder picker view.
+    let folderPicker = IndexPath(row: 4, section: 0)
```
yo1995, line 31: The variable naming here is a bit confusing. Maybe use `folderPickerIndexPath`?

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -38,12 +49,80 @@ class SaveAsViewController: UITableViewController {
...
-        
+        // Get the item description from the text field.
         let itemDescription = descriptionTextField.text ?? ""
-        delegate?.saveAsViewController(self, didInitiateSaveWithTitle: title, tags: tags, itemDescription: itemDescription)
+        
+        delegate?.saveAsViewController(self, didInitiateSaveWithTitle: title, tags: tags, itemDescription: itemDescription, folder: selectedFolder)
+    }
+    
+    /// Toggles visisbility of the reference scale picker.
+    func toggleFolderPickerVisibility() {
```
yo1995, line 63: FYI: OSA Maps app uses an actionSheet for the folder picker

https://github.com/Esri/maps-app-ios/blob/d7ff3e3562351bf595f469c4921f5500a4abf593/maps-app-ios/UI/Account%20View/AccountDetailsViewController.swift#L87

But personally I like your picker here better because it makes the UI more consistent, so let's keep it.

Also I love the attention to the details below using `.accentColor` 👍

yo1995 (commented):
Generally looking good. 👍  I like the new workflow.

I'll add more comments after these changes being addressed.

yo1995:
A few gotchas from today's review

- When the user cancel the login or the login fails, the `New` button should be disabled
- The map picker (options vc) does not have a cancel button. Either add a cancel button, or disable interactive swipe (swipe down gesture) to disable dismiss.

yo1995 (approved):
The sample works fine.

The only worry I have is there are some magic numbers (0, 3, 4, +1, -1) for calculating the folder picker's index path on show/hide. It might be an obstacle if we'll make future updates to the table view.

I'll leave them to Phil. :eyes:

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMapViewController.swift
```diff
@@ -46,14 +63,34 @@ class CreateSaveMapViewController: UIViewController, CreateOptionsViewController
...
+                self.newMapButton.isEnabled = true
+                self.saveButton.isEnabled = true
+            }
+        }
+    }
+    
+    deinit {
+        // Reset the API key after successful login.
+        AGSArcGISRuntimeEnvironment.apiKey = apiKey
+        if let config = oAuthConfiguration {
```
yo1995, line 90: I believe there would be a build [error](https://github.com/Esri/arcgis-runtime-samples-ios/pull/1086#discussion_r661878252) shown on this line.

yo1995 (dismissed):

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
philium: As cool as inline pickers are, I'm not sure that it is the best choice for this UI. I say that because of the following from the [HIG](https://developer.apple.com/design/human-interface-guidelines/ios/controls/pickers/):

image

It might make more sense to display the folders in a separate table view controller that is pushed onto the navigation stack when the Folder row is tapped.

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -38,12 +49,81 @@ class SaveAsViewController: UITableViewController {
...
+        tableView.performBatchUpdates({
+            if folderPickerHidden {
+                folderLabel.textColor = .accentColor
+                tableView.insertRows(at: [folderPickerIndexPath], with: .fade)
+                folderPickerHidden = false
```
philium, line 68: FYI: `folderPickerHidden.toggle()` could have been used after the `if` statement instead.

philium (commented):
I'm not convinced that `create-and-save-map-1.png` is pulling its own weight. The UI it shows isn't really specific to the sample. Perhaps it should be removed? Just a thought.

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -13,16 +13,23 @@
...
 class SaveAsViewController: UITableViewController {
     @IBOutlet private weak var titleTextField: UITextField!
     @IBOutlet private weak var tagsTextField: UITextField!
     @IBOutlet private weak var descriptionTextField: UITextField!
+    @IBOutlet private weak var folderLabel: UILabel!
     
+    /// The array of folders loaded from the portal.
+    var portalFolders = [AGSPortalFolder]()
+    /// Folder selected by the user.
+    var selectedFolder: AGSPortalFolder?
```
philium, line 31: Instead of storing the selected folder, store the index of the selected folder. Then make this a computed property:

```swift
var selectedFolderIndex: Int?
var selectedFolder: AGSPortalFolder? { selectedFolderIndex.map { portalFolders[$0] } }
```

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -38,12 +45,51 @@ class SaveAsViewController: UITableViewController {
...
+    func showFolderOptions() {
+        // Set index for the default option.
+        let selectedIndex = portalFolders.firstIndex { $0 == selectedFolder } ?? portalFolders.count
+        // The titles of the portal folders and a "No folder" option.
+        let folderTitles = portalFolders.map { $0.title! } + ["No folder"]
+        // Prepare the options table view controller and handle selection.
+        let optionsViewController = OptionsTableViewController(labels: folderTitles, selectedIndex: selectedIndex) { [weak self] newIndex in
+            guard let self = self else { return }
+            // If "No folder" was selected, set selectedFolder to nil.
+            if newIndex == self.portalFolders.count {
```
philium, line 68: It would be great if the user could just tap on a selected row to unselect it. I see that `OptionsTableViewController` doesn't currently support that, but it would be easy enough to add. So as not to change the behavior of existing samples, you could add an `allowsEmptySelection` property (with a default of `false`) to `OptionsTableViewController` and only allow deselection when that property is set to `true`.

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -38,12 +45,51 @@ class SaveAsViewController: UITableViewController {
...
+                self.selectedFolder = nil
+            } else {
+                // Select the appropriate folder.
+                self.selectedFolder = self.portalFolders[newIndex]
+            }
+            // Replace the label text with the selected option.
+            self.folderLabel.text = self.selectedFolder?.title ?? "No folder"
```
philium, line 75: I would suggest using "None" as the label text.

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -38,12 +45,51 @@ class SaveAsViewController: UITableViewController {
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+    }
+}
+
+extension SaveAsViewController /* UITableViewDelegate */ {
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let folderCell = IndexPath(row: 3, section: 0)
+        if indexPath == folderCell {
+            tableView.deselectRow(at: folderCell, animated: true)
```
philium, line 91: This is not needed (see `UITableViewController.clearsSelectionOnViewWillAppear`) and in fact prevents the neat deselection effect seen when interactively swiping from the left to pop the options view controller.

```suggestion
```

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMap.storyboard
```diff
@@ -225,6 +232,59 @@
...
+                                                        <outlet property="dataSource" destination="Kr2-KR-2IC" id="qZR-E6-jgu"/>
+                                                        <outlet property="delegate" destination="Kr2-KR-2IC" id="PKC-07-Iq6"/>
+                                                    </connections>
+                                                </pickerView>
+                                            </subviews>
+                                            <constraints>
+                                                <constraint firstItem="Xt4-kR-W8Z" firstAttribute="top" secondItem="ioJ-ni-tzA" secondAttribute="top" id="Sfx-w0-qV2"/>
+                                                <constraint firstItem="Xt4-kR-W8Z" firstAttribute="leading" secondItem="ioJ-ni-tzA" secondAttribute="leading" id="aaj-uv-75h"/>
+                                                <constraint firstAttribute="bottom" secondItem="Xt4-kR-W8Z" secondAttribute="bottom" id="bvX-QU-37l"/>
+                                                <constraint firstAttribute="trailing" secondItem="Xt4-kR-W8Z" secondAttribute="trailing" id="g17-uY-a7H"/>
```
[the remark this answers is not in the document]
vquach2404: I think this issue is resolved since I've changed the cell style to Right detail. Please let me know if I need to make further changes!

arcgis-ios-sdk-samples/Maps/Create and save a map/CreateSaveMap.storyboard
```diff
@@ -225,6 +232,30 @@
...
+                                                    <autoresizingMask key="autoresizingMask"/>
+                                                    <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                                    <nil key="textColor"/>
+                                                    <nil key="highlightedColor"/>
+                                                </label>
+                                                <label opaque="NO" multipleTouchEnabled="YES" contentMode="left" insetsLayoutMarginsFromSafeArea="NO" text="No folder" textAlignment="right" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" id="6Mm-O7-7a8">
+                                                    <rect key="frame" x="304.5" y="12" width="70.5" height="20.5"/>
+                                                    <autoresizingMask key="autoresizingMask"/>
+                                                    <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                                    <nil key="textColor"/>
```
philium, line 253: The text color should be a dark gray rather than black. See Settings > Wi-Fi/Bluetooth for examples.

```suggestion
                                                    <color key="textColor" systemColor="secondaryLabelColor"/>
```

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
vquach2404: Are you suggesting that I make this change in `OptionsTableViewController.swift` in the Cocoa repo?

yo1995: This https://github.com/Esri/arcgis-runtime-samples-ios/blob/v.next/arcgis-ios-sdk-samples/Content%20Display%20Logic/Controllers/OptionsTableViewController.swift

vquach2404: Update: this issue does not impede the user's workflow thus, will be handled in a separate PR `common-samples/issues/2850` after the release. This decision is was made to prioritize the release and to ensure that we'll release according to schedule. Please continue reviewing the sample in its current state without this feature for now. Thank you!

vquach2404:
> I'm not convinced that `create-and-save-map-1.png` is pulling its own weight. The UI it shows isn't really specific to the sample. Perhaps it should be removed? Just a thought.

Hm, although it isn't specific to this sample, it plays an integral role in the workflow. I don't think there's any harm in keeping it. @yo1995 do you have any strong opinions?

yo1995:
> > I'm not convinced that `create-and-save-map-1.png` is pulling its own weight. The UI it shows isn't really specific to the sample. Perhaps it should be removed? Just a thought.
> 
> Hm, although it isn't specific to this sample, it plays an integral role in the workflow. I don't think there's any harm in keeping it. @yo1995 do you have any strong opinions?

I don't have a preference on either keeping or removing the login screenshot. Though the 4th screenshot should be updated once the folder picker UI being finalized.

philium (approved):

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -38,12 +47,46 @@ class SaveAsViewController: UITableViewController {
...
+        
+        delegate?.saveAsViewController(self, didInitiateSaveWithTitle: title, tags: tags, itemDescription: itemDescription, folder: selectedFolder)
+    }
+    
+    /// Present the list of folders to choose from.
+    func showFolderOptions() {
+        // Set index for the default option.
+        let selectedIndex = selectedFolderIndex ?? portalFolders.count
+        // The titles of the portal folders and a "None" option.
+        let folderTitles = portalFolders.map { $0.title! } + ["None"]
```
yo1995, line 65:
```suggestion
        let folderTitles = portalFolders.map(\.title!) + ["None"]
```

Don't have to make this change. Just want to double check if we prefer the key path style.

yo1995 (approved):
LGTM!

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
philium: Does that work? If so, I didn't know keypaths to nullable properties could be force unwrapped like that.

yo1995: Seems to compile fine on my end. But I'm quite hesitant to write it like that

- I too have never seen key path unwrapping
- Even if the folder title is almost guaranteed not to be `nil`, I still don't like the ❗ . What about `let folderTitles = portalFolders.compactMap(\.title) + ["None"]`?

philium: Not a good idea to compact map because indices will be messed up in the event of `nil`. Better to just provide a default value.

```suggestion
        let folderTitles = portalFolders.map { $0.title ?? "" } + ["None"]
```

yo1995 (approved):

philium (approved):

saratchandrakarumuri (approved):
Changes looks good

vquach2404:
Thank you all for the reviews!

PR #1088, opened 2021-07-14
Update Options table view controller

vquach2404:
This PR addresses the [suggestion](https://github.com/Esri/arcgis-runtime-samples-ios/pull/1086#discussion_r665687127) that the Options table view controller should allow deselection. Create and save map is the only sample that allows deselection for now. The other impacted samples have been updated accordingly but their behaviors remain the same.


arcgis-ios-sdk-samples/Utility network/Configure subnetwork trace/ConfigureSubnetworkTraceOptionsViewController.swift
```diff
@@ -142,7 +142,7 @@ class ConfigureSubnetworkTraceOptionsViewController: UITableViewController {
     func showAttributePicker() {
         let selectedIndex = possibleAttributes.firstIndex { $0 == selectedAttribute }
         let optionsViewController = OptionsTableViewController(labels: possibleAttributes.map { $0.name }, selectedIndex: selectedIndex) { newIndex in
-            self.selectedAttribute = self.possibleAttributes[newIndex]
+            self.selectedAttribute = self.possibleAttributes[newIndex!!]
```
yo1995, line 145: There is an extra trailing bang ❗  that gives a build error.

vquach2404: Oops, forgot to push my last commit. It's fixed now!

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
```diff
@@ -95,6 +96,11 @@ class OptionsTableViewController: UITableViewController {
             }
             tableView.reloadRows(at: indexPathsToReload, with: .automatic)
             onChange(indexPath.row)
+        } else if allowsEmptySelection {
+            tableView.deselectRow(at: indexPath, animated: true)
```
yo1995, line 100:
```suggestion
```
Since the row is reloaded after the `selectedIndex` being nullified, no need to deselect the row. However, it might be good to keep it in terms of parity with the other conditional branches.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
```diff
@@ -31,15 +31,16 @@ class OptionsTableViewController: UITableViewController {
...
+    var allowsEmptySelection = false
     
     /// Creates a new instance with the options, selected index, and selection
     /// change handler.
     /// - Parameters:
     ///   - options: The options displayed by the view controller.
     ///   - selectedIndex: The index of the currently selected option or `nil`.
     ///   - onChange: A closure called when the selected option has changed.
-    init(options: [Option], selectedIndex: Int?, onChange: @escaping (Int) -> Void) {
+    init(options: [Option], selectedIndex: Int?, onChange: @escaping (Int?) -> Void) {
```
yo1995, line 43: You and Phil might have discussed here and came up with the instance property solution. I think that is fine, but I prefer to put it in the initializer as a parameter with default value like this

```swift
init(options: [Option], selectedIndex: Int?, allowsEmptySelection: Bool = false, onChange: @escaping (Int?) -> Void) { // ... }
```

so that in `Create and save the map` we don't need to set it on another line.

Note if you gonna apply this change, be sure to also update the comment docstrings for both `init` methods, and make the `allowsEmptySelection` property private.

yo1995 (commented):
LGTM, just a few small suggestions.

I pondered and am convinced that the `allowsEmptySelection` property is needed. 

- Initially I though that we can just nullify the `selectedIndex` and reload the row (without calling back).
- But the UI doesn't make sense in this case, when the `OptionsTableViewController` is presented as a popover, where uncheck a row does nothing.
- Therefore we have to have a boolean to indicate if the `selectedIndex` can be `nil`, and use it in those samples that require a 3-state selection.

So I'm good with these changes.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
```diff
@@ -31,30 +31,34 @@ class OptionsTableViewController: UITableViewController {
     /// The index of the currently selected option or `nil` if no option is
     /// selected.
     private var selectedIndex: Int?
-    private let onChange: (Int) -> Void
+    private let onChange: (Int?) -> Void
+    private var allowsEmptySelection: Bool
```
yo1995, line 35:
```suggestion
    private let allowsEmptySelection: Bool
```
as we'll only assign it in the initializers.

yo1995 (approved):

philium (commented):
Every consumer of `OptionsTableViewController` needing to deal with an optional index is most unfortunate. Thankfully, I don't think it is necessary. Instead of making all indexes optional and using a parameter to determine whether empty selection is allows, a second convenience initializer can be added that makes the integer optional:

```swift
convenience init(labels: [String], selectedIndex: Int, onChange: @escaping (Int) -> Void) {
    let options = labels.map { Option(label: $0) }
    self.init(options: options, selectedIndex: selectedIndex, allowsEmptySelection: false, onChange: { onChange($0!) })
}

convenience init(labels: [String], selectedIndex: Int?, onChange: @escaping (Int?) -> Void) {
    let options = labels.map { Option(label: $0) }
    self.init(options: options, selectedIndex: selectedIndex, allowsEmptySelection: true, onChange: onChange)
}
```

Which initializer should be used will be inferred by the compiler based on the `selectedIndex` argument. If it non-optional, the first initializer will be used. If it is optional, the second one will be used.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
```diff
@@ -31,30 +31,45 @@ class OptionsTableViewController: UITableViewController {
...
-        self.init(options: options, selectedIndex: selectedIndex, onChange: onChange)
+        self.init(options: options, selectedIndex: selectedIndex, allowsEmptySelection: false, onChange: onChange)
+    }
+    
+    /// Creates a new instance with the given labels, selected index, empty selection bool, and
+    /// selection change handler.
+    /// - Parameters:
+    ///   - labels: An array of labels for the options.
+    ///   - selectedIndex: The index of the currently selected option or `nil`.
+    ///   - allowsEmptySelection: A bool to determine if empty selection is allowed. Defaulted to false.
```
yo1995, line 68: The docstring here needs adjustment.

vquach2404: Thanks for the reminder! This PR is still under construction and not yet ready for a review

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
```diff
@@ -31,30 +31,66 @@ class OptionsTableViewController: UITableViewController {
...
+        self.init(options: options, selectedIndex: selectedIndex, allowsEmptySelection: false, onChange: { onChange($0!) })
+    }
+    
+    /// Creates a new instance with the given labels, selected index, empty selection bool, and
     /// selection change handler.
     /// - Parameters:
     ///   - labels: An array of labels for the options.
     ///   - selectedIndex: The index of the currently selected option or `nil`.
     ///   - onChange: A closure called when the selected option has changed.
     convenience init(labels: [String], selectedIndex: Int?, onChange: @escaping (Int) -> Void) {
```
vquach2404, line 79: There were a few existing samples that had an optional `selectedIndex` and non optional `onChange` so I added this convenience as well.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
```diff
@@ -31,30 +31,66 @@ class OptionsTableViewController: UITableViewController {
...
+    }
+    
+    /// Creates a new instance with the given labels, selected index, empty selection bool, and
+    /// selection change handler.
+    /// - Parameters:
+    ///   - labels: An array of labels for the options.
+    ///   - selectedIndex: The index of the currently selected option or `nil`.
+    ///   - allowsEmptySelection: A bool that determines whether or not empty selection is allowed.
+    ///   - onChange: A closure called when the selected option has changed.
+    convenience init(labels: [String], selectedIndex: Int?, allowsEmptySelection: Bool, onChange: @escaping (Int?) -> Void) {
```
vquach2404, line 91: I decided to include the `allowsEmptySelection` parameter here to easily differentiate this init from the others. 

There is one other case (non optional `selectedIndex` with optional `onChange`) that I didn't include because I don't think it would be useful but if anyone disagrees, I can add that as well.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
```diff
@@ -31,30 +31,66 @@ class OptionsTableViewController: UITableViewController {
...
+    /// Creates a new instance with the given labels, selected index, empty selection bool, and
     /// selection change handler.
     /// - Parameters:
     ///   - labels: An array of labels for the options.
     ///   - selectedIndex: The index of the currently selected option or `nil`.
     ///   - onChange: A closure called when the selected option has changed.
     convenience init(labels: [String], selectedIndex: Int?, onChange: @escaping (Int) -> Void) {
         let options = labels.map { Option(label: $0) }
-        self.init(options: options, selectedIndex: selectedIndex, onChange: onChange)
+        self.init(options: options, selectedIndex: selectedIndex, allowsEmptySelection: false, onChange: { onChange($0!) })
```
yo1995, line 81:
```suggestion
        self.init(options: options, selectedIndex: selectedIndex, onChange: { onChange($0!) })
```

If we set a default value for `allowsEmptySelection` in the base constructor, then we can omit it in the convenience ones. However it might be good to keep it here so the `convenience init`s echo each other.

yo1995 (approved):
I think the new solution works fine. 👍 
The only worry would be that if we gonna add more parameters to the initializer in the future, the number of `convenience init` methods will explode due to combinatory math.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
vquach2404: Hm, yeah I'm afraid it might cause confusion if we omit `allowsEmptySelection` in some but leave it in others. I'll leave this open for discussion.

vquach2404:
> The only worry would be that if we gonna add more parameters to the initializer in the future, the number of `convenience init` methods will explode due to combinatory math.

Can you explain further? Is there a limit of `convenience init` that we shouldn't exceed?

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
philium: This initializer isn't needed. Instead, the code in `SaveAsViewController.swift` should be updated to account for these changes:

image

philium (changes requested):

arcgis-ios-sdk-samples/Content Display Logic/Controllers/OptionsTableViewController.swift
vquach2404: @philium By removing this initializer, how will the `onChange` return an optional value? And wouldn't that defeat the purpose of adding the `allowsEmptySelection` property? The other initializers set `allowsEmptySelection` to false and `onChange` cannot return nil

philium: Oops! Not this one. I meant the one before it, where the optionality of the `selectedIndex` parameter and the index taken by the `onChange` closure do not match:

```swift
convenience init(labels: [String], selectedIndex: Int?, onChange: @escaping (Int) -> Void)
```

vquach2404: Oh I added this one because of situations like[ this](https://github.com/Esri/arcgis-runtime-samples-ios/blob/ee1bbe9b20a009770c12a3f2fb048edf552115e9/arcgis-ios-sdk-samples/Utility%20network/Configure%20subnetwork%20trace/ConfigureSubnetworkTraceOptionsViewController.swift#L166-L178). There were a handful of similar cases where `selectedIndex` can be nil due to no initial selection but would later need to return a non optional index.

philium: It still think some of these initializers can be consolidated, but I don't have time to look into it more at the moment.

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -64,7 +64,7 @@ class SaveAsViewController: UITableViewController {
         // The titles of the portal folders and a "None" option.
         let folderTitles = portalFolders.map { $0.title ?? "(No Title)" } + ["None"]
```
philium, line 65: Because empty selection is now supported, there is no need for the `None` option anymore. Could you please remove it?

philium (changes requested):

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
```diff
@@ -59,23 +59,14 @@ class SaveAsViewController: UITableViewController {
...
-            if newIndex == self.portalFolders.count {
-                self.selectedFolderIndex = nil
-            } else {
-                // Select the appropriate folder.
-                self.selectedFolderIndex = newIndex
-            }
+            self.selectedFolderIndex = newIndex
             // Replace the label text with the selected option.
             self.folderLabel.text = self.selectedFolder?.title ?? "None"
-            self.navigationController?.popViewController(animated: true)
```
philium, line 78: Did you mean to remove this? That's fine; haven't to tap the back button isn't that big of a deal here. Just wanted to double check.

philium (approved):

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
vquach2404: Yes I meant to remove that. IMO, automatically popping the view controller seems to make deselection a hassle in this case.

yo1995 (approved):

arcgis-ios-sdk-samples/Maps/Create and save a map/SaveAsViewController.swift
philium: If a user wants to select a different folder, they don't have to deselect the current one in order to select the new one, right? So a user could change the selection either by selecting a new row (going from no selection to a selection or changing the existing selection) or by selecting an existing selection (going from a selection to no selection). Either one of those seems like a perfectly fine reason to pop the view controller, since there isn't much else they can do (other than change the selection again, but who would want to do that?). But either way is fine.

saratchandrakarumuri (approved):

PR #1089, opened 2021-07-19
Update: Add true curves APIs to `Add graphics with renderer`

yo1995:
This is the first U12 API into `v.next` branch. Changes

- Add true curve APIs according to `common-samples/issues/2819`
- Update the screenshot with a heart shape
- Update docs to reflect the changes
- Rename the src files to follow the naming convention
- Remove `redirect_from` as it is no longer valid


arcgis-ios-sdk-samples/Display information/Add graphics with renderer/AddGraphicsWithRendererViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+        guard sideLength > 0 else { return nil }
+        let sr = center.spatialReference
+        // The x and y coordinates to simplify the calculation.
+        let minX = center.x - 0.5 * sideLength
+        let minY = center.y - 0.5 * sideLength
+        // The radius of the arcs.
+        let arcRadius = sideLength * 0.25
+        
+        // Bottom left curve.
+        let leftCurveStart = AGSPoint(x: center.x, y: minY, spatialReference: sr)
```
vquach2404, line 131: Since you've updated the other variable names, I recommend `sr` be changed as well.

arcgis-ios-sdk-samples/Display information/Add graphics with renderer/README.md
```diff
@@ -10,7 +10,7 @@ A renderer allows you to change the style of all graphics in an overlay by only
 
 ## How to use the sample
 
-Run the sample and view graphics for points, lines, and polygons, which are stylized using renderers.
+Run the sample and view graphics for points, lines, polygons, and shapes with curved segments which are stylized using renderers.
```
vquach2404, line 13:
```suggestion
Pan and zoom on the map to view graphics for points, lines, polygons, and shapes with curved segments which are stylized using renderers.
```
I think it's been a while since we've updated this sample and README. A slight change would keep this up to date with our current guidelines.

vquach2404 (commented):
Since the graphics are in the same area, it may be good to set the viewpoint to focus on the graphics. Overall looking good!

arcgis-ios-sdk-samples/Display information/Add graphics with renderer/AddGraphicsWithRendererViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+    
+    func makePointGraphicsOverlay() -> AGSGraphicsOverlay {
+        // Create a simple marker symbol.
+        let pointSymbol = AGSSimpleMarkerSymbol(style: .diamond, color: .green, size: 10)
+        // Create a graphics overlay for the points.
+        let pointGraphicsOverlay = AGSGraphicsOverlay()
+        // Create and assign a simple renderer to the graphics overlay.
+        pointGraphicsOverlay.renderer = AGSSimpleRenderer(symbol: pointSymbol)
+        return pointGraphicsOverlay
+    }
```
vquach2404, line 82: I think the `make___GraphicsOverlay` functions could be generalized so you can pass the symbol type and use a switch statement. What do you think?

yo1995: Good point. I mostly copy pasted the code from the original implementation. Due to renaming of the file they don't show as diff but as deletions, so there is no way to compare them on GitHub.

I think this sample is intended to face those more beginner-level developers. So I intentionally didn't make big structural changes to the code, so that they can be copied and pasted easily. E.g. when someone want to test a single symbol, they can just copy the 2 methods (`makeSomeGraphicsOverlay` and `addGraphicsOverlays`) into a new file and see them renders.

I did extract them into `makeSomeGraphicsOverlay` methods though because otherwise the original `addGraphicsOverlay` method would be too long after adding the new curvature graphics. 😅 

I'll leave this discussion open to Phil and Sarat.

yo1995:
Thanks for the review. I guess I'll hold off making changes to this PR until the design update has been merged - otherwise there are too many back and forth to keep them in sync 😅 .

arcgis-ios-sdk-samples/Display information/Add graphics with renderer/README.md
```diff
@@ -10,7 +10,7 @@ A renderer allows you to change the style of all graphics in an overlay by only
 
 ## How to use the sample
 
-Run the sample and view graphics for points, lines, and polygons, which are stylized using renderers.
+Pan and zoom on the map to view graphics for points, lines, polygons (including polygons with curve segments) which are stylized using renderers.
 
 ## How it works
```
vquach2404, line 15: I didn't notice this before since no changes were made to this section, but the list should be enumerated. I also wonder if some explanation should be added to go into depth about the added API. Seems a bit odd to have a long list of Relevant API but disregard half of them in this section.

yo1995: Do you mean to change it to ...?
> Pan and zoom on the map to view graphics for points, lines, **and** polygons (including polygons with curve segments) which are stylized using renderers.

> I also wonder if some explanation should be added to go into depth about the added API. Seems a bit odd to have a long list of Relevant API but disregard half of them in this section.

It would have been a good question to ask in the design update PR. 

IMO though it is OK to omit them from the `How it works` section. This sample is intended for exhibiting how to apply renderers to graphics overlays, not how to construct geometries. True curve APIs are only included in the code so that user can find example snippet when they search for these keywords. @TADraeseke may chime in here.

vquach2404: I think others may have missed this section in the design update as well since this section wasn't highlighted in either PRs. Feel free to request Phil's and Sarat's reviews so they can chime in as well

vquach2404 (dismissed):

arcgis-ios-sdk-samples/Display information/Add graphics with renderer/README.md
TADraeseke: Oh yeah! Should be enumerated according to our [README template](https://devtopia.esri.com/runtime/common-samples/wiki/README-Template#how-it-works). 

As to Relevant API. Really good point @vquach2404. This sample's main purpose is to show use of a renderer and it's been co-opted to show a curve since it's an easy win until we build out more functionality next release.

Let's not worry about changing the steps too much now (though they should be enumerated). This sample will serve as a basic story that we support curves, but we'll go into detail on how in another sample next release

arcgis-ios-sdk-samples/Display information/Add graphics with renderer/AddGraphicsWithRendererViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+        let rightCurveStart = AGSPoint(x: minX + sideLength, y: minY + 0.75 * sideLength, spatialReference: spatialReference)
+        let rightCurveEnd = leftCurveStart
+        let rightControlPoint1 = AGSPoint(x: minX + sideLength, y: center.y, spatialReference: spatialReference)
+        let rightControlPoint2 = leftControlPoint1
+        let rightCurve = AGSCubicBezierSegment(start: rightCurveStart, controlPoint1: rightControlPoint1, controlPoint2: rightControlPoint2, end: rightCurveEnd, spatialReference: spatialReference)!
+        
+        let heart = [leftCurve, leftArc, rightArc, rightCurve].reduce(AGSMutablePart(spatialReference: spatialReference)) { (part: AGSMutablePart, segment: AGSSegment) -> AGSMutablePart in
+            part.add(segment)
+            return part
+        }
```
philium, line 155: `reduce` is really suited more toward value types. For something like this, it seems a bit overkill. You might consider doing something a bit simpler:

```swift
let heart = AGSMutablePart(spatialReference: spatialReference)
[leftCurve, leftArc, rightArc, rightCurve].forEach { heart.add($0) }
```

arcgis-ios-sdk-samples/Display information/Add graphics with renderer/AddGraphicsWithRendererViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+    
+    // MARK: UIViewController
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Add the source code button item to the right of navigation bar.
+        (navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["AddGraphicsWithRendererViewController"]
+        
+        // Add the graphics overlays, with graphics added, to the map view.
+        addGraphicsOverlays()
```
philium, line 169: If it were me, I would do away with `addGraphicsOverlay()` and instead have methods for creating each of the overlays and then call them in the `mapView` property observer:

```swift
mapView.graphicsOverlays.addObjects(from: [
    makePointGraphicsOverlay(),
    makeLineGraphicsOverlay(),
    makePolygonGraphicsOverlay(),
    makeCurveGraphicsOverlay()
])
```

arcgis-ios-sdk-samples/Display information/Add graphics with renderer/AddGraphicsWithRendererViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+        return curvedGraphicsOverlay
+    }
+    
+    /// Create a heart-shape geometry with Bezier and elliptic arc segments.
+    ///
+    /// - Parameters:
+    ///   - center: The center of the square that contains the heart shape.
+    ///   - sideLength: The side length of the square.
+    /// - Returns: A heart-shape geometry.
+    func makeHeartGeometry(center: AGSPoint, sideLength: Double) -> AGSGeometry? {
```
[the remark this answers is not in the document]
yo1995: `AGSPoint`'s initializer takes `Double` for coordinates, thus I used double for the side length, instead of the conventional `CGFloat` (for core graphics classes such as `CGRect` and `CGSize`) to avoid casting back to double.

philium (changes requested):
Please update the description in `ContentPList.plist`.

philium (dismissed):

yo1995:
> Please update the description in `ContentPList.plist`.

Thanks for catching this! Though I wasn't sure why this sample was excluded from the description checker script in the first place 🤦‍♂️  https://github.com/Esri/arcgis-runtime-samples-ios/blob/ee1bbe9b20a009770c12a3f2fb048edf552115e9/Scripts/CI/README_Metadata_StyleCheck/description_differ.py#L33

I'll request Sarat's verification.

philium:
> Though I wasn't sure why this sample was excluded from the description checker script in the first place 🤦‍♂️

Can that exception be removed? And if so, could that be done in this PR?

saratchandrakarumuri (approved):
Looks good

philium (approved):

yo1995:
Thank you all!

PR #973, opened 2020-09-24
New: View hidden infrastructure in AR

yo1995:
This PR implements the new `View hidden infrastructure in AR` sample. Please refer to `/common-samples/issues/1464` for details.

Note: this sample only runs on real device.

|demo|
|-|
|200924-ViewHiddenInfraAR|

PR #1093, opened 2021-07-30
New: `Orbit camera around object`

yo1995:
## Description

This PR implements the `Orbit camera around object` sample. See details at `common-samples/issues/1188`

Branch URL: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/New-Orbit-vNext/arcgis-ios-sdk-samples/Scenes/Orbit%20camera%20around%20object

See a demo video below.

https://user-images.githubusercontent.com/9660181/127719794-b1811315-8d4e-45a3-90f9-721e77a9fa27.mp4

## Known issues

1. ~~If the animation for cockpit view is interrupted before finishing, there will be a timing bug. i.e. the camera may end up in an unexpected position. Since it doesn't matter too much, I didn't handle the cancelable job.~~ cancelable job is handled, and `animation finished` is checked.
2. ~~When the popover is presenting, touching on the scene view doesn't dismiss it. This is as designed, so that user can pan and zoom on the scene view to control the camera heading. To dismiss the popover, either tap on the nav bar, or on the button itself.~~ See discussion below.
3. The heading and pitch of the graphic is used instead of the symbol's. This is as designed, though I'm not sure why. (i.e. `planeGraphic.attributes["PITCH"]` is used instead of `planeSymbol.pitch`)

## How to test

Read through the README, and try all the UI controls to see if the rotation stuck at some point.


arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraAroundObjectViewController.swift
```diff
@@ -0,0 +1,176 @@
...
+        
+        // Animate the camera target to the cockpit.
+        cameraController.setTargetOffsetX(0, targetOffsetY: -2, targetOffsetZ: 1.1, duration: 1)
+        
+        // If the camera is already tracking the plane's pitch, subtract it from
+        // the delta angle for the animation.
+        let pitchDelta = cameraController.isAutoPitchEnabled ? 0 : 90 - cameraController.cameraPitchOffset + ((planeGraphic.attributes["PITCH"] as? NSNumber)?.doubleValue ?? 0)
+        // Animate the camera so that it is at the target (cockpit), facing
+        // forward (0 deg heading), and aligned with the horizon (90 deg pitch).
+        cameraController.moveCamera(
```
yo1995, line 134: Known issue 1

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraAroundObjectViewController.swift
```diff
@@ -0,0 +1,176 @@
...
+    
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        if let controller = segue.destination as? OrbitCameraSettingsViewController {
+            controller.orbitGeoElementCameraController = sceneView.cameraController as? AGSOrbitGeoElementCameraController
+            controller.planeGraphic = planeGraphic
+            controller.presentationController?.delegate = self
+            controller.preferredContentSize = CGSize(width: 300, height: 210)
+            // Pass through the scene view to allow pan and zoom while the
+            // popover is present.
+            controller.popoverPresentationController?.passthroughViews = [sceneView]
```
yo1995, line 157: Known issue 2

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraSettingsViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+    func updateUIForHeading() {
+        guard let cameraController = orbitGeoElementCameraController else { return }
+        headingSlider.value = Float(cameraController.cameraHeadingOffset)
+        let measurement = Measurement(value: cameraController.cameraHeadingOffset, unit: UnitAngle.degrees)
+        headingLabel.text = measurementFormatter.string(from: measurement)
+    }
+    
+    func updateUIForPitch() {
+        guard let graphic = planeGraphic,
+              let pitch = graphic.attributes["PITCH"] as? NSNumber else { return }
```
yo1995, line 54: Known issue 3

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/README.md
```diff
@@ -0,0 +1,46 @@
...
+* disable user interactions for rotating the camera
+* animate camera movement over a specified duration
+* control the vertical positioning of the target on the screen
+* set a target offset (e.g.to orbit around the tail of the plane) instead of defaulting to orbiting the center of the object
+
+## How to use the sample
+
+The sample loads with the camera orbiting an airplane model. The camera is preset with a restricted camera heading and pitch, and a limited minimum and maximum camera distance set from the plane. The position of the plane on the screen is also set just below center.
+
+Use the sliders to adjust the camera heading and the plane's pitch. When not in Cockpit view, the plane's pitch will change independently to that of the camera pitch. Toggle on the switch to allow zooming in and out with the mouse/keyboard; when the switch is off, the user won't be able to adjust with the camera distance.
```
yo1995, line 24:
```suggestion
Use the sliders to adjust the camera heading and the plane's pitch. When not in Cockpit view, the plane's pitch will change independently to that of the camera pitch. Toggle on the switch to allow zooming in and out with the mouse/keyboard; when the switch is off, the user won't be able to adjust the camera distance.
```
Should it be without "with"?

vquach2404: I agree, remove the "with". Additionally, adjust the phrasing to suit mobile gestures.

```suggestion
Use the sliders to adjust the camera heading and the plane's pitch. When not in Cockpit view, the plane's pitch will change independently to that of the camera pitch. Toggle on the switch to allow zooming in and out by pinching; when the switch is off, the user won't be able to adjust with the camera distance.
```

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Tap the "Cockpit view" button to offset and fix the camera into the cockpit of the airplane. Tap the "Center view" button to exit cockpit view mode and fix the camera controller on the center of the plane.
+
+## How it works
+
+1. Instantiate an `AGSOrbitGeoElementCameraController` object.
+2. Set the camera controller to the scene view.
+3. Set the heading, pitch and distance properties for the camera controller.
```
vquach2404, line 32:
```suggestion
3. Set the heading, pitch, and distance properties for the camera controller.
```

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/README.md
```diff
@@ -0,0 +1,46 @@
...
+4. Set the minimum and maximum angle of heading and pitch, and minimum and maximum distance for the camera.
+5. Set the distance from which the camera is offset from the plane.
+6. Set the `targetVerticalScreenFactor` property to determine where the plane appears in the scene.
+7. Animate the camera to the cockpit using `AGSOrbitGeoElementCameraController.setTargetOffsetX(_:targetOffsetY:targetOffsetZ:duration:completion:)`.
+8. Set `isCameraDistanceInteractive` if the camera distance will adjust when zooming or panning using mouse or keyboard (default is true).
```
vquach2404, line 37:
```suggestion
8. Set `isCameraDistanceInteractive` if the camera distance will adjust when zooming or panning by pinching (default is true).
```

vquach2404 (commented):
Looking good! I think the toolbar could use some improvements. 

I'm still able to select the other view options despite the camera settings popover being present and the other buttons being disabled. Additionally, I can't exit the popover unless I tap the "camera" button again. Tapping the scene should dismiss the popover.

https://user-images.githubusercontent.com/52215269/128094626-9ad676df-c33b-417f-8114-4ecf8b777f22.mov

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraAroundObject.storyboard
```diff
@@ -0,0 +1,225 @@
...
+                        <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <subviews>
+                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="8eT-qU-nyd" customClass="AGSSceneView">
+                                <rect key="frame" x="0.0" y="44" width="414" height="769"/>
+                            </view>
+                            <toolbar opaque="NO" clearsContextBeforeDrawing="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="Sqe-bQ-Lz3">
+                                <rect key="frame" x="0.0" y="813" width="414" height="49"/>
+                                <items>
+                                    <barButtonItem enabled="NO" title="Cockpit View" id="G3R-3A-vJ7">
```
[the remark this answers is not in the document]
yo1995: I've applied the suggestion from Mike to make those 2 buttons into 1. Also I changed the other to `Camera Settings` so that their widths are more visually balanced 😆

yo1995:
> I'm still able to select the other view options despite the camera settings popover being present and the other buttons being disabled.

Unfortunately that's how toolbar behaves with popover presentations. Other samples such as `Animate 3D graphic` also work like that. As long as it doesn't cause a bug, I tend to leave it as-is.

> Additionally, I can't exit the popover unless I tap the "camera" button again. Tapping the scene should dismiss the popover.

Please see the "Known issue 2" atop.

I'm not quite happy with what I came up here though. There would be a few possible solutions.

1. Don't pass through the scene view when the popover is presented (delete line 171). In this way, tapping on the scene view will dismiss the popover as normal. But there is no way to show the slider in sync with the camera, when the gestures are initiated by panning on the scene view. (which is the behavior on desktop platforms)
2. Add a single tap gesture recognizer to the scene view to dismiss the popover
3. Implement the touch delegate. i.e.

```swift
extension OrbitCameraAroundObjectViewController: AGSGeoViewTouchDelegate {
    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
        controller.dismiss(animated: true)
    }
}
```

Where the controller is a reference to the `OrbitCameraSettingsViewController` instance created by the segue.

I'll leave it unchanged and wait for your next comment.

vquach2404:
I like the changes you made to the camera view button!

> 1. But there is no way to show the slider in sync with the camera, when the gestures are initiated by panning on the scene view. (which is the behavior on desktop platforms)

IMO, I would sacrifice showing the sliders to be able to tap away from the popover. While it's nice to see the heading and pitch, being unable to dismiss the popover by tapping elsewhere is counterintuitive. Additionally, the [HIG](https://developer.apple.com/design/human-interface-guidelines/ios/views/popovers/) specifically states it as a method of of dismissing the popover. 

I don't really have a preference over the other 2 options you've listed. We can keep this conversation open for further discussion

vquach2404 (dismissed):

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraAroundObjectViewController.swift
```diff
@@ -0,0 +1,185 @@
...
+    }
+    
+    // MARK: UIViewController
+    
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        if let controller = segue.destination as? OrbitCameraSettingsViewController {
+            controller.orbitGeoElementCameraController = sceneView.cameraController as? AGSOrbitGeoElementCameraController
+            controller.planeGraphic = planeGraphic
+            controller.presentationController?.delegate = self
+            controller.preferredContentSize = CGSize(width: 300, height: 210)
```
philium, line 167: Instead of using a fixed size for the popover, consider having the settings controller size itself based on its contents. Search for `tableViewContentSizeObservation` for examples of how to do that.

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraAroundObjectViewController.swift
```diff
@@ -0,0 +1,185 @@
...
+        // the delta angle for the animation.
+        let pitchDelta = cameraController.isAutoPitchEnabled ? 0 : 90 - cameraController.cameraPitchOffset + ((planeGraphic.attributes["PITCH"] as? NSNumber)?.doubleValue ?? 0)
+        // Animate the camera so that it is at the target (cockpit), facing
+        // forward (0 deg heading), and aligned with the horizon (90 deg pitch).
+        cameraController.moveCamera(
+            withDistanceDelta: 0.1 - cameraController.cameraDistance,
+            headingDelta: -cameraController.cameraHeadingOffset,
+            pitchDelta: pitchDelta,
+            duration: 1
+        ) { _ in
```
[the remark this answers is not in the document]
yo1995: > Is there a reason why moving outside of the cockpit can't be animated?

Not sure about this - probably it was intended to show that the camera angles can be set without animation? Anyway it was from the design so I followed. 😅 

> the closure should check the finished parameter

Good call. I'll `guard animationFinished else { return }` here to return early if animation was interrupted. This also resolves https://github.com/Esri/arcgis-runtime-samples-ios/pull/1093#discussion_r680263529

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraAroundObjectViewController.swift
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Add the source code button item to the right of navigation bar.
+        (navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["OrbitCameraAroundObjectViewController", "OrbitCameraSettingsViewController"]
+        // Enable the buttons.
+        changeViewBarButtonItem.isEnabled = true
+        cameraControllersBarButtonItem.isEnabled = true
```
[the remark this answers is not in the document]
yo1995: Good point. The back story here is that I wanted to enable the buttons only when certain criteria are met. Later I decided not to hook up the model scene symbol or the scene's load closures, but forgot to remove these lines..

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraSettingsViewController.swift
```diff
@@ -0,0 +1,89 @@
...
+    weak var orbitGeoElementCameraController: AGSOrbitGeoElementCameraController?
+    weak var planeGraphic: AGSGraphic?
+    
+    private var headingObservation: NSKeyValueObservation?
+    
+    private let measurementFormatter: MeasurementFormatter = {
+        let formatter = MeasurementFormatter()
+        formatter.numberFormatter.maximumFractionDigits = 0
+        formatter.unitStyle = .short
+        formatter.unitOptions = .providedUnit
```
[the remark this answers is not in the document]
yo1995: 😅  As you may tell from the latter storyboard issues, I copy-pasted from the `Animate 3D graphic` sample and this was adopted from there... I should've examined these stuff though.

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraAroundObjectViewController.swift
```diff
@@ -0,0 +1,185 @@
...
+    // MARK: UIViewController
+    
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        if let controller = segue.destination as? OrbitCameraSettingsViewController {
+            controller.orbitGeoElementCameraController = sceneView.cameraController as? AGSOrbitGeoElementCameraController
+            controller.planeGraphic = planeGraphic
+            controller.presentationController?.delegate = self
+            controller.preferredContentSize = CGSize(width: 300, height: 210)
+        }
+    }
```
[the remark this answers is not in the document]
yo1995: Nice example! This is added to the iOS samples contrib guidelines.

yo1995:
Thanks for the feedback! I've applied most of them. Please review again.

P.S. GitHub Action seems to be down now: https://www.githubstatus.com/incidents/lc8s9rtyjmgq so no worries for the absence of a green checkmark ✅

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraSettingsViewController.swift
```diff
@@ -0,0 +1,120 @@
...
+        headingObservation = orbitGeoElementCameraController.observe(\.cameraHeadingOffset) { [weak self] _, _ in
+            DispatchQueue.main.async {
+                self?.updateUIForHeading()
+            }
+        }
+        isCameraDistanceInteractiveObservation = orbitGeoElementCameraController.observe(\.isCameraDistanceInteractive) { [weak self] cameraController, _ in
+            DispatchQueue.main.async {
+                self?.distanceInteractionSwitch.isOn = cameraController.isCameraDistanceInteractive
+            }
+        }
```
philium, line 118: A couple things:

- Since the properties being observed are set on the main queue, the change handler will be called on the main thread, so dispatching isn't necessary.
- For the heading observation, `updateUIForHeading()` is being called twice: once for the observation, and then again inside `headingValueChanged(_:)` (on line `76`). I would suggest removing that one and just use the observation to update the UI.

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraAroundObjectViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+    
+    @IBAction func changeViewBarButtonItemTapped(_ sender: UIBarButtonItem) {
+        moveCameraAnimationCancelable?.cancel()
+        if sender.title == "Cockpit View" {
+            cockpitViewBarButtonItemTapped(sender)
+            sender.title = "Center View"
+        } else if sender.title == "Center View" {
+            centerViewBarButtonItemTapped(sender)
+            sender.title = "Cockpit View"
+        }
```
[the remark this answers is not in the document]
yo1995: I agree, and I remember you've mentioned it here: https://github.com/Esri/arcgis-runtime-samples-ios/pull/973#discussion_r500774882 . 

Enum is the right way to go, but sometimes I feel adding an enum only for 1 button is an overkill so I took the shortcut to be lazy... 😅  Let me fix that.

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraAroundObjectViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+        let pitchDelta = cameraController.isAutoPitchEnabled ? 0 : 90 - cameraController.cameraPitchOffset + (planeGraphic.attributes["PITCH"] as! Double)
+        // Animate the camera so that it is at the target (cockpit), facing
+        // forward (0 deg heading), and aligned with the horizon (90 deg pitch).
+        moveCameraAnimationCancelable = cameraController.moveCamera(
+            withDistanceDelta: 0.1 - cameraController.cameraDistance,
+            headingDelta: -cameraController.cameraHeadingOffset,
+            pitchDelta: pitchDelta,
+            duration: 1
+        ) { [weak self] animationFinished in
+            self?.moveCameraAnimationCancelable = nil
```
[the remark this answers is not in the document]
yo1995: I'm following the discussion here: https://github.com/Esri/arcgis-runtime-samples-ios/pull/1074#discussion_r641710722 - nullify the cancelable task once it's finished.

The `cancel()` is called in `changeViewBarButtonItemTapped`, so that if the cancelable is non-nil (which means animation in action), we can cancel it before the camera moves.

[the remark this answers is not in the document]
philium: Ah, I see. I think I was thinking that this was explicitly calling `cancel()`, not just setting the cancelable to `nil`. Never mind!

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraSettingsViewController.swift
yo1995: The second bug is a nice catch! 🤦

philium: The reason I suggested using the observer to update the UI is that you can specify `initial` as an option to update the UI with the initial value. It doesn't really make a difference because we know that no one outside of this view controller modifies `isCameraDistanceInteractive`, but this view controller doesn't know that and shouldn't really know that, so updating the UI to match the current value when it is first displayed is a good practice.

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraSettingsViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        updateUIForHeading()
+        updateUIForPitch()
+        distanceInteractionSwitch.isOn = orbitGeoElementCameraController.isCameraDistanceInteractive
+        // Add an observer to sync the UI for camera heading.
+        headingObservation = orbitGeoElementCameraController.observe(\.cameraHeadingOffset) { [weak self] _, _ in
+            self?.updateUIForHeading()
+        }
+        isCameraDistanceInteractiveObservation = orbitGeoElementCameraController.observe(\.isCameraDistanceInteractive) { [weak self] cameraController, _ in
+            self?.distanceInteractionSwitch.isOn = cameraController.isCameraDistanceInteractive
+        }
```
philium, line 113:
```suggestion
        updateUIForPitch()
        // Add an observer to sync the UI for camera heading.
        headingObservation = orbitGeoElementCameraController.observe(\.cameraHeadingOffset, options: .initial) { [weak self] _, _ in
            self?.updateUIForHeading()
        }
        isCameraDistanceInteractiveObservation = orbitGeoElementCameraController.observe(\.isCameraDistanceInteractive, options: .initial) { [weak self] cameraController, _ in
            self?.distanceInteractionSwitch.isOn = cameraController.isCameraDistanceInteractive
        }
```

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/OrbitCameraSettingsViewController.swift
yo1995: Ah now I see what you means! Thanks for the suggestion.

philium (dismissed):

arcgis-ios-sdk-samples/Scenes/Orbit camera around object/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Fix the camera to point at and rotate around a target object.
+
+![Image of orbit camera around object](orbit-camera-around-object.png)
+
+## Use case
+
+The orbit geoelement camera controller provides control over the following camera behaviors:
+
+* automatically track the target
```
saratchandrakarumuri, line 11: I feel it is better to start all this bullet points in upper case instead of lower case

saratchandrakarumuri (commented):
Samples looks good other than the small comment below

philium (approved):

vquach2404 (approved):

saratchandrakarumuri (approved):

PR #1102, opened 2021-09-01
Update: Add buttons and links to sample viewer

vquach2404:
This PR responds to `common-samples/issues/3001`. I originally suggested these ideas to Divesh. These buttons will open URLs to Esri community and the developer site. The addition of these buttons will synchronize our products and allow the user easy access to more of our resources. Find the new additions in the app info view using the info icon in the top left corner of the home page.

### DIFF
--- a/arcgis-ios-sdk-samples/Content Display Logic/Controllers/AppInfoViewController.swift
+++ b/arcgis-ios-sdk-samples/Content Display Logic/Controllers/AppInfoViewController.swift
@@ -31,7 +31,7 @@ class AppInfoViewController: UIViewController {
         
         self.preferredContentSize = CGSize(width: 350, height: 350)
     }
-   
+    
     // MARK: - Actions
     
     @IBAction func esriCommunityAction() {


yo1995 (approved):
Changes look good. My only question is the arrangement of the buttons - should we come up with a future-proof layout for a series of buttons? My other comments are in 3001.

I'd suggest adding Divesh to the reviewers.

vquach2404:
> Changes look good. My only question is the arrangement of the buttons - should we come up with a future-proof layout for a series of buttons? My other comments are in 3001.

Hm I'm not sure I understand. How would adding/embedding the buttons in different views make it easier or better for the future?

yo1995:
> How would adding/embedding the buttons in different views make it easier or better for the future?

If we group the buttons in a stack view or a table view somehow, then it would take less efforts in the future to add more buttons, which means less line-of-change in both code and storyboard.

For example, say we want to add another "Download offline data" button in the future. If we add the button separately, then it would be similar as this PR: adding a button to the storyboard, add 4 constraints, and create a target-action binding for the button. While using a stack/table view means we only need to append a button/table cell to the view without setting its constraints.

But it might not worth the effort to make the UI future proof, so it's just a thought. I'd suggest adhering to Divesh's feedback.

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/Main.storyboard
```diff
@@ -78,13 +78,33 @@
                                     <action selector="closeAction" destination="HRz-Eg-7ly" eventType="touchUpInside" id="JSz-9v-XS9"/>
                                 </connections>
                             </button>
+                            <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="system" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="erh-VF-k0q">
```
philium, line 81: These buttons seem a little small. They have a point size of 15, but I think the standard size is 17. If you create a new button, it actually doesn't have a font at all (and presumably uses the default system font and size). I couldn't find a way to remove the font from a button (so that it goes back to having no font), but you could re-create the buttons to achieve that.

arcgis-ios-sdk-samples/Content Display Logic/Storyboard/Main.storyboard
```diff
@@ -78,13 +78,33 @@
...
+                                <rect key="frame" x="52" y="491" width="155" height="30"/>
+                                <state key="normal" title="Help"/>
```
philium, line 83: I was surprised to see that this button was titled Help, but it took me to a community forum. I would expect it to take me to a help page specifically for the sample view, which would then have links to the guide, the reference doc, the community, etc. That would also mean we would only need one button, not two. I was also confused by the title of the other button, for I expected it to show me more info about the sample app, not the SDK itself.

philium (commented):
I would suggest putting the button within a stack view as well. Really all of the views in the center there could be laid out with stack views, but that is work for another PR. However, stack views are much easier to re-arrange. If you put the button in a stack view, it would be easier to then go and put that stack view in another stack view if the info view were to ever be updated.

dg0yal:
@vquach2404 Can you add a screenshot showing what this looks like in the app?

vquach2404:
@dg0yal On the right is my original design. On the left is the modifications I made based on some of the review comments. I tried to make the titles a bit more descriptive. Please let me know what you think! 
| Original | Modified |
|----------|----------|
| Simulator Screen Shot - iPhone 12 Pro Max - 2021-09-27 at 15 24 53 |  Simulator Screen Shot - iPhone 12 Pro Max - 2021-09-27 at 16 12 44 |

dg0yal:
@vquach2404 Looks ok, but consider - 
1) Removing the additional link  _`ArcGIS for iOS`_ and just inserting a hyperlink into existing text like so :  _`Built with [ArcGIS Runtime SDK for iOS]()`_
2) For the _`Esri Community`_ link, perhaps add some preamble. For eg: _`Got a question about the SDK? Ask the [Esri Community]()`_
3) Adding a way for people to get the source code for the app and submit feedback. Something like : _`Check out this app's source code on [GitHub]() and log an issue or enhancement request.`_

yo1995:
field-maps

I tried out Field maps and it uses an About page like this. I think it is pretty similar to lots of other apps - using a grouped table view for all the app info. A few benefits

- We don't need to create separate buttons when adding more info to this page
- The UI is more consistent
- Using table cells to show/push view controllers seems more natural in terms of navigation
- More consistent with other Esri apps

I'd [still](https://github.com/Esri/arcgis-runtime-samples-ios/pull/1102#issuecomment-912048605) encourage using table view, given that Divesh asked for more info to be shown alongside the buttons. We can put them in section header and footers like the screenshot.

vquach2404:
Thanks everyone for the feedback! I've converted this issue into a draft for now so I can research and consider all options.

PR #1103, opened 2021-09-01
New: Swift UI map view sample

yo1995:
more to add here

### DIFF
--- a/arcgis-ios-sdk-samples/Maps/Display a map SwiftUI/DisplayMapSwiftUIView.swift
+++ b/arcgis-ios-sdk-samples/Maps/Display a map SwiftUI/DisplayMapSwiftUIView.swift
@@ -1,0 +1,73 @@
+// Copyright 2021 Esri
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+import ArcGIS
+
+struct DisplayMapSwiftUIView: View {
+    /// A boolean that indicates whether the action sheet is presented.
+    @State private var showingOptions = false
+    /// An object used for setting up a `SwiftUIMapView`.
+    @StateObject private var mapViewContext = MapViewContext()
+    
+    var body: some View {
+        VStack {
+            SwiftUIMapView(mapViewContext: mapViewContext)
+                .onMapViewTap { _, mapPoint in
+                    let graphic = AGSGraphic(geometry: mapPoint, symbol: nil)
+                    mapViewContext.graphicsOverlay.graphics.add(graphic)
+                    mapViewContext.graphicsEmpty = false
+                }
+                .edgesIgnoringSafeArea(.all)
+            HStack {
+                Button("Change Basemap") {
+                    showingOptions = true
+                }
+                .actionSheet(isPresented: $showingOptions) {
+                    ActionSheet(
+                        title: Text("Choose a basemap."),
+                        buttons: [
+                            .default(Text("Topographic")) {
+                                let basemap = AGSBasemap(style: .arcGISTopographic)
+                                mapViewContext.map.basemap = basemap
+                            },
+                            .default(Text("Oceans")) {
+                                let basemap = AGSBasemap(style: .arcGISOceans)
+                                mapViewContext.map.basemap = basemap
+                            },
+                            .default(Text("Navigation")) {
+                                let basemap = AGSBasemap(style: .arcGISNavigation)
+                                mapViewContext.map.basemap = basemap
+                            },
+                            .cancel()
+                        ]
+                    )
+                }
+                Spacer()
+                Button("Clear Graphics") {
+                    mapViewContext.graphicsOverlay.graphics.removeAllObjects()
+                    mapViewContext.graphicsEmpty = true
+                }
+                .disabled(mapViewContext.graphicsEmpty)
+            }
+            .padding()
+        }
+    }
+}
+
+struct DisplayMapSwiftUIView_Previews: PreviewProvider {
+    static var previews: some View {
+        DisplayMapSwiftUIView()
+    }
+}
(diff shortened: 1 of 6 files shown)


arcgis-ios-sdk-samples/Maps/Display a map SwiftUI/DisplayMapSwiftUIView.swift
```diff
@@ -0,0 +1,73 @@
...
+// limitations under the License.
+
+import SwiftUI
+import ArcGIS
+
+struct DisplayMapSwiftUIView: View {
+    /// A boolean that indicates whether the action sheet is presented.
+    @State private var showingOptions = false
+    /// An object used for setting up a `SwiftUIMapView`.
+    @ObservedObject private var mapViewContext = MapViewContext()
```
yo1995, line 22: iOS 13 doesn't support [`StateObject`](https://developer.apple.com/documentation/swiftui/stateobject) so I have to use [`ObservedObject`](https://developer.apple.com/documentation/swiftui/observedobject) here instead. Also please see my comment about `EnvironmentObject` in `SwiftUIMapView`.

yo1995 (commented):
A few comments below.

arcgis-ios-sdk-samples/Maps/Display a map SwiftUI/SwiftUIMapView.swift
```diff
@@ -0,0 +1,88 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+import ArcGIS
+
+struct SwiftUIMapView: UIViewRepresentable {
```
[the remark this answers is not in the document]
yo1995: Thanks for the suggestions and the rewrite. They are really helpful. 🙇  I've applied most of the suggestions. Please see my other comments below.

arcgis-ios-sdk-samples/Maps/Display a map SwiftUI/SwiftUIMapView.swift
```diff
@@ -0,0 +1,88 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+import ArcGIS
+
+struct SwiftUIMapView {
```
yo1995, line 18: > I'm guessing, based on how they are written, that the types in this file aren't intended to be reused by multiple samples. Therefore they should be made private.

> ... implementation can be more easily copied-and-pasted into another sample

Discussed with Divesh and he said this will be the only SwiftUI sample in the current sample viewer. Thus I'll keep them `internal` and in this separate file, so that the file structure is clearer.

arcgis-ios-sdk-samples/Maps/Display a map SwiftUI/SwiftUIMapView.swift
```diff
@@ -0,0 +1,88 @@
...
+}
+
+extension SwiftUIMapView {
+    /// You can use this coordinator to implement common Cocoa patterns, such as
+    /// delegates, data sources, and responding to user events via target-action.
+    class Coordinator: NSObject {
+        var onSingleTapAction: ((CGPoint, AGSPoint) -> Void)?
+        
+        init(
+            onSingleTapAction: ((CGPoint, AGSPoint) -> Void)?
```
yo1995: After reading the SwiftUI tutorials, I'm still not fully understanding what a [`Coordinator`](https://developer.apple.com/documentation/swiftui/uiviewrepresentable/coordinator) should hold. In the [tutorial](https://developer.apple.com/tutorials/swiftui/interfacing-with-uikit#:~:text=declare%20a%20nested%20coordinator%20class%20inside%20pageviewcontroller) it holds its parent, but IMO it can hold anything, including those which can also be the stored properties of the parent view struct. A few possible options...

- Hold its parent (so that it can access its parent's properties)
- Hold the target-action delegate methods and corresponding action closures
- Hold everything. In our case, the map, the overlays array, and the `onSingleTapAction` closure

Can you help me to understand which is preferred and why? Thanks.

arcgis-ios-sdk-samples/Maps/Display a map SwiftUI/DisplayMapSwiftUIView.swift
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+    
+    func clearGraphics() {
+        graphicsOverlay.graphics.removeAllObjects()
+        updateClearGraphicsButtonState()
+    }
+    
+    func updateClearGraphicsButtonState() {
+        clearGraphicsButtonDisabled = (graphicsOverlay.graphics as! [AGSGraphic]).isEmpty
```
yo1995, line 81: As you may see, I hesitated about the button state binding in 90ce1a6 😵.  Is it a pure styling choice to use methods over observations to decide the button state, or are there other reasons?

arcgis-ios-sdk-samples/Maps/Display a map SwiftUI/SwiftUIMapView.swift
```diff
@@ -0,0 +1,90 @@
...
+    
+    init(
+        map: AGSMap,
+        graphicsOverlays: [AGSGraphicsOverlay] = []
+    ) {
+        self.map = map
+        self.graphicsOverlays = graphicsOverlays
+    }
+    
+    private var onSingleTapGesture: ((CGPoint, AGSPoint) -> Void)?
```
yo1995, line 30: Chang the naming to follow the new design. 😏

yo1995 (commented):
Thanks for the detailed response!

yo1995:
Thanks for the feedback!

---

> * In the video, basemaps are changed seamlessly, but for me VP is getting reset when ever map is changed. In regular change basemap sample, this doesn't happen. Is this expected?

Yes it is expected. At the beginning I swapped the basemap (same as the original `Change basemap` sample), but later on we decide to swap the map, which resets and recreates the map.

> * On iOS 13 and 14, `Read Me` and `back` button are working fine for me on an iPad. Maybe phone factor specific issue?

Yes it might be a phone specific issue. On iPad the split view seems to work fine. Maybe it is an iOS bug and isn't a big deal.

> * On iOS 13 and 14, Source code page is rendering empty as expected, 

Hmm, no good. I also see the source code page empty on iOS 13.

> but if the device is locked when in that page, app is crashing at `SourceCodeViewController.swift` at following line

It would crash also if I rotate the device on this page. The `currentFilename` is empty and got force unwrapped, which crashes the app. I'll check why it is empty.

> * When tried to open the same sample when you are already in the sample, I am getting the following error in console and black blank screen instead of mapView

Confirmed on iPad Simulator, it seems opening the sample on even tries will give an empty mapView. Maybe something is holding the sample from being released. I'll double check.

PR #1107, opened 2021-09-22
Fix file error

vquach2404:
This PR resolves an issue brought to my attention by Divesh. After launching the "Identify raster cell" sample, the data did not load properly and an error message said "File error". See `common-samples/3016` for more info.

### DIFF
--- a/arcgis-ios-sdk-samples/Layers/Identify raster cell/IdentifyRasterCellViewController.swift
+++ b/arcgis-ios-sdk-samples/Layers/Identify raster cell/IdentifyRasterCellViewController.swift
@@ -35,7 +35,11 @@ class IdentifyRasterCellViewController: UIViewController {
     // MARK: Properties
     
     /// The raster layer created using local raster file.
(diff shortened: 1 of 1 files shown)


yo1995:
It is a tricky issue and I'm not quite sure what's the root cause. My speculations below...

- The `AGSRaster.init(name:extension:)` initializer uses an underlying method `ags_pathForResource`, which can be found in Cocoa repo under `NSFileManager+AGSAdditions`
- This method recursively check the main bundle folders and try to match the `SA_EVI_8Day_03May20.tif` file.
- When the app is running in Simulator or built to phone without going through App Store, all the asset packs, on-demand resources are bundled in the app itself (which bloats it to ~500MB, comparing to the ~90MB released version)
- Due to some unknown reason, the path to this tif file cannot be found for App Store released version 🤔 

I'd suggest...

- Create a TestFlight upload using this branch, and test if the bug fix works well with App Store Connect
- If so, we can merge this PR
- If not, we need to figure out why on-demand is behaving weirdly in this case...

yo1995 (approved):

philium (approved):

saratchandrakarumuri (approved):

PR #1119, opened 2021-10-29
Remove ArcGIS vector tiled layer URL sample

vquach2404:
This PR addresses `common-samples/3023` and removes a redundant sample.


yo1995:
Did you checkout your branch from `main`? 🤔  It would be better if you checkout from `v.next` so that we don't mess up the commit history... but it is not a big deal.

vquach2404:
> Did you checkout your branch from `main`? 🤔 It would be better if you checkout from `v.next` so that we don't mess up the commit history... but it is not a big deal.

Oh yes I did. I can redo it

PR #1123, opened 2021-11-16
Update: Remove 2 unused exception HTTP URLs for `Display KML network links`

yo1995:
This PR removes the unused HTTP exception URLs for `Display KML network links` sample, which was added in #500 . Now that the dataset has been updated all URLs to HTTPS, these 2 can be removed.

Please see `common-samples/issues/3160` for details. I reviewed their `git blame` history and they seem to be solely used by this sample.

To double check this doesn't affect other samples, we can go through every KML sample to see if they are working fine. (but I really don't think that is necessary 😏 )

### DIFF
--- a/arcgis-ios-sdk-samples/Info.plist
+++ b/arcgis-ios-sdk-samples/Info.plist
@@ -53,13 +53,6 @@
 	<dict>
 		<key>NSExceptionDomains</key>
 		<dict>
-			<key>buienradar.nl</key>
-			<dict>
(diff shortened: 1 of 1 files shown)


philium (approved):

vquach2404 (approved):
Tested during stand up today and works well

yo1995:
Hey Sarat, feel free to take a look.

saratchandrakarumuri (approved):
Sample works as expected after the change

PR #1126, opened 2021-12-11
New: `Browse building floors`

yo1995:
## Description

This PR implements `Browse building floors` in `Maps` category.
Branch: [here](https://github.com/Esri/arcgis-runtime-samples-ios/tree/Ting/New-BrowseBuildingFloors-vNext/arcgis-ios-sdk-samples/Maps/Browse%20building%20floors)

## Linked Issue(s)

`common-samples/issues/2983`

## How To Test

Run the sample and ensure all the floors are loading OK, no UI bug.

## Screenshots

curr

## To Discuss

Instead of the classic toolbar, action sheet design, I chose to use a picker here to better match the Android design (`common-samples/pull/3206`) that uses a "spinner" control. Also the UI similarity makes it an easier transition to the future toolkit component: https://github.com/Esri/arcgis-runtime-toolkit-ios/pull/162


yo1995:
Hey Vivian, feel free to start reviewing the code - I'll add the README later when it is ready.

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+        (navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["BrowseBuildingFloorsViewController"]
+        
+        // Set the appearance of the floor level picker view.
+        setPickerViewLayout()
+        
+        // A floor-aware web map for floors of Esri Building L in Redlands.
+        let map = AGSMap(item: AGSPortalItem(
+            portal: .arcGISOnline(withLoginRequired: false),
+            itemID: "f133a698536f44c8884ad81f80b6cfc7"
+        ))
```
vquach2404, line 66: Can this be set within the global property?

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+        
+        map.load { [weak self] error in
```
vquach2404, line 68: I would also suggest making a separate function for loading the map and floor manager to keep `viewDidLoad` succinct.

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
yo1995: It can be an instance property, but there is no need to have it held by the view controller. 
Phil usually suggests using local variables when possible, so I create the map locally here.

vquach2404 (approved):
Looking good!

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+            mapView.map = makeMap()
+        }
+    }
+    /// The floor level picker view.
+    @IBOutlet var floorLevelPickerView: UIPickerView!
+    
+    // MARK: Properties
+    
+    /// The levels of the floor-aware web map, with an addition of `nil`.
+    var levels: [AGSFloorLevel?] = [nil]
```
philium, line 33: This is the first time I have seen an array of optionals in a long time because the use case for that sort of thing is fairly rare. I can't help but wonder if there's a different solution, which I'll touch on in subsequent comments.

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+    
+    // MARK: Methods
+    
+    /// Set the selected level to visible and all other levels to invisible.
+    /// - Parameter selectedLevel: The `AGSFloorLevel` selected.
+    func setVisibleFloorLevel(_ selectedLevel: AGSFloorLevel?) {
+        levels.forEach { level in
+            level?.isVisible = level == selectedLevel
+        }
```
philium, line 42: I know that our floor manager only has three levels, but users may have significantly more. Iterating through all of them and update the `isVisible` property of all of them is highly inefficient. Instead, it would be better to store the selected floor and use that to update the `isVisible` property of the previously selected floor and the newly selected floor:

```swift
var selectedFloorLevel: AGSFloorLevel? {
    didSet {
        oldValue?.isVisible = false
        selectedFloorLevel?.isVisible = true
        
        ... // Update picker view selection.
    }
}
```

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+    }
+    
+    func setPickerViewLayout() {
+        floorLevelPickerView.translatesAutoresizingMaskIntoConstraints = false
+        let margin: CGFloat = 10.0
+        floorLevelPickerView.layer.cornerRadius = margin
+        NSLayoutConstraint.activate([
+            floorLevelPickerView.widthAnchor.constraint(equalToConstant: 120.0),
+            floorLevelPickerView.bottomAnchor.constraint(equalTo: mapView.attributionTopAnchor, constant: -margin),
+            floorLevelPickerView.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor, constant: -margin)
```
philium, line 52: Consider using system spacing.
```suggestion
            view.safeAreaLayoutGuide.trailingAnchor.constraint(equalToSystemSpacingAfter: floorLevelPickerView.trailingAnchor, multiplier: 1),
            mapView.attributionTopAnchor.constraint(equalToSystemSpacingBelow: floorLevelPickerView.bottomAnchor, multiplier: 1)
```

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+            itemID: "f133a698536f44c8884ad81f80b6cfc7"
+        ))
+        
+        map.load { [weak self] error in
+            guard error == nil, let self = self,
+                  // The floor manager of the web map, which exposes the sites,
+                  // facilities, and levels of the floor-aware data model.
+                  let floorManager = map.floorManager else { return }
+            
+            floorManager.load { error in
```
philium, line 70: This closure is strongly capturing `self`, which it shouldn't be. Since it is often hard to track that sort of thing through nested closures, what I have done and what I would recommend doing is delegate each completion handler to a method. That way, `self` only needs to be unwrapped through optional chaining. It might look something like this:

```swift
func makeMap() -> AGSMap {
    // A floor-aware web map for floors of Esri Building L in Redlands.
    let map = AGSMap(item: AGSPortalItem(
        portal: .arcGISOnline(withLoginRequired: false),
        itemID: "f133a698536f44c8884ad81f80b6cfc7"
    ))
    
    map.load { [weak self] error in
        self?.map(map, didLoadWith: error)
    }
    
    return map
}

func map(_ map: AGSMap, didLoadWith loadError: Error?) {
    guard loadError == nil else {
        return
    }
    
    // The floor manager of the web map, which exposes the sites,
    // facilities, and levels of the floor-aware data model.
    guard let floorManager = map.floorManager else {
        return
    }
    
    floorManager.load { [weak self] error in
        self?.floodManager(floorManager, didLoadWith: error)
    }
}

func floodManager(_ floorManager: AGSFloorManager, didLoadWith loadError: Error?) {
    guard loadError == nil else {
        return
    }
    
    guard let geometry = floorManager.sites.first?.geometry else {
        return
    }
    
    mapView.setViewpointGeometry(geometry)
    
    // Update floor picker and select the first floor.
    floorLevels = floorManager.levels
    let firstFloorIndex = floorLevels.firstIndex { $0.longName == "Level 1" }!
    selectedFloorLevel = floorLevels[firstFloorIndex]
}
```

You also might consider at least printing the load errors.

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        (navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["BrowseBuildingFloorsViewController"]
+        
+        // Set the appearance of the floor level picker view.
+        setPickerViewLayout()
+    }
+}
+
+// MARK: - UIPickerViewDataSource and UIPickerViewDelegate
+
+extension BrowseBuildingFloorsViewController: UIPickerViewDataSource, UIPickerViewDelegate {
```
philium, line 99: I would suggesting splitting this into two extensions. When adding conformance via an extension, standard practice is one extension per protocol.

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
philium, line 33: You might consider renaming this property to `floorLevels`.

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        levels.count
+    }
+    
+    func pickerView(_ pickerView: UIPickerView, titleForRow row: Int, forComponent component: Int) -> String? {
+        levels[row]?.shortName ?? "None"
+    }
+    
+    func pickerView(_ pickerView: UIPickerView, didSelectRow row: Int, inComponent component: Int) {
+        setVisibleFloorLevel(levels[row])
+    }
```
philium, line 114: Remember how I mentioned not using an array of optionals? Here's how it can be done with an array of non-optionals:

```swift
extension BrowseBuildingFloorsViewController: UIPickerViewDataSource {
    func numberOfComponents(in pickerView: UIPickerView) -> Int {
        1
    }
    
    func pickerView(_ pickerView: UIPickerView, numberOfRowsInComponent component: Int) -> Int {
        floorLevels.count + 1
    }
    
    func pickerView(_ pickerView: UIPickerView, titleForRow row: Int, forComponent component: Int) -> String? {
        let index = row - 1
        return index >= floorLevels.startIndex ? floorLevels[index].shortName : "None"
    }
}

extension BrowseBuildingFloorsViewController: UIPickerViewDelegate {
    func pickerView(_ pickerView: UIPickerView, didSelectRow row: Int, inComponent component: Int) {
        let index = row - 1
        selectedFloorLevel = index >= floorLevels.startIndex ? floorLevels[index] : nil
    }
}
```

And then in the property observer to update the picker view:

```swift
let row: Int
if let selectedFloorLevel = selectedFloorLevel {
    row = floorLevels
        .firstIndex(of: selectedFloorLevel)
        .map { $0 + 1 } ?? .zero
} else {
    row = .zero
}
floorLevelPickerView.selectRow(row, inComponent: 0, animated: true)
```

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
yo1995: Thanks for pointing this out! New method learned... 😸

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
yo1995: Great suggestion!

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
yo1995: > strongly capturing self, which it shouldn't be.

My bad. I mis-thought the lifespan of map view should be the same as `floorManager`, so strongly captured.

arcgis-ios-sdk-samples/Maps/Browse building floors/BrowseBuildingFloorsViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+        ))
+        map.load { [weak self] error in
+            self?.map(map, didLoadWith: error)
+        }
+        return map
+    }
+    
+    func map(_ map: AGSMap, didLoadWith loadError: Error?) {
+        guard loadError == nil else {
+            presentAlert(error: loadError!)
```
philium, line 79: Having to force unwrap something checked for `nil` in a guard statement is a code smell and can almost always be avoided. Instead, I would suggest moving this check to where this method (`map(_:didLoadWith:)`) is called:

```swift
if let error = error {
    self?.mapDidLoad(map)
} else {
    self?.presentAlert(error: error)
}
```

Similar with the error in `floorManager(_:didLoadWith:)`.

philium (approved):

saratchandrakarumuri (approved):
Sample looks good

PR #150, opened 2016-03-18
Added new samples

singhgag:
Route and Navigation > Route around barriers
Layers > Export tiles
Features > Generate geodatabase
Edit data > Offline edit and sync

PR #1139, opened 2022-02-04
Fix: `Route around barriers` misc issues

yo1995:
## Description

This PR fixes `Route around barriers` in `Route and directions` category. The sample was first added in #150

## Linked Issue(s)

`common-samples/issues/814`

## How To Test

Run the sample, and solve a route.

- The solved route graphic should be padded with a margin to the screen
- No warning message in the console

@philium when you have a chance, can you elaborate on the first 2 issues you mentioned in the original post?

I cannot figure out which part of sample code should be improved, or maybe they are already fixed in the SDK or core.

## To Discuss

- Respond appropriately if the service is down. Just got this in the console.

I cannot repro this issue, so didn't add error handling.

- Runtime Error for some portions of the route

I'm not sure where does this come from either. 😳 

- Pad the portion of the route more as you display it

I padded the viewpoint around the solved route geometry.

### DIFF
--- a/arcgis-ios-sdk-samples/Route and directions/Route around barriers/RouteAroundBarriersViewController.swift
+++ b/arcgis-ios-sdk-samples/Route and directions/Route around barriers/RouteAroundBarriersViewController.swift
@@ -73,7 +73,7 @@ class RouteAroundBarriersViewController: UIViewController, AGSGeoViewTouchDelega
         // hide directions list
         self.setRouteDetailsVisibility(visible: generatedRoute != nil, animated: false)
     }
-
+    
     // MARK: - Route logic
     
     func getDefaultParameters() {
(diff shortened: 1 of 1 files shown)


philium:
> I cannot repro this issue… I'm not sure where does this come from

This might have been a service issue. It is possible that this sample is now using a different service or that the reliability of the service has improved. If it happens again, we can create another issue.

philium:
I'm seeing the following logged to the console when the sample is opened:

> [TableView] Warning once only: UITableView was told to layout its visible cells and other contents without being in the view hierarchy (the table view or one of its superviews has not been added to a window). This may cause bugs by forcing views inside the table view to load and perform layout without accurate information (e.g. table view bounds, trait collection, layout margins, safe area insets, etc), and will also cause unnecessary performance overhead due to extra layout passes. Make a symbolic breakpoint at UITableViewAlertForLayoutOutsideViewHierarchy to catch this in the debugger and see what caused this to occur, so you can avoid this action altogether if possible, or defer it until the table view has been added to a window.

Could you please investigate?

yo1995:
> I'm seeing the following logged to the console when the sample is opened:
> Could you please investigate?

I think the problem is that the directions lists' height being zero when it's first laid out.

The `DirectionsListViewController` is embedded in a container view and instantiated when the container view loads. In the code, it refers to the container view's `directionsBottomConstraint` from the storyboard, and before I fixed [here](https://github.com/Esri/arcgis-runtime-samples-ios/pull/1139/files#diff-654f4c0db276cf930c4042d231eab7a4f4a868569ebd153ffd476d9dc206e7e8L70), the constraint is set to -150, which equals the height of the container view, in `viewWillAppear` method.
That makes the container view invisible when the table view is loading, which probably caused this warning message.

I simply changed it to `viewDidAppear`, so that the table view is first laid out, then hidden by setting the bottom constraint. It solves the warning, but I cannot be totally sure if there isn't side effect 😳 .

yo1995:
It seems also related to #636

philium (approved):

vquach2404 (approved):

saratchandrakarumuri (approved):

PR #1141, opened 2022-02-11
New: Add features with contingent values

vquach2404:
## Description

    This PR implements `Add features with contingent values` in `Edit data` category.
    Branch: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Viv/sample_contingentValues

## Linked Issue(s)

`common-samples/issues/3173`

## How To Test

Tap on the map, select values, tap "Done" to add the feature

## Screenshots

add-features-contingent-values-1
add-features-contingent-values-2


arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -4761,12 +4788,15 @@
...
 					AuroraCOGeopackage,
+					BirdsNestGDB,
 					Bradley3DModel,
 					Bristol,
 					Bristol3DModel,
+					ContingentValuesBirdNests,
 					DodgeCity,
 					Dolmus3DModel,
 					EdinburghPylonsDimensionsMMPK,
+					EsriCarsMMPK,
```
yo1995, line 4799: Is this intended to be added?

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -4761,12 +4788,15 @@
 			attributes = {
 				KnownAssetTags = (
```
yo1995, line 4789: It seems there are extra tags included. From the dependency array I only see `ContingentValuesBirdNests` and `fillmoreTopographic`.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValues.storyboard
```diff
@@ -0,0 +1,225 @@
...
+                <placeholder placeholderIdentifier="IBFirstResponder" id="9Fa-V5-ysP" userLabel="First Responder" customClass="UIResponder" sceneMemberID="firstResponder"/>
+            </objects>
+            <point key="canvasLocation" x="901" y="119"/>
+        </scene>
+    </scenes>
+    <resources>
+        <namedColor name="statusLabelBackgroundColor">
+            <color red="1" green="1" blue="1" alpha="0.60000002384185791" colorSpace="custom" customColorSpace="sRGB"/>
+        </namedColor>
+        <systemColor name="systemBackgroundColor">
```
yo1995, line 218: Please use tertiarySystemBackgroundColor for the entry point view controller.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,332 @@
...
+        // Select the buffer size if it is the only option available.
+        if bufferSizes?.count == 1 {
+            selectedBufferSize = bufferSizes?[0]
+        }
+        // Update the picker view.
+        bufferSizePickerView.reloadAllComponents()
+    }
+    
+func createFeature(with status: AGSCodedValue) {
+    // Get the contingent values defintion from the feature table.
```
yo1995, line 194:
```suggestion
    // Get the contingent values definition from the feature table.
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,332 @@
...
+        if selectedProtection == nil {
+            bufferSizeCell.textLabel?.isEnabled = false
+            bufferSizeCell.isUserInteractionEnabled = false
+        } else {
+            bufferSizeCell.textLabel?.isEnabled = true
+            bufferSizeCell.isUserInteractionEnabled = true
+        }
+    }
+    
+    /// Toggles visisbility of the buffer size scale picker.
```
yo1995, line 249:
```suggestion
    /// Toggles visibility of the buffer size scale picker.
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,332 @@
...
+    }
+}
+
+extension ContingentValuesTableViewController: UIPickerViewDelegate {
+    func pickerView(_ pickerView: UIPickerView, titleForRow row: Int, forComponent component: Int) -> String? {
+        if let bufferSizes = bufferSizes {
+            let bufferSizeTitle = String(bufferSizes[row])
+            return bufferSizeTitle
+        }
+        return ""
```
yo1995, line 324:
```suggestion
```
Since this delegate method returns a optional string, no need to return an empty string.

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -4761,12 +4788,15 @@
...
 					AuroraCOGeopackage,
+					BirdsNestGDB,
```
vquach2404, line 4791:
```suggestion
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
yo1995: I made the wrong suggestion here 🤥 . It should be `return nil`. No need to fix now, I'll make another suggestion.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,331 @@
...
+        return bufferSizes!.count
+    }
+}
+
+extension ContingentValuesTableViewController: UIPickerViewDelegate {
+    func pickerView(_ pickerView: UIPickerView, titleForRow row: Int, forComponent component: Int) -> String? {
+        if let bufferSizes = bufferSizes {
+            let bufferSizeTitle = String(bufferSizes[row])
+            return bufferSizeTitle
+        }
```
yo1995, line 323:
```suggestion
        } else {
            return nil
        }
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,202 @@
...
+        // Use the vector tiled layer as a basemap.
+        let basemap = AGSBasemap(baseLayer: fillmoreVectorTiledLayer)
+        let map = AGSMap(basemap: basemap)
+        return map
+    }
+    
+    /// Add a single feature to the map.
+    func addFeature(at mapPoint: AGSPoint) {
+        // Create a symbol to represent a bird's nest.
+        let symbol = AGSSimpleMarkerSymbol(style: .circle , color: .black, size: 11)
```
yo1995, line 116:
```suggestion
        let symbol = AGSSimpleMarkerSymbol(style: .circle, color: .black, size: 11)
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,331 @@
...
+    func showProtectionOptions() {
+        // Get the previously selected protection value.
+        let previouslySelectedProtection = selectedProtection
+        guard let feature = feature else { return }
+        // Get the contingent value results with the feature for the protection field.
+        let contingentValuesResult = featureTable?.contingentValues(with: feature, field: "Protection")
+        // Get contingent coded values by field group.
+        guard let protectionGroupContingentValues = contingentValuesResult?.contingentValuesByFieldGroup["ProtectionFieldGroup"] as? [AGSContingentCodedValue] else { return }
+        // Get the selected index if applicable.
+        let selectedIndex = protectionGroupContingentValues.firstIndex { $0.codedValue.name == self.selectedProtection?.codedValue.name} ?? nil
```
yo1995, line 156:
```suggestion
        let selectedIndex = protectionGroupContingentValues.firstIndex { $0.codedValue.name == self.selectedProtection?.codedValue.name }
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,331 @@
...
+        // Get the previously selected status value.
+        let previouslySelectedStatus = selectedStatus
+        // Get the first field by name.
+        let statusField = featureTable?.field(forName: "Status")
+        // Get the field's domains as coded value domain.
+        let codedValueDomain = statusField?.domain as! AGSCodedValueDomain
+        // Get the coded value domain's coded values.
+        let statusCodedValues = codedValueDomain.codedValues
+        // Get the selected index if applicable.
+        let selectedIndex = statusCodedValues.firstIndex { $0.name == self.selectedStatus?.name } ?? nil
```
yo1995, line 124:
```suggestion
        let selectedIndex = statusCodedValues.firstIndex { $0.name == self.selectedStatus?.name }
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,331 @@
...
+        // Validate the feature's contingencies.
+        let contingencyViolations = featureTable.validateContingencyConstraints(with: feature)
+        if contingencyViolations.isEmpty {
+            // If there are no contingency violations in the array, the feature is valid and ready to add to the feature table.
+            doneBarButtonItem.isEnabled = true
+        } else {
+            // Present an alert if there are contingency violations.
+            self.presentAlert(title: "", message: "Invalid contingent values")
+        }
+
```
yo1995, line 221:
```suggestion
```

yo1995 (commented):
A few more linter warnings.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,202 @@
...
+        do {
+            // Create a temporary directory URL.
+            let temporaryDirectoryURL = try FileManager.default.url(
+                for: .itemReplacementDirectory,
+                in: .userDomainMask,
+                appropriateFor: geodatabaseURL,
+                create: true
+            )
+            // Create a temporary URL where the geodatabase URL can be copied to.
+            let temporaryGeodatabaseURL = temporaryDirectoryURL.appendingPathComponent(ProcessInfo.processInfo.globallyUniqueString)
```
yo1995, line 44: I'm not sure if the URL is correct or not. When I put a breakpoint here and go into the folder, it seems there are 3 files stored in a folder called `(A Document Being Saved By ArcGIS Runtime SDK Samples)` and are prepended with an UUID string, without filename extension.

I'm assuming you want to put files in a temporary folder. In that case, the subfolder needs to be created first.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+    
+    /// Called in response to the geodatabase load operation completing.
+    func geodatabaseDidLoad(with result: Result<Void, Error>) {
+        switch result {
+        case .success:
+            // Get and load the first feature table in the geodatabase.
+            featureTable = geodatabase.geodatabaseFeatureTables[0] as AGSArcGISFeatureTable
+            featureTable?.load { [weak self] error in
```
yo1995, line 80: Can you try not to load the feature table, and directly access it here? 

When I put the break point on line 63 where the geodatabase is loaded, it seems the feature table is also already loaded (I can access the table name). I'm not totally sure though.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,202 @@
...
+    }
+    
+    /// Add a single feature to the map.
+    func addFeature(at mapPoint: AGSPoint) {
+        // Create a symbol to represent a bird's nest.
+        let symbol = AGSSimpleMarkerSymbol(style: .circle , color: .black, size: 11)
+        // Add the graphic to the graphics overlay.
+        self.graphicsOverlay.graphics.add(AGSGraphic(geometry: mapPoint, symbol: symbol))
+        // Show the attributes table view.
+        performSegue(withIdentifier: "AddFeature", sender: self)
```
yo1995, line 120: Phil might suggest using an `@IBSegueAction` here. See https://github.com/Esri/arcgis-runtime-samples-ios/pull/1038#discussion_r583962813 and https://github.com/Esri/arcgis-runtime-samples-ios/pull/1127#discussion_r773379574 for instance.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,202 @@
...
+    /// Create buffer graphics for the features.
+    func createBufferGraphics() {
+        // Create the query parameters.
+        let queryParameters = AGSQueryParameters()
+        // Set the where clause to filter for buffer sizes greater than 0.
+        queryParameters.whereClause = "BufferSize > 0"
+        // Create an array of graphics to add to the graphics overlay.
+        var graphics = [AGSGraphic]()
+        // Create a graphics overlay.
+        let graphicsOverlay = AGSGraphicsOverlay()
```
yo1995, line 161: Is it required to created a separate graphics overlay for each buffer graphic?

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,202 @@
...
+        // Create the outline for the buffers.
+        let lineSymbol = AGSSimpleLineSymbol(style: .solid, color: .black, width: 2)
+        // Create the buffer symbol.
+        let bufferSymbol = AGSSimpleFillSymbol(style: .forwardDiagonal, color: .red, outline: lineSymbol)
```
yo1995, line 165: Similar to a comment above, can you create a single graphics overlay for the buffer graphics, set a renderer with outlined polygon symbol, and use it here?

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,331 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+protocol ContingentValuesDelegate: AnyObject {
+    func createBufferGraphics()
```
yo1995, line 19: - Better to follow the standard naming pattern as something did finish
- It seems this method should not create all the graphics, but rather only the graphic for the newly added feature. Can you split it into 2 method, one for initializing existing graphics when the geodatabase loads, the other for adding 1 graphic when the contingent value process finishes?

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,331 @@
...
+        // Dismiss the table view.
+        dismiss(animated: true)
+    }
+    
+    @IBAction func doneBarButtonItemTapped(_ sender: Any) {
+        guard let feature = feature, let featureTable = featureTable else { return }
+        // Set the feature's geometry to the map point.
+        feature.geometry = mapPoint
+        // Add the feature to the feature table.
+        featureTable.add(feature) { _ in
```
yo1995, line 43:
```suggestion
        featureTable.add(feature) { [unowned self] _ in
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,331 @@
...
+            resetCellStates()
+        }
+    }
+    
+    /// The selected buffer size.
+    var selectedBufferSize: Int? {
+        didSet {
+            if let selectedBufferSize = selectedBufferSize {
+                // Set the the buffer size attribute to the selected integer.
+                feature?.attributes["BufferSize"] = self.selectedBufferSize
```
yo1995, line 95:
```suggestion
                feature?.attributes["BufferSize"] = selectedBufferSize
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,331 @@
...
+                let bufferSize = String(selectedBufferSize)
+                // Display the selected buffer size.
+                editRightDetail(cell: bufferSizeCell, rightDetailText: bufferSize)
+                // Validate the contingency.
+                validateContingency()
+            } else {
+                // If the value is nil, clear the right detail text.
+                editRightDetail(cell: bufferSizeCell, rightDetailText: " ")
```
yo1995, line 103: This 1 space string seems a bit odd. Can it be `nil`?

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,331 @@
...
+        bufferSizes = Array(minValue...maxValue)
+        // Select the buffer size if it is the only option available.
+        if bufferSizes?.count == 1 {
+            selectedBufferSize = bufferSizes?[0]
+        }
+        // Update the picker view.
+        bufferSizePickerView.reloadAllComponents()
+    }
+    
+func createFeature(with status: AGSCodedValue) {
```
yo1995, line 193: This method needs re-indent.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,331 @@
...
+                toggleBufferSizePickerVisibility()
+            } else {
+                return
+            }
+        default:
+            return
+        }
+    }
+        
+    /* UITableViewDataSource */
```
yo1995, line 293:
```suggestion
```

Let's use a unified comment string style 😏

yo1995 (commented):
I'll stop reviewing before getting into the contingent value APIs to avoid giving overwhelming numbers of comments. After the comments above are addressed, I'll continue the review.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
vquach2404: I can eliminate feature layer loading but we are required to load the feature table in order to load the contingent values definition (due to contingent values workflow).

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
vquach2404: This is actually not possible since each feature has a different buffer size. More info from Preeti on the Symbology team: 

> Unfortunately we don't have a mechanism to define symbols based on attribute coming from data source. In ArcGIS system this is done via VisualVariables and we currently don't have it.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
vquach2404: When I set it to nil, the text doesn't show up after re-selection (you can reproduce by replace `" "` to `nil` and make selections for all 3 fields and then change "Status". You'll notice the text in the right detail text doesn't appear). This was just an easy fix to that solution.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
yo1995: I think I might have made a wrong suggestion here - no need to change, I'll take another look in next round of review.
It seems like we can have a function to create one buffer graphic for each bird nest feature, which takes the feature and its map point as parameter. So that
- when the geodatabase loads, use a forEach to create graphics for existing features
- when new feature gets added to the geodatabase, call the function in the delegate

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+                result = .success(())
+            }
+            self?.geodatabaseDidLoad(with: result)
+        }
+    }
+    
+    deinit {
+        if let geodatabase = geodatabase {
+            geodatabase.close()
+            try? FileManager.default.removeItem(at: geodatabase.fileURL)
```
yo1995, line 67: It would be good to also remove the temporary directory that contains the copied geodatabase files.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+    
+    /// Called in response to the geodatabase load operation completing.
+    func geodatabaseDidLoad(with result: Result<Void, Error>) {
+        switch result {
+        case .success:
+            // Get and load the first feature table in the geodatabase.
+            featureTable = geodatabase.geodatabaseFeatureTables[0] as AGSArcGISFeatureTable
+            featureTable?.load { [weak self] error in
```
yo1995, line 77:
```suggestion
            let featureTable = geodatabase.geodatabaseFeatureTables[0] as AGSArcGISFeatureTable
            self.featureTable = featureTable
            featureTable.load { [weak self] error in
```
So it doesn't need to be conditionally unwrapped.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+                guard let self = self else { return }
+                if let error = error {
+                    self.presentAlert(error: error)
+                } else {
+                    // Create and load the feature layer from the feature table.
+                    let featureLayer = AGSFeatureLayer(featureTable: self.featureTable!)
```
yo1995, line 83:
```suggestion
                    let featureLayer = AGSFeatureLayer(featureTable: featureTable)
```
If the suggestion above is adopted, this can be replaced.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+                    // Add buffer graphics for the feature layer.
+                    self.queryFeatures()
+                }
+            }
+        case .failure(let error):
+            presentAlert(error: error)
+        }
+    }
+    
+    /// Creates a map with a vector tiled layer basemap.
```
yo1995, line 98:
```suggestion
    /// Create a map with a vector tiled layer basemap.
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+        // Get the vector tiled layer by name.
+        let fillmoreVectorTiledLayer = AGSArcGISVectorTiledLayer(name: "FillmoreTopographicMap")
+        // Use the vector tiled layer as a basemap.
+        let basemap = AGSBasemap(baseLayer: fillmoreVectorTiledLayer)
+        let map = AGSMap(basemap: basemap)
+        return map
+    }
+    
+    /// Create buffer graphics for the features.
+    func queryFeatures() {
```
yo1995, line 109:
```suggestion
    func queryFeatures() {
        guard let featureTable = featureTable else { return }
```
Add a guard statement at top so that when the feature table is not present, no need to create the parameters.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+        // Create the query parameters.
+        let queryParameters = AGSQueryParameters()
+        // Set the where clause to filter for buffer sizes greater than 0.
+        queryParameters.whereClause = "BufferSize > 0"
+        // Create an array of graphics to add to the graphics overlay.
+        var graphics = [AGSGraphic]()
+        // Query the features with the query parameters.
+        featureTable?.queryFeatures(with: queryParameters) { [weak self ] result, error in
```
yo1995, line 117:
```suggestion
        featureTable.queryFeatures(with: queryParameters) { [weak self ] result, error in
```
Remove ❓ if the suggestion above is adopted.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+                let features = result.featureEnumerator().allObjects
+                // For each feature, add the buffer symbol according to its buffer size.
+                features.forEach { feature in
+                    let graphic = self.createGraphics(for: feature)
+                    graphics.append(graphic)
+                }
+                // Add the graphics to the graphics overlay.
+                self.graphicsOverlay.graphics.addObjects(from: graphics)
+                // Add the graphics overlay to the map view.
+                self.mapView.graphicsOverlays.add(self.graphicsOverlay)
```
yo1995, line 133: Although the `queryFeatures` method is only called once and there is no side effect to add graphics overlay to the map view here, I still suggest adding the graphics overlay in mapView didSet, as well as to clear its graphics before adding new ones.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+class AddFeaturesContingentValuesViewController: UIViewController {
+    @IBOutlet var mapView: AGSMapView! {
+        didSet {
+            mapView.map = makeMap()
+        }
+    }
+    
+    let geodatabase: AGSGeodatabase!
+    /// The graphics overlay to add the feature to.
+    var graphicsOverlay = AGSGraphicsOverlay()
```
yo1995, line 27:
```suggestion
    let graphicsOverlay = AGSGraphicsOverlay()
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+        // Add the graphic to the graphics overlay.
+        graphicsOverlay.graphics.add(AGSGraphic(geometry: mapPoint, symbol: symbol))
+        // Show the attributes table view.
+        performSegue(withIdentifier: "AddFeature", sender: self)
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Set the map view's touch delegate.
+        mapView.touchDelegate = self
```
yo1995, line 151: Move this to mapView didSet makes better sense.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+        
+        // Add the source code button item to the right of navigation bar.
+        (navigationItem.rightBarButtonItem as! SourceCodeBarButtonItem).filenames = [
```
yo1995, line 154:
```suggestion
        (navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = [
```
It's highly unlikely that we'll change the UI structure of this sample viewer, but using ❓ is safer.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+            "AddFeaturesContingentValuesViewController",
+            "ContingentValuesTableViewController"
+        ]
+    }
+    
+    // MARK: - Navigation
```
yo1995, line 160:
```suggestion
    // MARK: Navigation
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,200 @@
...
+            controller.featureTable = featureTable
+            controller.graphicsOverlay = graphicsOverlay
+            controller.mapPoint = mapPoint
+            controller.delegate = self
+        }
+    }
+}
+
+extension AddFeaturesContingentValuesViewController: ContingentValuesDelegate {
+    func createGraphics(for feature: AGSFeature) -> AGSGraphic {
```
yo1995, line 177:
```suggestion
    func createGraphic(for feature: AGSFeature) -> AGSGraphic {
```
Don't know if this is a correct suggestion haha 😅

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,324 @@
...
+        case bufferSizeCell:
+            tableView.deselectRow(at: indexPath, animated: true)
+            showBufferSizeOptions()
+            toggleBufferSizePickerVisibility()
+        default:
+            return
+        }
+    }
+        
+
```
yo1995, line 285:
```suggestion
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,324 @@
...
+            let graphic = self.delegate?.createGraphics(for: feature)
+            graphicsOverlay?.graphics.add(graphic)
+            self.dismiss(animated: true)
+        }
+    }
+    
+    // MARK: Properties
+    
+    /// The geodatabase's feature table.
+    var featureTable: AGSArcGISFeatureTable?
```
yo1995, line 54: The properties passed along with the segue oughtn't to be optional, as you pass them from the presenting controller purposefully.
same for the ones below.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,324 @@
...
+    
+    @IBAction func doneBarButtonItemTapped(_ sender: Any) {
+        guard let feature = feature, let featureTable = featureTable else { return }
+        // Set the feature's geometry to the map point.
+        feature.geometry = mapPoint
+        // Add the feature to the feature table.
+        featureTable.add(feature) { [unowned self] _ in
+            // Create buffer graphics for the new feature.
+            let graphic = self.delegate?.createGraphics(for: feature)
+            graphicsOverlay?.graphics.add(graphic)
```
yo1995, line 46: A few problems here:

1. An async method is blocking the dismiss UI, which means the done button tap event can halt for indefinite time if add feature operation takes longer.
2. The result of the delegate method is used in this `ContingentValuesTableViewController`, which is not what delegate pattern should be.

It would be better like this:

1. Have a delegate method such as `func contingentValuesTableViewController(_ controller: ContingentValuesTableViewController, didFinishWith feature: AGSFeature)`
2. No need to have `createGraphics` as a protocol stub. When the `didFinishWith` method is called, you can call `createGraphics` in it, in the `AddFeaturesContingentValuesViewController`, which means

```swift
extension AddFeaturesContingentValuesViewController: ContingentValuesDelegate {
    func contingentValuesTableViewController(_ controller: ContingentValuesTableViewController, didFinishWith feature: AGSFeature) {
        let graphic = createGraphics(for: feature)
        // Add the feature to the feature table.
        featureTable?.add(feature) { [weak self] _ in
            self?.graphicsOverlay.graphics.add(graphic)
        }
    }
}
```

3. This means when the done button is tapped, only a contingent-value-constrained feature gets passed back to the `AddFeaturesContingentValuesViewController`. The other work can be done afterwards there.

yo1995 (commented):
I'll give another review when these suggestions are addressed.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,313 @@
...
+    }
+    
+    // MARK: Properties
+    
+    /// The geodatabase's feature table.
+    var featureTable: AGSArcGISFeatureTable!
+    /// The feature to add to the feature table.
+    var feature: AGSArcGISFeature!
+    /// An array of buffer sizes valid for the feature.
+    var bufferSizes: [Int]!
```
yo1995, line 50:
```suggestion
    var bufferSizes = [Int]()
```
to get rid of the ❗

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,313 @@
...
+        // Get the contingent value results using the feature and field.
+        let contingentValueResult = featureTable?.contingentValues(with: feature, field: "BufferSize")
+        guard let bufferSizeGroupContingentValues = contingentValueResult?.contingentValuesByFieldGroup["BufferSizeFieldGroup"] as? [AGSContingentRangeValue] else { return }
+        // Set the minimum and maximum possible buffer sizes.
+        let minValue = bufferSizeGroupContingentValues[0].minValue as! Int
+        let maxValue = bufferSizeGroupContingentValues[0].maxValue as! Int
+        bufferSizes = Array(minValue...maxValue)
+        // Select the buffer size if it is the only option available.
+        if bufferSizes?.count == 1 {
+            selectedBufferSize = bufferSizes?[0]
```
yo1995, line 177:
```suggestion
            selectedBufferSize = bufferSizes[0]
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,313 @@
...
+            // Set the attributes to nil if the protection value has changed.
+            if self.selectedProtection != previouslySelectedProtection, self.selectedBufferSize != nil {
+                self.selectedBufferSize = nil
+            }
+            // Remove the options view controller after selection.
+            self.navigationController?.popViewController(animated: true)
+        }
+        // Set the options view controller's title and present it.
+        optionsViewController.title = "Protection"
+        self.show(optionsViewController, sender: self)
```
yo1995, line 162:
```suggestion
        show(optionsViewController, sender: self)
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,313 @@
...
+    func validateContingency() {
+        guard let featureTable = featureTable, let feature = feature else { return }
+        // Validate the feature's contingencies.
+        let contingencyViolations = featureTable.validateContingencyConstraints(with: feature)
+        if contingencyViolations.isEmpty {
+            // If there are no contingency violations in the array, the feature is valid and ready to add to the feature table.
+            doneBarButtonItem.isEnabled = true
+        } else {
+            // Present an alert if there are contingency violations.
+            self.presentAlert(title: "", message: "Invalid contingent values")
```
yo1995, line 209:
```suggestion
            presentAlert(title: "", message: "Invalid contingent values")
```
It may be too much to add here, but do you think printing out the field group of the contingency is helpful?

yo1995 (commented):
Mostly good, feel free to add Phil afterwards.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
vquach2404: "" Will produce the same result as nil unfortunately.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
vquach2404: Sure, I can add that. I think it'd be helpful for troubleshooting if users enter their own data.

vquach2404:
Just aligning this PR with updates from the design PR. But this is now ready for review!

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+            protectionCell.textLabel?.isEnabled = true
+            protectionCell.isUserInteractionEnabled = true
+        }
+        if selectedProtection == nil {
+            bufferSizeCell.textLabel?.isEnabled = false
+            bufferSizeCell.isUserInteractionEnabled = false
+        } else {
+            bufferSizeCell.textLabel?.isEnabled = true
+            bufferSizeCell.isUserInteractionEnabled = true
+        }
```
philium, line 235: Instead of enabling these based on what "step" the user is on, consider exposing the rows one at a time. So when the view first appears, only the status row is visible. When a status has been selected, the protection row becomes visible. And so on. That would be pretty easy to do in `tableView(_: numberOfRowsInSection:)`.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+        // Get the previously selected status value.
+        let previouslySelectedStatus = selectedStatus
+        // Get the first field by name.
+        let statusField = featureTable?.field(forName: "Status")
+        // Get the field's domains as coded value domain.
+        let codedValueDomain = statusField?.domain as! AGSCodedValueDomain
+        // Get the coded value domain's coded values.
+        let statusCodedValues = codedValueDomain.codedValues
+        // Get the selected index if applicable.
+        let selectedIndex = statusCodedValues.firstIndex { $0.name == selectedStatus?.name }
```
philium, line 114:
```suggestion
        let selectedIndex = statusCodedValues.firstIndex(where: { $0.name == selectedStatus?.name })
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+        // Use the coded value names as labels to show the option table view controller.
+        let optionsViewController = OptionsTableViewController(labels: statusCodedValues.map { $0.name }, selectedIndex: selectedIndex) { [weak self] newIndex in
```
philium, line 116:
```suggestion
        let optionsViewController = OptionsTableViewController(labels: statusCodedValues.map(\.name), selectedIndex: selectedIndex) { [weak self] newIndex in
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+            // Set the attributes to nil if the status value has changed.
+            if self.selectedStatus != previouslySelectedStatus, self.selectedProtection != nil {
+                self.selectedProtection = nil
+                if self.selectedBufferSize != nil {
+                    self.selectedBufferSize = nil
+                }
+            }
+            self.createFeature(with: self.selectedStatus!)
+            // Remove the options view controller after selection.
+            self.navigationController?.popViewController(animated: true)
```
philium, line 129: Based on experience, I've found it helpful to be explicit as to which view controller needs to be popped to. That way you can be sure that the user will return to this one.
```suggestion
            self.navigationController?.popToViewController(self, animated: true)
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+    func showProtectionOptions() {
+        // Get the previously selected protection value.
+        let previouslySelectedProtection = selectedProtection
+        guard let feature = feature else { return }
+        // Get the contingent value results with the feature for the protection field.
+        let contingentValuesResult = featureTable?.contingentValues(with: feature, field: "Protection")
+        // Get contingent coded values by field group.
+        guard let protectionGroupContingentValues = contingentValuesResult?.contingentValuesByFieldGroup["ProtectionFieldGroup"] as? [AGSContingentCodedValue] else { return }
+        // Get the selected index if applicable.
+        let selectedIndex = protectionGroupContingentValues.firstIndex { $0.codedValue.name == selectedProtection?.codedValue.name }
```
philium, line 146:
```suggestion
        let selectedIndex = protectionGroupContingentValues.firstIndex(where: { $0.codedValue.name == selectedProtection?.codedValue.name })
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+        // Use the coded value names as labels to show the option table view controller.
+        let optionsViewController = OptionsTableViewController(labels: protectionGroupContingentValues.map { $0.codedValue.name }, selectedIndex: selectedIndex) { [weak self] newIndex in
```
philium, line 148:
```suggestion
        let optionsViewController = OptionsTableViewController(labels: protectionGroupContingentValues.map(\.codedValue.name), selectedIndex: selectedIndex) { [weak self] newIndex in
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+            guard let self = self else { return }
+            // Set the selected protection value.
+            self.selectedProtection = protectionGroupContingentValues[newIndex]
+            feature.attributes["Protection"] = self.selectedProtection?.codedValue.code
+            // Set the attributes to nil if the protection value has changed.
+            if self.selectedProtection != previouslySelectedProtection, self.selectedBufferSize != nil {
+                self.selectedBufferSize = nil
+            }
+            // Remove the options view controller after selection.
+            self.navigationController?.popViewController(animated: true)
```
philium, line 158:
```suggestion
            self.navigationController?.popToViewController(self, animated: true)
```

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+    /// Get the minimum and maximum values of the possible buffer sizes.
+    func showBufferSizeOptions() {
+        guard let feature = feature else { return }
+        // Get the contingent value results using the feature and field.
+        let contingentValueResult = featureTable?.contingentValues(with: feature, field: "BufferSize")
+        guard let bufferSizeGroupContingentValues = contingentValueResult?.contingentValuesByFieldGroup["BufferSizeFieldGroup"] as? [AGSContingentRangeValue] else { return }
+        // Set the minimum and maximum possible buffer sizes.
+        let minValue = bufferSizeGroupContingentValues[0].minValue as! Int
+        let maxValue = bufferSizeGroupContingentValues[0].maxValue as! Int
+        bufferSizes = Array(minValue...maxValue)
```
philium, line 174: You might consider changing the type of `bufferSizes` to `ClosedRange<Int>`.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+    
+    func pickerView(_ pickerView: UIPickerView, numberOfRowsInComponent component: Int) -> Int {
+        return bufferSizes.count
+    }
+}
+
+extension ContingentValuesTableViewController: UIPickerViewDelegate {
+    func pickerView(_ pickerView: UIPickerView, titleForRow row: Int, forComponent component: Int) -> String? {
+        let bufferSizeTitle = String(bufferSizes[row])
+        return bufferSizeTitle
```
philium, line 297: Ideally a number formatter would be used to format an integer as a string so that the number is localized. That might be a good thing to model (using a number formatter).

philium (commented):
I haven't reviewed everything yet, but here is some preliminary feedback.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+            try? FileManager.default.removeItem(at: temporaryGeodatabaseURL)
+        }
+    }
+    
+    /// Called in response to the geodatabase load operation completing.
+    func geodatabaseDidLoad(with result: Result<Void, Error>) {
+        switch result {
+        case .success:
+            // Get and load the first feature table in the geodatabase.
+            let featureTable = geodatabase.geodatabaseFeatureTables[0] as AGSArcGISFeatureTable
```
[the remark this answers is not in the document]
vquach2404: This cast is needed in order to access the `contingentValuesDefinitionProperty`. See more info on the [doc](https://developers.arcgis.com/ios/api-reference/interface_a_g_s_arc_g_i_s_feature_table.html#a4355a887684421547efeb9adeee74eac)

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
yo1995: I was also thinking about making this suggestion, but wasn't sure if a `ClosedRange` can be the data source of a picker.

vquach2404: I don't think I'm able to since I can't return the value using an index from `ClosedRange`

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
vquach2404: I've updated this to use a number formatter but it is slightly more complicated since I have to convert the Int into an NSNumber first.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/ContingentValuesTableViewController.swift
vquach2404: Hm, I'm trying it out now but I finding that it may cause confusion, having to use "magic numbers" to determine how many rows to delete. Also, I'd have to reload the table view data after each selection. I'd like to keep it the way it is now. What're your thoughts?

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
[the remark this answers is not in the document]
philium: As far as I can tell, that property isn't being accessed anywhere and everything compiles without the cast. Am I missing something?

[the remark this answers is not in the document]
vquach2404: Hm, we were under the impression that it needed to be casted as part of the contingent values workflow. I'll check with the Serenity team to confirm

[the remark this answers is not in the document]
vquach2404: Ah Rama said only `AGSArcGISFeatureTables` support contingent values but `AGSGeodatabaseFeatureTable` inherits that so it is also valid. I've removed the cast!

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,212 @@
...
+            // Create a temporary directory.
+            let temporaryDirectoryURL = try FileManager.default.url(for: .itemReplacementDirectory, in: .userDomainMask, appropriateFor: geodatabaseURL, create: true)
+            // Create a temporary URL where the geodatabase URL can be copied to.
+            let temporaryGeodatabaseURL = temporaryDirectoryURL.appendingPathComponent("birdsNestGDB", isDirectory: false).appendingPathExtension("geodatabase")
+            // Copy the item to the temporary URL.
+            try FileManager.default.copyItem(at: geodatabaseURL, to: temporaryGeodatabaseURL)
+            // Create the geodatabase with the URL.
+            geodatabase = AGSGeodatabase(fileURL: temporaryGeodatabaseURL)
+        } catch {
+            print("Error setting up geodatabase: \(error)")
```
yo1995, line 49: No need to change, can also be a fatalError.

arcgis-ios-sdk-samples/Edit data/Add features with contingent values/AddFeaturesContingentValuesViewController.swift
```diff
@@ -0,0 +1,212 @@
...
+                let newGraphics = result.featureEnumerator().allObjects.map(self.makeGraphic(feature:))
+                // Replace existing graphics with the new graphics.
+                self.graphicsOverlay.graphics.setArray(newGraphics)
+            }
+        }
+    }
+    
+    /// Add a single feature to the map.
+    func addFeature(at mapPoint: AGSPoint) {
+        // Create a symbol to represent a bird's nest.
```
yo1995, line 140: Might be worth checking if the mapPoint is within the map package's extent. Currently I can add features outside the basemap visible area.

vquach2404: Good idea. How can I check?

yo1995: Sth like below, but you might need to improvise as I've never done this before 😅 
```swift
guard let mapExtent = fillmoreVectorTiledLayer.fullExtent, AGSGeometryEngine.geometry(mapPoint, within: mapExtent) else { return }
```
I'm OK without this change.

yo1995 (dismissed):

yo1995 (approved):

vquach2404:
Thanks everyone for the reviews!

PR #1146, opened 2022-03-02
[Review Only] Google Analytics changes

yo1995:
## Description

This is a review only PR - **we'll not merge this branch** to `v.next` or `main`.

This branch contains all the changes except the API key file `GoogleService-Info.plist`, so that we can review the code changes without leaking secrets.

## Linked Issue(s)

- `common-samples/issues/3282`
- Epic: `common-samples/issues/535`

## How To Test

1. If you are in the GA account group, you'll be able to see the analytics dashboard
2. When the GA API key file is applied, events can be logged into GA
3. It takes up to 2 days for the data to aggregate and display on the dashboard

## Screenshots

The telemetry design at this stage

image

Some test data as below

data_Page_1

## To Discuss

1. How shall we stage this branch in the future? Locally or in private repo
2. GDPR compliance and future tooling

### DIFF
--- a/arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
+++ b/arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
@@ -100,6 +100,7 @@
 		00ADC3D12464DD3600A3B88D /* IdentifyRasterCellViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 00ADC3D02464DD3600A3B88D /* IdentifyRasterCellViewController.swift */; };
 		00ADC3D22464DDA200A3B88D /* IdentifyRasterCellViewController.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = 00ADC3D02464DD3600A3B88D /* IdentifyRasterCellViewController.swift */; };
 		00BA068525311B0D0039EF4D /* hydrography in Resources */ = {isa = PBXBuildFile; fileRef = 00BA068325311B0D0039EF4D /* hydrography */; settings = {ASSET_TAGS = (Hydrography, ); }; };
+		00BE7B38275AEFD60012CE95 /* FirebaseAnalyticsWithoutAdIdSupport in Frameworks */ = {isa = PBXBuildFile; productRef = 00BE7B37275AEFD60012CE95 /* FirebaseAnalyticsWithoutAdIdSupport */; };
 		00BE7B3E2764274C0012CE95 /* BrowseBuildingFloors.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 00BE7B3D2764274C0012CE95 /* BrowseBuildingFloors.storyboard */; };
 		00BE7B402764275D0012CE95 /* BrowseBuildingFloorsViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 00BE7B3F2764275D0012CE95 /* BrowseBuildingFloorsViewController.swift */; };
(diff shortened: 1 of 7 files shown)


arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -91,6 +95,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
         }
     }
     
+    // MARK: - Google Analytics settings
+    
+    func setGoogleAnalyticsPref() {
+        // Enable/disable GA based on settings.
+        let disableGoogleAnalytics = UserDefaults.standard.bool(forKey: "disableGoogleAnalytics")
```
yo1995, line 102: This is due to an iOS limitation, see: https://stackoverflow.com/questions/36253998

arcgis-ios-sdk-samples/Content Display Logic/Settings.bundle/Root.plist
```diff
@@ -22,6 +22,16 @@
 			<key>DefaultValue</key>
 			<false/>
 		</dict>
+		<dict>
+			<key>Type</key>
+			<string>PSToggleSwitchSpecifier</string>
+			<key>Title</key>
+			<string>Disable Google Analytics</string>
```
yo1995, line 29: Same reason as above - have to have a default value as `false`.

yo1995:
@TADraeseke We'll use this PR as a preview - API keys will be added when we release with GA.

arcgis-ios-sdk-samples/AppDelegate.swift
philium: This isn't an iOS limitation but rather a misunderstanding of the User Defaults API. It has been a while since I've used it, but if I remember quickly, default values must be registered using `UserDefaults.register(defaults:)`. Also, be aware that the setting can change while the app is active, so the setting should be observed for changes as well. In fact, using the `initial` option, the observation itself can be what enables and disables analytics exclusively.

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -5897,6 +5907,14 @@
...
 			};
 		};
+		00BE7B36275AEFD60012CE95 /* XCRemoteSwiftPackageReference "firebase-ios-sdk" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/firebase/firebase-ios-sdk";
+			requirement = {
+				kind = upToNextMajorVersion;
+				minimumVersion = 8.10.0;
+			};
+		};
```
philium, line 5917: I've never used Firebase before. What impact does adding this dependency have on app size (how large is the Firebase framework)? Also, will we need to change the App Privacy section of our App Store page?

For what it is worth, I generally am not in favor of adding third part dependencies unless there is an extremely compelling reason to do so. Each dependency, especially one like this that executes code even before that app is useable by the user, is a failure point. Remember a while back when a bunch of prominent apps stopped working for a while because they used a popular third party analytics framework that had a problem? That is the power that dependencies like this hold. In my opinion, it isn't worth it. But that's up to you and the samples team to decide.

yo1995: - app size
The impact is quite minimum. The previous version (100.13.0.1) of sample viewer app is ~85 MB, and the current version (100.13.0.2) is 87 MB, according to App Store on iPhone 11.
- App Privacy section
I've included usage data (select sample, category, and source code view) and search history (search bar) policies, prior to the release.
- third party dependencies
I agree. I also tried to avoid using 3rd party dependencies in my personal project.  
We researched a few options: Google Analytics, Amazon Pinpoint, Adobe Analytics, and building an AWS based in house solution. Eventually Trevor decided GA is the easiest way to go for now, as we wanted to use it as an experiment to see what data we can get. In the future we may decide to move to other solutions or remove GA dependency.

PR #1149, opened 2022-03-15
[Logistics] Uncomment PR template content

yo1995:
## Description

This PR updates the PR template to make it easier to use.

Had a discussion with Viv and we agreed that there is no need to keep the sections commented out. Given that we need to remove unused section from the template anyway, it is easier to just leave them as visible text so that there is no need to first delete the comment strings, then reindent before starting to write.


vquach2404 (approved):
Thanks for making this change!

philium (approved):

PR #1148, opened 2022-03-15
[Update] Replace instances of `AGSServiceFeatureTable.applyEdits` , 1st PR

yo1995:
## Description

This PR replaces various usage of `applyEdits` method on `AGSServiceFeatureTable` to `AGSServiceGeodatabase`, to prevent data inconsistency issue.

I'll break down the Epic into 2 PRs to make it easier to review. This is the 1st PR.

All the behaviors should remain the same. Coding style has been updated/refactored to follow our latest style guide. Also, a few UI related bugs are fixed. But we'll still request Sarat's review to ensure nothing is breaking the tests.

Also, all comment strings are formatted.

Finally, READMEs will be updated in yet another separate PR, after we finish updating the common designs.

## Linked Issue(s)

`common-samples/issues/3371`

## How To Test

Go though all changed samples and see if they work as expected.

<details><summary>Example change</summary>

```swift
/// Load and set a service geodatabase from a feature service URL.
/// - Parameter serviceURL: The URL to the feature service.
func loadServiceGeodatabase(from serviceURL: URL) {
    let serviceGeodatabase = AGSServiceGeodatabase(url: serviceURL)
    serviceGeodatabase.load { [weak self] error in
        guard let self = self else { return }
        if let error = error {
            self.presentAlert(error: error)
        } else {
            let featureTable = serviceGeodatabase.table(withLayerID: 0)!
            self.featureTable = featureTable
            self.serviceGeodatabase = serviceGeodatabase
            // Add the feature layer to the operational layers on map.
            let featureLayer = AGSFeatureLayer(featureTable: featureTable)
            self.mapView.map?.operationalLayers.add(featureLayer)
        }
    }
}

// then...
serviceGeodatabase.applyEdits { [weak self] (featureTableEditResults: [AGSFeatureTableEditResult]?, error: Error?) in
    if let error = error {
        self?.presentAlert(message: "Error while applying edits:\(error.localizedDescription)")
    } else {
        if let featureTableEditResults = featureTableEditResults,
           featureTableEditResults.first?.editResults.first?.completedWithErrors == false {
            self?.presentAlert(message: "Edits applied successfully")
        }
    }
}
```

</details>

PR #1151, opened 2022-03-17
[Update] Replace instances of AGSServiceFeatureTable.applyEdits , 2nd PR

yo1995:
## Description

This is a follow up to #1148 

This PR intended to update the following samples, however due to various reasons, some samples cannot adopt the new `applyEdits` on service geodatabase approach.

- [x] Edit features connected
- [ ] Edit feature attachments: currently blocked due to an attachment globalID issue, see issues/3012#issuecomment-3335519 . The commit is stashed at f8116f169821fed37ee65d50fed3cdabdb625441 and can be re-introduced once the service config is changed
- [ ] Offline edit and sync: this is an offline workflow, which means no changes need to be made
- [ ] Collect data AR: this also suffers from the attachment issue as `Edit feature attachments` sample above 
- [ ] Add delete related features: after making the change in this commit d4309cc0e8c5d94249dbfaf9d2527f5740a14e9f, the sample doesn't work correctly, thus not making the change. More specifically, the data is applied to remote database, but query and delete functions are not correct.

More to do

- Update the design docs
- Update the README for all samples

### DIFF
--- a/arcgis-ios-sdk-samples/Edit data/Edit features (connected)/EditFeaturesOnlineViewController.swift
+++ b/arcgis-ios-sdk-samples/Edit data/Edit features (connected)/EditFeaturesOnlineViewController.swift
@@ -45,8 +45,6 @@ class EditFeaturesOnlineViewController: UIViewController, AGSGeoViewTouchDelegat
         super.viewDidLoad()
         // Add the source code button item to the right of navigation bar.
         (navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["EditFeaturesOnlineViewController", "FeatureTemplatePickerViewController"]
-        // Hide the sketchToolbar initially.
-        sketchToolbar.isHidden = true
         // Load the service geodatabase.
         let damageFeatureService = URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/DamageAssessment/FeatureServer")!
         loadServiceGeodatabase(from: damageFeatureService)
--- a/arcgis-ios-sdk-samples/Edit data/Edit features (connected)/EditFeaturesOnlineViewController.swift
+++ b/arcgis-ios-sdk-samples/Edit data/Edit features (connected)/EditFeaturesOnlineViewController.swift
@@ -16,58 +16,79 @@ import UIKit
 import ArcGIS
 
 class EditFeaturesOnlineViewController: UIViewController, AGSGeoViewTouchDelegate, AGSPopupsViewControllerDelegate, FeatureTemplatePickerDelegate {
-    @IBOutlet private weak var mapView: AGSMapView!
-    @IBOutlet private weak var sketchToolbar: UIToolbar!
-    @IBOutlet private weak var doneBBI: UIBarButtonItem!
+    @IBOutlet var mapView: AGSMapView! {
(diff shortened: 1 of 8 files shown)


vquach2404 (approved):
Awesome job! Since you updated the comment styling in the main view controller file, can. you update the other file as well?

vquach2404:
Also, we may considering renewing the the workflow of this sample. Rather than tapping the the toolbar button to add a feature, perhaps tapping on the map the prompt the user to add a new feature. I think this workflow is a bit more intuitive and would save the user a few taps. We can log a separate issue for this suggestion since it's out of the scope for this PR.

arcgis-ios-sdk-samples/Edit data/Edit features (connected)/EditFeaturesOnlineViewController.swift
```diff
@@ -16,58 +16,79 @@ import UIKit
...
-        
-        (featureLayer.featureTable as! AGSServiceFeatureTable).applyEdits { [weak self] (_, error) in
-            UIApplication.shared.hideProgressHUD()
-            
-            if let error = error {
-                self?.presentAlert(message: "Error while applying edits: \(error.localizedDescription)")
-            } else {
-                self?.presentAlert(message: "Edits applied successfully!")
+        if serviceGeodatabase.hasLocalEdits() {
+            serviceGeodatabase.applyEdits { [weak self] (featureTableEditResults: [AGSFeatureTableEditResult]?, error: Error?) in
```
vquach2404, line 84: Does the README need to be updated since we are now applying edits on the service geodatabase?

yo1995: Yes, it will be done in a 3rd PR. I'll first update the designs internally before making changes to public PRs. Internal number: `common-samples/pull/3380`

arcgis-ios-sdk-samples/Edit data/Edit features (connected)/EditFeaturesOnlineViewController.swift
```diff
@@ -16,58 +16,79 @@ import UIKit
...
-        
-        // hide the sketchToolbar initially
-        self.sketchToolbar.isHidden = true
-        
-        // store the feature layer for later use
-        self.featureLayer = featureLayer
+        // Add the source code button item to the right of navigation bar.
+        (navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["EditFeaturesOnlineViewController", "FeatureTemplatePickerViewController"]
+        // Hide the sketchToolbar initially.
+        sketchToolbar.isHidden = true
```
philium, line 49: Could this be set in the storyboard?

yo1995:
> Awesome job! Since you updated the comment styling in the main view controller file, can you update the other file as well?

I typically tries to avoid adding changes to a PR if they are in a separate file that is unrelated to the PR. 😅  

> Also, we may considering renewing the the workflow of this sample.

We can fix the comments in that PR.

philium (approved):

vquach2404 (approved):

saratchandrakarumuri (approved):
Sample is worked as expected

PR #1153, opened 2022-03-24
New: Map image layer tables

vquach2404:
## Description

This PR implements `Map image layer tables` in `Layers` category.
Branch: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Viv/manageMapImage

## Linked Issue(s)

- `common-samples/issues/3262`

## How To Test

Tap on bottom toolbar and select some of the available comments. Notice the first result will be cyan.


arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -3297,6 +3303,7 @@
 				4CAC91D72242AAB700663A10 /* Load WFS with XML query */,
 				12D6602B226697040005DE8A /* Group layers */,
 				4CE15D4822F37968005F2968 /* Control annotation sublayer visibility */,
+				F1EF67DB27E2AE200049FA9A /* Map image layer tables */,
```
yo1995, line 3306: We tend to put latest samples at the top of each folder. Until we make the sort-alphabetically change, let's still keep that.

arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
```diff
@@ -312,7 +312,7 @@
 				<key>displayName</key>
 				<string>Show device location using indoor positioning</string>
 				<key>descriptionText</key>
-				<string>Show your device's real-time location while inside a building by using signals from indoor positioning beacons.</string>
+				<string>Show your device&apos;s real-time location while inside a building by using signals from indoor positioning beacons.</string>
```
yo1995, line 315: Though this change is not expected, I don't have problem keeping it in this PR.

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTables.storyboard
```diff
@@ -0,0 +1,64 @@
...
+                        <outlet property="queryBarButtonItem" destination="NZ4-Ky-Fcl" id="jh4-Ma-ixn"/>
+                    </connections>
+                </viewController>
+                <placeholder placeholderIdentifier="IBFirstResponder" id="Ief-a0-LHa" userLabel="First Responder" customClass="UIResponder" sceneMemberID="firstResponder"/>
+            </objects>
+            <point key="canvasLocation" x="106" y="42"/>
+        </scene>
+    </scenes>
+    <resources>
+        <systemColor name="systemBackgroundColor">
```
yo1995, line 60: Should be tiertiarySystemBackgroundColor for root view, default (no color) for map view.

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+                // Present an error message is the related feature is not found.
+                self.presentAlert(title: "Related feature not found. No Feature", message: nil)
+            }
+        }
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Add the source code button item to the right of navigation bar.
+        (self.navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = [ "MapImageLayerTablesViewController"]
```
yo1995, line 148:
```suggestion
        (self.navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = ["MapImageLayerTablesViewController"]
```

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+            // Set the viewpoint.
+            mapView.setViewpoint(AGSViewpoint(latitude: 41.734152, longitude: -88.163718, scale: 2e5))
+            // Create a graphics overlay to show selected features and add it to the map view.
+            mapView.graphicsOverlays.add(selectedFeaturesOverlay)
+        }
+    }
+    
+    @IBOutlet var queryBarButtonItem: UIBarButtonItem!
+    
+    @IBAction func queryFeaturesActions() {
```
yo1995, line 32: Let's move this method below the properties to keep it clean looking.

Also, the signature can be more standardized to reflect the target-action pattern, i.e. 

```swift
@IBAction func queryFeaturesActions(_ sender: UIBarButtonItem)
```

Use the right click -> Refactor -> Rename to avoid detaching from storyboard.

Method name can also be queryButtonTapped

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+        // Create an action sheet to display the various comments to choose from.
+        let alertController = UIAlertController(title: "Related Service Requests", message: "Select a comment to view related spatial features on the map.", preferredStyle: .actionSheet)
+        // Create an action for each comment.
+        commentsArray.forEach { feature in
+            // Extract the "comments" attribute as a string.
+            let commentsTitle = feature.attributes["comments"] as! String
+            // Create an action with the comments title.
+            let action = UIAlertAction(title: commentsTitle, style: .default) { [weak self] (_) in
```
yo1995, line 40: It's OK to use weak self; unowned self is also OK because the async UI always returns when the view controller exists.

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+                guard let self = self else { return }
+                // Clear the former graphics.
+                self.selectedFeaturesOverlay.graphics.removeAllObjects()
+                // Get the map image layer that contains the service request sublayer and the service request comments table.
+                let serviceRequestsMapImageLayer = self.mapView.map?.operationalLayers.firstObject as? AGSArcGISMapImageLayer
```
yo1995, line 45: It may be worth to hold this layer as a instance property when it first loads.

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+                // Get the (non-spatial) table that contains the service request comments.
+                guard let commentsTable = serviceRequestsMapImageLayer?.tables.first else { return }
```
yo1995, line 47: Or just hold this table as the map image layer is not essential.

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+                // Get the relationship that defines related service request features for features in the comments table (this is the first and only relationship).
+                let commentsRelationshipInfo = commentsTable.layerInfo?.relationshipInfos.first
+                // Create query parameters to get the related service request for features in the comments table.
+                let relatedQueryParameters = AGSRelatedQueryParameters(relationshipInfo: commentsRelationshipInfo!)
+                relatedQueryParameters.returnGeometry = true
```
yo1995, line 52: Hmm, now I wonder if the `relatedQueryParameters` is always the same. How about holding both the table and the parameter as an instance property? Or only hold the table, and construct the parameters in `queryCommentsTable` method.

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+        let alertController = UIAlertController(title: "Related Service Requests", message: "Select a comment to view related spatial features on the map.", preferredStyle: .actionSheet)
+        // Create an action for each comment.
+        commentsArray.forEach { feature in
+            // Extract the "comments" attribute as a string.
+            let commentsTitle = feature.attributes["comments"] as! String
+            // Create an action with the comments title.
+            let action = UIAlertAction(title: commentsTitle, style: .default) { [weak self] (_) in
+                guard let self = self else { return }
+                // Clear the former graphics.
+                self.selectedFeaturesOverlay.graphics.removeAllObjects()
```
yo1995, line 43: This can be moved outside the closure to the top of this method, so whenever the bar button is tapped, the graphics are cleared.

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+        alertController.popoverPresentationController?.barButtonItem = queryBarButtonItem
+        present(alertController, animated: true)
+    }
+    
+    /// The service request URL.
+    static let serviceRequestURL = URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/ServiceRequest/MapServer")!
+    /// The map image layer that uses the service URL.
+    let serviceRequestsMapImageLayer = AGSArcGISMapImageLayer(url: serviceRequestURL)
+    /// The array to store the possible comments.
+    var commentsArray = [AGSFeature]()
```
yo1995, line 73: I believe Phil once suggested against this style.
```suggestion
    var commentsArray: [AGSFeature] = []
```

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+    
+    func makeMap() -> AGSMap {
+        // Create a map with the ArcGIS streets basemap style.
+        let map = AGSMap(basemapStyle: .arcGISStreets)
+        // Load the map image layer's tables and layers.
+        serviceRequestsMapImageLayer.loadTablesAndLayers { _ in
+            // Add the layer to the map.
+            map.operationalLayers.add(self.serviceRequestsMapImageLayer)
+            self.queryFeatures()
+        }
```
yo1995, line 85:
```suggestion
        // Add the layer to the map.
        map.operationalLayers.add(serviceRequestsMapImageLayer)
        // Load the map image layer's tables and layers.
        serviceRequestsMapImageLayer.loadTablesAndLayers { [weak self] _ in
            self?.queryFeatures()
        }
```

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+        return map
+    }
+    
+    /// Query features on the first table in the map image layer.
+    func queryFeatures() {
+        // Create query parameters and set its where clause.
+        let nullCommentsParameters = AGSQueryParameters()
+        nullCommentsParameters.whereClause = "requestid <> '' AND comments <> ''"
+        // Get the first table from the map image layer.
+        guard let commentsTable = serviceRequestsMapImageLayer.tables.first else { return }
```
yo1995, line 95: Yeah, this makes me believe it's worth to hold the table as an instance property. (and construct different parameters in query methods respectively.)

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+        // Query features on the feature table with the query parameters and all feature fields.
+        commentsTable.queryFeatures(with: nullCommentsParameters, queryFeatureFields: .loadAll) { [weak self] result, error in
+            guard let self = self else { return }
+            if let error = error {
```
yo1995, line 99: Phil likes result first error second pattern. ref https://github.com/Esri/arcgis-runtime-samples-ios/pull/970#discussion_r492305608

which makes it 

```swift
            if let comments = result?.featureEnumerator().allObjects {
                // Show the records from the service request comments table in the list view control.
                self.commentsArray = comments
                // Enable the button after the map and features have been loaded.
                self.queryBarButtonItem.isEnabled = true
            } else if let error = error {
            }
```

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+                self.queryBarButtonItem.isEnabled = true
+            }
+        }
+    }
+    
+    /// Query related features for the selected feature.
+    func queryCommentsTable(feature: AGSFeature, commentsTable: AGSServiceFeatureTable, relatedQueryParameters: AGSRelatedQueryParameters) {
+        // Get the feature as an ArcGIS feature.
+        guard let selectedComment = feature as? AGSArcGISFeature else { return }
+        // Query related features for the selected comment and its related query parameters.
```
yo1995, line 114: Same as comment above, maybe consider constructing the query parameter from the commentsTable in this method.

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+        commentsTable.queryRelatedFeatures(for: selectedComment, parameters: relatedQueryParameters) { results, error in
+            let relatedResult = results?.first
```
yo1995, line 116: Why not just conditionally unwrap with relatedFeature?

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+            if let relatedFeature = relatedResult?.featureEnumerator().nextObject() as? AGSArcGISFeature {
+                // Load the feature and get its geometry to show as a graphic on the map.
+                relatedFeature.load { error in
+                    if let error = error {
+                        self.presentAlert(error: error)
+                    } else {
+                        // Get the feature's geometry.
+                        guard let serviceRequestPoint = relatedFeature.geometry as? AGSPoint else { return }
+                        // Create a symbol for the feature.
+                        let selectedRequestSymbol = AGSSimpleMarkerSymbol(style: .circle, color: .cyan, size: 14)
```
yo1995, line 127: This can be set as a simple renderer, to avoid the hassle of constructing a symbol each time.

arcgis-ios-sdk-samples/Layers/Map image layer tables/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## How to use the sample
+
+Tap the toolbar button to prompt a list of comment data from non-spatial features. Tap on one of the comments to query related spatial features and display the first result on the map.
+
+## How it works
+
+1. Create an `AGSArcGISMapImageLayer` with the URL of a map image service.
+2. Load the tables and layers using `loadTablesAndLayers(completion:)` and get one its first table.
+3. To query the table, create`AGSQueryParameters`. Set its `whereClause` to filter the request features.
```
yo1995, line 19:
```suggestion
3. To query the table, create `AGSQueryParameters`. Set its `whereClause` to filter the request features.
```

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
vquach2404: I decided against that so the graphic would still appear in case the user selected the "cancel" option

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
vquach2404: I don't really understand your suggestion. `relatedFeature` derives from the first object in `results` and both are already conditionally unwrapped

yo1995: Sorry I didn't make myself clear. I mean `if let relatedFeature = results?.first?.featureEnumerator().nextObject() as? AGSArcGISFeature`

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+            }
+        }
+    }
+    
+    /// Query related features for the selected feature.
+    func queryCommentsTable(feature: AGSArcGISFeature) {
+        // Get the relationship that defines related service request features for features in the comments table (this is the first and only relationship).
+        let commentsRelationshipInfo = commentsTable?.layerInfo?.relationshipInfos.first
+        // Create query parameters to get the related service request for features in the comments table.
+        let relatedQueryParameters = AGSRelatedQueryParameters(relationshipInfo: commentsRelationshipInfo!)
```
yo1995, line 122:
```suggestion
        // Get the relationship that defines related service request features for features in the comments table (this is the first and only relationship).
        guard let commentsRelationshipInfo = commentsTable?.layerInfo?.relationshipInfos.first else { return }
        // Create query parameters to get the related service request for features in the comments table.
        let relatedQueryParameters = AGSRelatedQueryParameters(relationshipInfo: commentsRelationshipInfo)
```
Consider guard it so that it doesn't need to be force unwrapped on next line.

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+        relatedQueryParameters.returnGeometry = true
+        // Query related features for the selected comment and its related query parameters.
+        commentsTable?.queryRelatedFeatures(for: feature, parameters: relatedQueryParameters) { [weak self] results, error in
+            guard let self = self else { return }
+            // Get the first related feature.
+            if let relatedFeature = results?.first?.featureEnumerator().nextObject() as? AGSArcGISFeature {
+                // Load the feature and get its geometry to show as a graphic on the map.
+                relatedFeature.load { error in
```
yo1995, line 130: This still needs to be weakly captured again in `relatedFeature.load(completion:)`.

FYI: https://stackoverflow.com/questions/41991467/where-does-the-weak-self-go/41992442#41992442

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+                    if let error = error {
+                        self.presentAlert(error: error)
+                    } else {
+                        // Get the feature's geometry.
+                        guard let serviceRequestPoint = relatedFeature.geometry as? AGSPoint else { return }
+                        // Create a graphic to add to the graphics overlay.
+                        let graphic = AGSGraphic()
+                        graphic.geometry = serviceRequestPoint
```
yo1995, line 138:
```suggestion
                        let graphic = AGSGraphic(geometry: serviceRequestPoint, symbol: nil)
```

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+            let commentsTitle = feature.attributes["comments"] as! String
+            // Create an action with the comments title.
+            let action = UIAlertAction(title: commentsTitle, style: .default) { [weak self] (_) in
+                guard let self = self else { return }
+                // Clear the former graphics.
+                self.selectedFeaturesOverlay.graphics.removeAllObjects()
+                // Disable the query button while the feature loads.
+                self.queryBarButtonItem.isEnabled = false
+                // Cast the selected feature as an AGSArcGISFeature.
+                guard let selectedFeature = feature as? AGSArcGISFeature else { return }
```
yo1995, line 59: IMO the `selectedFeature` is always an `AGSArcGISFeature`, as `commentsArray` is guaranteed to be an array of `AGSArcGISFeature`. So you can force unwrap here. 🤔

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+                relatedFeature.load { error in
+                    if let error = error {
+                        self.presentAlert(error: error)
+                    } else {
+                        // Get the feature's geometry.
+                        guard let serviceRequestPoint = relatedFeature.geometry as? AGSPoint else { return }
+                        // Create a graphic to add to the graphics overlay.
+                        let graphic = AGSGraphic()
+                        graphic.geometry = serviceRequestPoint
+                        self.selectedFeaturesOverlay.graphics.add(graphic)
```
yo1995, line 139: This workflow actually makes me wonder: is there a reason why we use `MapServer` over `FeatureServer` for this sample? Because the feature service https://sampleserver6.arcgisonline.com/arcgis/rest/services/ServiceRequest/FeatureServer is also present. Using feature service/feature layer makes it easier to show the select feature UI, by simply calling `AGSFeatureLayer.select(_:)`.

Maybe it is by design, so I don't have problem if we keep the current workflow.

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
vquach2404: `commentsArray` is actually type `[AGSFeature]`. Do you still think I should force unwrap?

yo1995: I'm not too sure. I'll need to talk to others to understand the difference between `AGSFeature` and `AGSArcGISFeature`. 🧐 
I made this suggestion because all options in the action sheet work, so the cast always succeed in this sample... 😅

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
vquach2404: I'm also unsure about this. I'll bring this topic to the common design PR.

yo1995: Nevermind, I tried to use the feature service and it doesn't seem to work under the current workflow, showing duplicate comments.

dup

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
vquach2404: >  If however, the outer closure has already weakly captured self, then the inner closure will already have a weak reference to self, thus adding [weak self] to the inner closure will have no effect.

The response says that adding `[weak self]` to the inner closure will have no effect. Am I understanding correctly?

yo1995: Haha it is a confusion brought by reusing the name `self`. If you change the left `self` on line 126 to `strongSelf`, you'll see what I mean 😏

yo1995 (approved):

arcgis-ios-sdk-samples/Layers/Map image layer tables/MapImageLayerTablesViewController.swift
yo1995: Today I had a brief chat with Divesh and he pointed out a few things, and I read the [doc](https://developers.arcgis.com/ios/api-reference/interface_a_g_s_arc_g_i_s_feature.html#:~:text=An%20AGSArcGISFeature%20object,entity%20it%20represents) again and believe that casting to `AGSArcGISFeature` should be fine for a map service.

arcgis-ios-sdk-samples/Layers/Query related features from non-spatial table/QueryRelatedFeaturesNonSpatialTableViewController.swift
```diff
@@ -0,0 +1,157 @@
...
+                        self.selectedFeaturesOverlay.graphics.add(graphic)
+                        // Set the viewpoint to the the related feature.
+                        self.mapView.setViewpointCenter(serviceRequestPoint, scale: 150_000)
+                        // Enable the button after the feature has finished loading.
+                        self.queryBarButtonItem.isEnabled = true
+                    }
+                }
+            } else {
+                // Present an error message is the related feature is not found.
+                self.presentAlert(title: "Related feature not found. No Feature", message: nil)
```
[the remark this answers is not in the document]
vquach2404: This contradicts Ting's [suggestion](https://github.com/Esri/arcgis-runtime-samples-ios/pull/1153#discussion_r836817254). @yo1995 I'm a little confused, can you elaborate?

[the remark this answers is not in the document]
yo1995: Oh my bad, I didn't notice on line 147 `self` is used to present alert. I thought the first completion block after `queryRelatedFeatures` didn't use self.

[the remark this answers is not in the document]
philium: The inner completion handler will capture `self` however it is in the current context. If it is weak, then it will capture it as weak. So as long as the outer completion handler captures it as weak and doesn't shadow `self` with a strong reference (`guard let self = self else { return }`), then the inner completion handler doesn't need `[weak self]`. Does that make sense?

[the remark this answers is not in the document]
vquach2404: Yes, that all clears things up. Thank you both!

yo1995 (dismissed):

arcgis-ios-sdk-samples/Layers/Query related features from non-spatial table/QueryRelatedFeaturesNonSpatialTableViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+        map.operationalLayers.add(serviceRequestsMapImageLayer)
+        // Load the map image layer's tables and layers.
+        serviceRequestsMapImageLayer.loadTablesAndLayers { [weak self] _ in
+            self?.queryFeatures()
+        }
+        return map
+    }
+    
+    /// Make a renderer for the graphics overlay.
+    func makeRenderer () -> AGSSimpleRenderer {
```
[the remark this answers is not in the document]
philium: This method is no longer being used, right? If so, please just remove it.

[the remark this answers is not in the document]
yo1995: True!! 🤦

philium (approved):

yo1995 (approved):

saratchandrakarumuri (approved):
New sample is working as expected

vquach2404:
Thanks everyone!

PR #1134, opened 2022-01-13
Display feature layers

vquach2404:
This sample replaces 4 feature layer samples (feature service, geodatabase, shapefile, geopackage). 

    This PR implements `Display feature layers` in `feature` category.
    Branch: [URL_TO_BRANCH](https://github.com/Esri/arcgis-runtime-samples-ios/tree/Viv/featureLayerSamples)

## Linked Issue(s)

`common-samples/issues/3245`

## How To Test

Use the bottom toolbar to view different feature layers from various sources.

## To Discuss

During review, someone suggested that this sample be moved to the top of the list since it showcases foundational workflows. This also led us to discuss reorganizing the order of samples within their categories. So perhaps we can reorder this sample in another PR if we decide to reorder the samples.

PR #1156, opened 2022-04-11
[Fix] ODR tags and redundant shapefile removal

yo1995:
## Description

This PR fixes the issue brought by #1134 

- It used Aurora, CO GeoPackage and LA Trails Geodatabase, but didn't include them in the dependency plist entry
- It removed the sample `Feature layer (shapefile)` that uses `Public_Art.shp`, but didn't remove the resource files which are now unused.

Currently `Display feature layers` sample crashes in the App Store version (but not the local build) because both resources are not downloaded as ODR when the sample loads.

## Linked Issue(s)

- `common-samples/issues/3425`
- `common-samples/issues/3407`

## How To Test

After this PR is merged and subsequently uploaded to App Store Connect for release or TestFlight, we shall verify that `Display feature layers` is working as expected.


philium (approved):

vquach2404 (approved):

saratchandrakarumuri (approved):
Changes look good locally with all resourced downloading correctly

PR #1158, opened 2022-04-18
Bug fix: Show labels on layer in 3D

vquach2404:
## Description

This PR fixes a bug in the Show labels on layer in 3D sample. The data seems to have changed sometime after the original implementation. The change in data made the code outdated.
Branch: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Viv/fix3DLabel

## Linked Issue(s)

- `common-samples/3451`

## How To Test

Check to ensure the labels are shown in the Show labels on layer in 3D sample.

### DIFF
--- a/arcgis-ios-sdk-samples/Scenes/Show labels on layer in 3D/ShowLabelsOnLayer3DViewController.swift
+++ b/arcgis-ios-sdk-samples/Scenes/Show labels on layer in 3D/ShowLabelsOnLayer3DViewController.swift
@@ -34,7 +34,7 @@ class ShowLabelsOnLayer3DViewController: UIViewController {
         scene.load { [weak self] error in
             guard let self = self else { return }
             // Get the "Gas Main" feature layer.
-            if let groupLayer = scene.operationalLayers.compactMap({ $0 as? AGSGroupLayer }).first(where: { $0.name == "Gas" }),
+            if let groupLayer = scene.operationalLayers.lazy.compactMap({ $0 as? AGSGroupLayer }).first(where: { $0.name == "Gas" }),
                let layers = groupLayer.layers as? [AGSLayer],
                let gasFeatureLayer = layers.first(where: { $0.name == "Gas Main" }) as? AGSFeatureLayer {
                 let labelDefinition = self.makeLabelDefinition()
--- a/arcgis-ios-sdk-samples/Scenes/Show labels on layer in 3D/ShowLabelsOnLayer3DViewController.swift
+++ b/arcgis-ios-sdk-samples/Scenes/Show labels on layer in 3D/ShowLabelsOnLayer3DViewController.swift
@@ -34,8 +34,7 @@ class ShowLabelsOnLayer3DViewController: UIViewController {
         scene.load { [weak self] error in
             guard let self = self else { return }
             // Get the "Gas Main" feature layer.
-            let operationalLayers = scene.operationalLayers as [AnyObject]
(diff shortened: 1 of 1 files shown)


arcgis-ios-sdk-samples/Scenes/Show labels on layer in 3D/ShowLabelsOnLayer3DViewController.swift
```diff
@@ -33,9 +33,9 @@ class ShowLabelsOnLayer3DViewController: UIViewController {
         // Load the scene.
         scene.load { [weak self] error in
             guard let self = self else { return }
-            // Get the feature layer.
-            if let operationalLayers = scene.operationalLayers as? [AGSGroupLayer],
-               let groupLayer = operationalLayers.first(where: { $0.name == "Gas" }),
+            // Get the "Gas Main" feature layer.
+            let operationalLayers = scene.operationalLayers as [AnyObject]
+            if let groupLayer = operationalLayers.first(where: { $0.name == "Gas" }) as? AGSGroupLayer,
```
yo1995, line 38:
```suggestion
            if let groupLayer = scene.operationalLayers.compactMap({ $0 as? AGSGroupLayer }).first(where: { $0.name == "Gas" })
```
Instead of casting the array to ` [AnyObject]`, I prefer first filter the elements down to `AGSGroupLayer`, then find the proper layer from it.

yo1995 (approved):

yo1995 (dismissed):

arcgis-ios-sdk-samples/Scenes/Show labels on layer in 3D/ShowLabelsOnLayer3DViewController.swift
```diff
@@ -33,9 +33,8 @@ class ShowLabelsOnLayer3DViewController: UIViewController {
         // Load the scene.
         scene.load { [weak self] error in
             guard let self = self else { return }
-            // Get the feature layer.
-            if let operationalLayers = scene.operationalLayers as? [AGSGroupLayer],
-               let groupLayer = operationalLayers.first(where: { $0.name == "Gas" }),
+            // Get the "Gas Main" feature layer.
+            if let groupLayer = scene.operationalLayers.compactMap({ $0 as? AGSGroupLayer }).first(where: { $0.name == "Gas" }),
                let layers = groupLayer.layers as? [AGSLayer],
```
philium, line 38: Does the scene have many operational layers? If not, then what you have is fine. But if it does, it would be good to lazily compact map, so as not to _map_ all layers and then find the first one (which is the case now):

```swift
if let groupLayer = scene.operationalLayers.lazy.compactMap({ $0 as? AGSGroupLayer }).first(where: { $0.name == "Gas" }),
```

Another way is to just use `first(where:)`, although that does require casting the found layer twice:

```swift
if let groupLayer = scene.operationalLayers.first(where: { ($0 as? AGSGroupLayer)?.name == "Gas" }) as? AGSGroupLayer,
```

philium (dismissed):

arcgis-ios-sdk-samples/Scenes/Show labels on layer in 3D/ShowLabelsOnLayer3DViewController.swift
vquach2404: This scene has 7 layers. I'll update it to lazily compact!

yo1995 (approved):

philium (approved):

saratchandrakarumuri (approved):
Labels are now rendered as expected

PR #1159, opened 2022-04-19
Enhancement: Add favorites capability

vquach2404:
## Description

This PR adds a "Favorites" category with the ability to track the user's favorite samples.
Branch: https://github.com/Esri/arcgis-runtime-samples-ios/tree/Viv/addFavorites

## Linked Issue(s)

- `common-samples/3274`

## How To Test

Add/remove some samples by tapping the star in the upper right hand corner of a sample. 

## To Discuss

There is an observer which is removed [here](https://github.com/Esri/arcgis-runtime-samples-ios/blob/7768c650435739e8f66ff4c77082b3b786bdc412/arcgis-ios-sdk-samples/Content%20Display%20Logic/Controllers/ContentTableViewController.swift#L69-L72), but I've found that this is not the best place to remove it because the observer is removed before the sample is shown. Thus, `UserDefaults` is never notified of a change. Where else can I remove the observer?


arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
```diff
@@ -312,7 +312,7 @@
 				<key>displayName</key>
 				<string>Show device location using indoor positioning</string>
 				<key>descriptionText</key>
-				<string>Show your device's real-time location while inside a building by using signals from indoor positioning beacons.</string>
+				<string>Show your device&apos;s real-time location while inside a building by using signals from indoor positioning beacons.</string>
```
yo1995, line 315: Hmm, weird, I believe this is already on `v.next`. Try to rebase or merge v.next to your branch to keep up to date.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -56,6 +62,13 @@ class ContentTableViewController: UITableViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         selectedSample = nil
+
```
yo1995, line 65:
```suggestion
```

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -187,34 +200,37 @@ class ContentTableViewController: UITableViewController {
...
+        // Present the sample view controller.
         presentingController?.showDetailViewController(navController, sender: self)
         
-        // create and setup the info button
+        // Create and setup the info button.
         let infoBBI = SourceCodeBarButtonItem()
         infoBBI.readmeURL = sample.readmeURL
         infoBBI.navController = navController
-        controller.navigationItem.rightBarButtonItem = infoBBI
+        infoBBI.sample = sample
```
yo1995, line 222: This line is wrong, I assume it should be `favoritesBarButtonItem.sample = sample` ?

Also let's rename `infoBBI` to `sourceCodeBarButtonItem`.

yo1995 (changes requested):

vquach2404:
@yo1995 I've applied your suggestions but another issue remains:
Simulator Screen Recording - iPhone 13 Pro Max - 2022-04-21 at 17 14 01
The `UserDefaults` notification is set off before the new dictionary is re-set. So the favorites table view is reloaded before the dictionary is set.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -44,10 +46,17 @@ class ContentTableViewController: UITableViewController {
         IndexPath(row: displayedSamples.firstIndex(of: sample)!, section: 0)
     }
     
+    @objc
+    func userDefaultsDidChange() {
+        guard isViewLoaded && isFavoritesCategory else { return }
+        allSamples = allSamples.filter { $0.isFavorite }
+        tableView.reloadData()
+    }
+
```
yo1995, line 55:
```suggestion
```

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -56,6 +65,16 @@ class ContentTableViewController: UITableViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         selectedSample = nil
+        if isFavoritesCategory {
+            NotificationCenter.default.removeObserver(self, name: UserDefaults.didChangeNotification, object: nil)
+        }
+    }
```
yo1995, line 71:
```suggestion
    override func viewWillAppear(_ animated: Bool) {
        super.viewWillAppear(animated)
        selectedSample = nil
        if isFavoritesCategory {
            allSamples = allSamples.filter { $0.isFavorite }
            tableView.reloadData()
        }
    }
```

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -56,6 +65,16 @@ class ContentTableViewController: UITableViewController {
...
+    
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+        if isFavoritesCategory {
+            NotificationCenter.default.addObserver(self, selector: #selector(userDefaultsDidChange), name: UserDefaults.didChangeNotification, object: nil)
+        }
     }
```
yo1995, line 78:
```suggestion
```

yo1995 (commented):
Reply to https://github.com/Esri/arcgis-runtime-samples-ios/pull/1159#issuecomment-1105886316

In this case we may reload the table every time it is displayed, regardless whether or not a sample favorite status is changed when the table isn't visible, so that when favorite change from true to false to true, the sample isn't removed from `allSamples`

However this won't work for iPad with split view controller, so don't apply these changes now... let's come up with some other ideas.

yo1995:
The other idea is to maintain 2 arrays of samples: includedSamples, excludedSamples, like https://github.com/Esri/arcgis-runtime-samples-ios/blob/main/arcgis-ios-sdk-samples/Utility%20network/Create%20load%20report/CreateLoadReportViewController.swift#L41 

But I think that is an overkill for this simple task. 🤔

yo1995 (commented):
Thanks for applying the change. Don't know if you saw my comments below. Since you applied them, I suggest let's add a few more reviewers to this PR to have their ideas.

https://user-images.githubusercontent.com/9660181/165140845-5693af2a-ec79-4c2a-9202-932eef676e65.mp4

As you can see in the video, the list doesn't update on iPad because the content table VC is present when we favorite/unfavorite the sample. Unless we keep the notification alive for the entire time, I cannot come up with a good way to listen to the favorite button's action.

The problem is that the favorite button is in the children VC of the content table VC, and tries to change the data in its parent. Again, let's see what others say.

yo1995 (commented):
Try to unset the "change requested" status

yo1995 (dismissed):

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -143,6 +146,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
         return nil
     }
     
+    // MARK: - User data
+    
+    /// Creates a dictionary to store the user's favorited samples.
+    func createUserDefaults() {
+        if UserDefaults.standard.dictionary(forKey: "favorites") == nil {
```
mhdostal, line 153: Can the string "favorites" be made a static property, so you don't have to use the same string twice?

yo1995:
> Instead of including `star.fill.svg`, can you use the SFSymbol: "star.fill"?

I believe this file is included to keep consistent with other categories' logos - they are all custom so they are all loaded from the image assets, see here: https://github.com/Esri/arcgis-runtime-samples-ios/blob/main/arcgis-ios-sdk-samples/Content%20Display%20Logic/Controllers/ContentCollectionViewController.swift#L75

If change to SFSymbol, then some code needs to be added there to handle the special case of Favorites "category" cell.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
```diff
@@ -89,12 +89,20 @@ class ContentCollectionViewController: UICollectionViewController, UICollectionV
...
+        // Hide keyboard if visible.
         view.endEditing(true)
-        
         let category = categories[indexPath.item]
         let controller = storyboard!.instantiateViewController(withIdentifier: "ContentTableViewController") as! ContentTableViewController
-        controller.allSamples = category.samples
+        // Filter and display the favorited samples.
+        if category.name == "Favorites" {
+            let allSamples = categories.flatMap { $0.samples }
+            let allFavorites = allSamples.filter { $0.isFavorite }
```
mhdostal, line 99: `ContentTableViewController.swift` is also performing a `.filter { $0.isFavorite }`.  Can one of them be removed?  If you do the filter here, is `ContentTableViewController.isFavoritesCategory` really necessary?

mhdostal (changes requested):

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentCollectionViewController.swift
vquach2404: Yes, both are necessary. Unfortunately, there's no simple way to generalize both cases. Both portions load the favorited samples but each are part of an existing process to handle all the categories.

mhdostal (approved):

arcgis-ios-sdk-samples/Content Display Logic/Custom BarButtonItem/FavoritesBarButtonItem.swift
```diff
@@ -0,0 +1,51 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+
+class FavoritesBarButtonItem: UIBarButtonItem {
```
philium, line 17:
```suggestion
final class FavoritesBarButtonItem: UIBarButtonItem {
```

arcgis-ios-sdk-samples/Content Display Logic/Custom BarButtonItem/FavoritesBarButtonItem.swift
```diff
@@ -0,0 +1,51 @@
...
+    @objc
+    func toggleIsFavorite() {
+        // Update the bool.
+        sample.isFavorite.toggle()
+        // Update the image.
+        self.image = updateImage(isFavorite: sample.isFavorite)
+    }
+    
+    /// Update the image to indicate if the sample is favorited.
+    func updateImage(isFavorite: Bool) -> UIImage {
```
philium, line 43: I would suggest renaming this to `makeImage(isFavorite:)`.

arcgis-ios-sdk-samples/Content Display Logic/Custom BarButtonItem/FavoritesBarButtonItem.swift
```diff
@@ -0,0 +1,51 @@
...
+        switch isFavorite {
+        case true:
+            return UIImage(systemName: "star.fill")!
+        case false:
+            return UIImage(systemName: "star")!
+        }
```
philium, line 49: As a rule of thumb, _never_ switch over a Boolean value. There's always a better way.

```suggestion
        let systemName = isFavorite ? "star.fill" : "star"
        return UIImage(systemName: systemName)!
```

arcgis-ios-sdk-samples/Content Display Logic/Custom BarButtonItem/FavoritesBarButtonItem.swift
```diff
@@ -0,0 +1,51 @@
...
+
+class FavoritesBarButtonItem: UIBarButtonItem {
+    var sample: Sample
+    
+    init(sample: Sample) {
+        self.sample = sample
+        super.init()
+        self.image = updateImage(isFavorite: sample.isFavorite)
+        self.target = self
+        self.action = #selector(FavoritesBarButtonItem.toggleIsFavorite)
```
philium, line 25:
```suggestion
            action: #selector(toggleIsFavorite)
```

arcgis-ios-sdk-samples/Content Display Logic/Model/Sample.swift
```diff
@@ -21,6 +21,23 @@ struct Sample: Hashable {
     var description: String
     var storyboardName: String
     var dependencies: [String]
+    var isFavorite: Bool {
```
philium, line 24: Please move this into an extension to keep from cluttering the type. Also, think about this. The value stored in the dictionary is always `true`, which means that the value isn't needed. The presence of the key/value pair is enough to indicate whether the sample is a favorite. So what we really only need is an unordered collection of unique keys. What does that sound like? A set. `UserDefaults ` doesn't support sets, but it does support arrays. The storage should be an array of sample names viewed as a set. Since that set is really a property of the user defaults instance, I would recommend modeling it as such:

```swift
extension UserDefaults {
    static let favoriteSamplesKey = "favoriteSamples"
    
    var favoriteSamples: Set<String> {
        get {
            Set(stringArray(forKey: Self.favoriteSamplesKey)!)
        }
        set {
            set(newValue.sorted(), forKey: Self.favoriteSamplesKey)
        }
    }
}
```

And then the implementation of `isFavorite` would look something like this:

```swift
extension Sample {
    var isFavorite: Bool {
        get {
            UserDefaults.standard.favoriteSamples.contains(name)
        }
        nonmutating set {
            if newValue {
                UserDefaults.standard.favoriteSamples.insert(name)
            } else {
                UserDefaults.standard.favoriteSamples.remove(name)
            }
        }
    }
}
```

arcgis-ios-sdk-samples/Content Display Logic/ContentPList.plist
```diff
@@ -1965,5 +1965,11 @@
 			</dict>
 		</array>
 	</dict>
+	<dict>
+		<key>displayName</key>
+		<string>Favorites</string>
+		<key>children</key>
+		<array/>
+	</dict>
```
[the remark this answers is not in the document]
vquach2404: Putting the category first is a great idea! Although, I disagree with dynamically including the favorites category. I think it could be confusing for the user. For example, if a user forgets that they have no favorited samples remaining or if they unknowingly unfavorited the sample on accident. They may wonder why they can't access their favorites anymore. Instead, we can put placeholder text when the view controller is empty.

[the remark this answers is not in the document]
vquach2404: I like the idea of adding the favorites at the top of the list since a benefit of having a favorites category is having quick access to specific samples. But I will add text when there are no favorited samples.

arcgis-ios-sdk-samples/Content Display Logic/Custom BarButtonItem/FavoritesBarButtonItem.swift
vquach2404: Is there a typo here? There are build errors when I tried to apply the suggestion.

philium: Uh, yes, there is. The suggestion should have just removed the `FavoritesBarButtonItem.` part.

arcgis-ios-sdk-samples/Content Display Logic/Controllers/ContentTableViewController.swift
```diff
@@ -56,6 +58,10 @@ class ContentTableViewController: UITableViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         selectedSample = nil
+        if isFavoritesCategory {
+            allSamples = allSamples.filter { $0.isFavorite }
+            tableView.reloadData()
+        }
```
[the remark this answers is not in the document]
yo1995: Do you mean to subclass `ContentTableViewController` as `class FavoritesCategoryViewController: ContentTableViewController`? Because the favorites are displayed in a table view.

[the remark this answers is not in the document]
philium: Yes, although it has been so long since I've used a UITableView that I forgot about the data source. Instead of subclassing `ContentTableViewController`, you use two different data sources (instance of type that conforms to `UITableViewDataSource`--`UITableViewDiffableDataSource` would be a good option): one that's backed by an array of samples and another one that is backed by the favorite samples, with the logic to get the favorites and keep it up-to-date.

mhdostal (approved):

arcgis-ios-sdk-samples/Content Display Logic/Controllers/CategoryTableViewController.swift
```diff
@@ -47,15 +52,29 @@ class ContentTableViewController: UITableViewController {
...
+        dataSource = UITableViewDiffableDataSource(tableView: tableView, cellProvider: { [unowned self] tableView, indexPath, _ in
+            let sample = self.displayedSamples[indexPath.row]
+            
+            let cell = tableView.dequeueReusableCell(withIdentifier: "ContentTableCell", for: indexPath) as! ContentTableCell
+            cell.titleLabel.text = sample.name
+            cell.detailLabel.text = sample.description
+            cell.isExpanded = expandedRowIndexPaths.contains(indexPath)
+            return cell
+        })
+        updateDataSource()
```
yo1995, line 69: For next reviewer's context: the table view won't reload if the sample's favorite state is changed when the table view is presented (like an iPad split view). We decided that favorite state-change notification is an overkill here.

yo1995 (approved):

arcgis-ios-sdk-samples/AppDelegate.swift
```diff
@@ -164,6 +176,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
...
 }
 
+extension UserDefaults {
+    static let favoriteSamplesKey = "favoriteSamples"
+    var favoriteSamples: Set<String> {
+        get {
+            Set(stringArray(forKey: Self.favoriteSamplesKey)!)
+        }
+        set {
+            set(newValue.sorted(), forKey: Self.favoriteSamplesKey)
```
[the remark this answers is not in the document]
vquach2404: Not really sure what comment you're referring to. Are you suggesting a change?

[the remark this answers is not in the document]
yo1995: Yes, I'm asking why it's `newValue.sorted()`. IMO it doesn't need to be sorted because set is unordered data structure.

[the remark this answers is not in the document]
vquach2404: Ahh I'll leave this open for discussion since Phil made this change during our pair programming session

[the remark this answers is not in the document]
philium: Sorting the elements of the set ensures that the stores representation is consistent. A similar thing is seen with property list dictionary where the key value pairs are sorted by key.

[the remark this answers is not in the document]
yo1995: That makes sense.

philium (approved):

saratchandrakarumuri (approved):

PR #1162, opened 2022-04-26
[Update] Raster service data URL

yo1995:
## Description

This PR fixes an invalid raster image service.

## Linked Issue(s)

- `common-samples/issues/3499`
- `common-samples/issues/3503`

## How To Test

Old doesn't work, new works.

It does take a few seconds to load though.


vquach2404 (approved):
Data shows as expected. Although, it was slow to load. We should consider adding a loading indicator (to be discussed and resolved in a separate issue given the current time constraint).

saratchandrakarumuri (approved):
New service is working as expecting

PR #1174, opened 2022-06-01
Fix: Authenticate with OAuth bug

vquach2404:
## Description

This PR fixes `Token Authentication` in `Cloud and Portal` category.
Branch: [URL_TO_BRANCH](https://github.com/Esri/arcgis-runtime-samples-ios/tree/Viv/fixAuthenticateOAuth)

## Linked Issue(s)

- `common-samples/issues/3491`

## How to Test

First open the `Authenticate with OAuth` sample, log in, then open the `Token Authentication` sample. Ensure that you're prompted to log in upon `Token Authentication` launching.

## To Discuss

Although this fixes the problem, it may not be the best solution because it deviates from the token authentication workflow and best practices. Additionally, it's redundant since the credential cache is cleared twice.

### DIFF
--- a/arcgis-ios-sdk-samples/Cloud and portal/Authenticate with OAuth/AuthenticateWithOAuthViewController.swift
+++ b/arcgis-ios-sdk-samples/Cloud and portal/Authenticate with OAuth/AuthenticateWithOAuthViewController.swift
@@ -58,7 +58,8 @@ class AuthenticateWithOAuthViewController: UIViewController {
         AGSAuthenticationManager.shared().oAuthConfigurations.add(oAuthConfiguration)
     }
     
-    deinit {
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
         AGSAuthenticationManager.shared().oAuthConfigurations.remove(oAuthConfiguration)
         AGSAuthenticationManager.shared().credentialCache.removeAllCredentials()
     }
--- a/arcgis-ios-sdk-samples/Cloud and portal/Authenticate with OAuth/AuthenticateWithOAuthViewController.swift
+++ b/arcgis-ios-sdk-samples/Cloud and portal/Authenticate with OAuth/AuthenticateWithOAuthViewController.swift
@@ -58,12 +58,6 @@ class AuthenticateWithOAuthViewController: UIViewController {
         AGSAuthenticationManager.shared().oAuthConfigurations.add(oAuthConfiguration)
     }
     
-    override func viewDidDisappear(_ animated: Bool) {
-        super.viewDidDisappear(animated)
-        AGSAuthenticationManager.shared().oAuthConfigurations.remove(oAuthConfiguration)
-        AGSAuthenticationManager.shared().credentialCache.removeAllCredentials()
-    }
-    
     // MARK: UIViewController
     
     override func viewDidLoad() {
@@ -72,6 +66,13 @@ class AuthenticateWithOAuthViewController: UIViewController {
         (navigationItem.rightBarButtonItem as? SourceCodeBarButtonItem)?.filenames = [
             "AuthenticateWithOAuthViewController"
         ]
+    }
+    
+    // Clear the credential cache upon exiting the sample.
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+        AGSAuthenticationManager.shared().oAuthConfigurations.remove(oAuthConfiguration)
+        AGSAuthenticationManager.shared().credentialCache.removeAllCredentials()
     }
 }
 
--- a/arcgis-ios-sdk-samples/Cloud and portal/Authenticate with OAuth/AuthenticateWithOAuthViewController.swift
+++ b/arcgis-ios-sdk-samples/Cloud and portal/Authenticate with OAuth/AuthenticateWithOAuthViewController.swift
@@ -68,7 +68,7 @@ class AuthenticateWithOAuthViewController: UIViewController {
         ]
     }
     
-    // Clear the credential cache upon exiting the sample.
+    // Clearing the credential cache so sample go through authentication every time.
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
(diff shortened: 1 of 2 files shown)


arcgis-ios-sdk-samples/Cloud and portal/Token authentication/TokenAuthenticationViewController.swift
```diff
@@ -34,6 +34,12 @@ class TokenAuthenticationViewController: UIViewController {
         return AGSMap(item: portalItem)
     }
     
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        
+        AGSAuthenticationManager.shared().credentialCache.removeAllCredentials()
```
vquach2404, line 40: The credentials are cleared at the end of this [sample](https://github.com/Esri/arcgis-runtime-samples-ios/blob/8e269567c08f19c626a62940b60dd4003ee9d651/arcgis-ios-sdk-samples/Cloud%20and%20portal/Authenticate%20with%20OAuth/AuthenticateWithOAuthViewController.swift#L61-L64) as well. So it is redundant but we should clear the cache in `Authenticate with OAuth` to showcase the correct workflow.

arcgis-ios-sdk-samples/Cloud and portal/Token authentication/TokenAuthenticationViewController.swift
```diff
@@ -34,6 +34,12 @@ class TokenAuthenticationViewController: UIViewController {
         return AGSMap(item: portalItem)
     }
     
+    required init?(coder aDecoder: NSCoder) {
```
yo1995, line 37: I'm in favor of not making this change, although it might cause problem when users are trying out this sample on an iPad. Ideally credential caches should be handled/cleared when a certain view controller is dismissed from view hierarchy or deinit'd. 

The unique sample viewer architecture caused a sample to persist in memory until a new sample is loaded, which makes it kind of a racing condition - old sample is de-initializing and clearing cache, and the new sample is loading from cache. This should hardly happen in a real app.

However I'm OK to add this workaround if we provide some detailed "why we add this" explanation here.

njarecha: The purpose of the samples are to demonstrate the functionality so it should do the right thing. I would suggest using the `viewDidDisappear` for OAuth and Token samples to clear the cache.

AuthenticateWithOAuthViewController - Move code from `deinit` to `viewDidDisappear` 
```
    override func viewDidDisappear(_ animated: Bool) {
        AGSAuthenticationManager.shared().oAuthConfigurations.remove(oAuthConfiguration)
        AGSAuthenticationManager.shared().credentialCache.removeAllCredentials()
    }
``

TokenAuthenticationViewController - Add code
```
    override func viewDidDisappear(_ animated: Bool) {
        AGSAuthenticationManager.shared().credentialCache.removeAllCredentials()
    }
```

Comment (njarecha) [reply arcgis-ios-sdk-samples/Cloud and portal/Token authentication/TokenAuthenticationViewController.swift:37]: I don't think so. I can see in all places credential is either created while handling authentication challenge or set on the object before loading. That is fine.

Comment (yo1995) [reply arcgis-ios-sdk-samples/Cloud and portal/Token authentication/TokenAuthenticationViewController.swift:37]: If the credentialCache is cleared on `viewDidDisappear`, say in the OAuth sample, when the user first logs in, then tap the upper right source code page button, the credentials will be cleared. When the source code view controller is popped, the user will see a blank screen?

Comment (njarecha) [reply arcgis-ios-sdk-samples/Cloud and portal/Token authentication/TokenAuthenticationViewController.swift:37]: Yes, `viewDidDisappear` will be called in that case but everything will continue to work on the map. When any object used the credential to load than that credential will be available on that object's `credential` property. Only other option is to do that in init before doing anything else in both samples.

Comment (njarecha) [reply arcgis-ios-sdk-samples/Cloud and portal/Token authentication/TokenAuthenticationViewController.swift:37]: Also, comments should be added to let user know why we are clearing the credentials.

Comment (vquach2404) [reply arcgis-ios-sdk-samples/Cloud and portal/Token authentication/TokenAuthenticationViewController.swift]: Thanks for the discussion! From what I've gathered, implementing `viewDidDisappear` will suffice and there's no need for `viewWillAppear`. I've made the changes accordingly. Please let me know if I understood correctly!

Comment (yo1995) [reply arcgis-ios-sdk-samples/Cloud and portal/Token authentication/TokenAuthenticationViewController.swift]: > ... but everything will continue to work on the map. When any object used the credential to load than that credential will be available on that object's `credential` property. 

This is crucial info, thanks for the clarification. I was assuming by clearing the credentialCache the user's authenticated state will be nullified and cannot make further requests to the protected endpoint.

**Review by yo1995 — APPROVED** [review]

File: arcgis-ios-sdk-samples/Cloud and portal/Authenticate with OAuth/AuthenticateWithOAuthViewController.swift [hunk 3]
```diff
@@ -72,6 +67,13 @@ class AuthenticateWithOAuthViewController: UIViewController {
             "AuthenticateWithOAuthViewController"
         ]
     }
+    
+    // Clear the credential cache upon exiting the sample.
```
njarecha, line 71:
```suggestion
    // Clearing the credential cache so sample to go through authentication every time.
```

arcgis-ios-sdk-samples/Cloud and portal/Token authentication/TokenAuthenticationViewController.swift
```diff
@@ -43,4 +43,10 @@ class TokenAuthenticationViewController: UIViewController {
             "TokenAuthenticationViewController"
         ]
     }
+    
+    // Clear the credential cache upon exiting the sample.
```
njarecha, line 47:
```suggestion
    // Clearing the credential cache so sample to go through authentication every time.
```

njarecha (changes requested):

saratchandrakarumuri (dismissed):
Both the samples are now working as intended

njarecha (approved):

PR #1182, opened 2022-06-27
New: Query features with Arcade expression

vquach2404:
## Description

This PR implements `Query features with Arcade expression` in `Display information` category.
Branch: [URL_TO_BRANCH](https://github.com/Esri/arcgis-runtime-samples-ios/tree/Viv/queryFeaturesArcadeExpression)

## Linked Issue(s)

- `common-samples/issues/#3611`

## To discuss

The data changes often and is inconsistent. While testing, you may see different numbers than a previous test; this is expected.


arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,111 @@
...
+    var evaluateOperation: AGSCancelable?
+    
+    /// Make and load a map.
+    func makeMap() -> AGSMap {
+        // Create a portal item with the portal and item ID.
+        let portalItem = AGSPortalItem(portal: .arcGISOnline(withLoginRequired: false), itemID: "14562fced3474190b52d315bc19127f6")
+        // Make a map with the portal item.
+        let map = AGSMap(item: portalItem)
+        // Load the map.
+        map.load() { _ in
```
yo1995, line 37:
```suggestion
        map.load { _ in
```

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,111 @@
...
+                }
+            }
+        }
+        return map
+    }
+    
+    /// Evaluate the arcade expression for the selected feature at the map point.
+    func evaluateArcadeInCallout(for feature: AGSArcGISFeature, at mapPoint: AGSPoint) {
+        // Instantiate a string containing the arcade expression.
+        let expressionValue = "var crimes = FeatureSetByName($map, 'Crime in the last 60 days');\n" + "return Count(Intersects($feature, crimes));"
```
yo1995, line 52:
```suggestion
        let expressionValue =
        """
        var crimes = FeatureSetByName($map, 'Crime in the last 60 days');
        return Count(Intersects($feature, crimes));
        """
```
How about using multiline string here?

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,111 @@
...
+        // Create an arcade evaluator with the arcade expression and an arcade profile.
+        let evaluator = AGSArcadeEvaluator(expression: expression, profile: .formCalculation)
+        guard let map = mapView.map else { return }
+        let profileVariables = ["$feature": feature, "$map": map]
+        // Get the arcade evaluation result given the previously set profile variables.
+        evaluateOperation = evaluator.evaluate(withProfileVariables: profileVariables) { [weak self] result, error in
+            guard let self = self else { return }
+            if let result = result, let crimeCount = result.cast(to: .string) as? String {
+                // Dismiss progress hud.
+                UIApplication.shared.hideProgressHUD()
```
yo1995, line 64:
```suggestion
            // Dismiss progress hud.
            UIApplication.shared.hideProgressHUD()
            guard let self = self else { return }
            if let result = result, let crimeCount = result.cast(to: .string) as? String {
```
Dismiss the HUD regardless result or error, so it won't overlay upon the alert.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,111 @@
...
+                // Hide the accessory button.
+                self.mapView.callout.isAccessoryButtonHidden = true
```
yo1995, line 66: This can be in mapView didSet.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,111 @@
...
+                // Set the detail text.
+                self.mapView.callout.detail = "Crimes in the last 60 days: \(crimeCount)"
```
yo1995, line 68: Phil may suggest using string formatting (`String(format:)`) here to advocate supporting localization.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,111 @@
...
+                // Prompt the callout at the map point.
+                self.mapView.callout.show(at: mapPoint, screenOffset: .zero, rotateOffsetWithMap: false, animated: true)
```
yo1995, line 70:
```suggestion
                self.mapView.callout.show(for: feature, tapLocation: mapPoint, animated: true)
```

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,111 @@
...
+extension QueryFeaturesArcadeExpressionViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        // Dismiss any presenting callout.
+        mapView.callout.dismiss()
+        // Identify features at the tapped location.
+        mapView.identifyLayers(atScreenPoint: screenPoint, tolerance: 12, returnPopupsOnly: false) { [weak self] results, error in
+            guard let results = results, let self = self else { return }
+            if let elements = results.first?.geoElements {
+                // Get the selected feature.
+                guard let identifiedFeature = elements.first as? AGSArcGISFeature else { return }
```
yo1995, line 100: This line can be part of the if let condition.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,111 @@
...
+        // Instantiate a string containing the arcade expression.
+        let expressionValue = "var crimes = FeatureSetByName($map, 'Crime in the last 60 days');\n" + "return Count(Intersects($feature, crimes));"
+        // Create an arcade expression using the string.
+        let expression = AGSArcadeExpression(expression: expressionValue)
+        // Create an arcade evaluator with the arcade expression and an arcade profile.
+        let evaluator = AGSArcadeEvaluator(expression: expression, profile: .formCalculation)
+        guard let map = mapView.map else { return }
+        let profileVariables = ["$feature": feature, "$map": map]
+        // Get the arcade evaluation result given the previously set profile variables.
+        evaluateOperation = evaluator.evaluate(withProfileVariables: profileVariables) { [weak self] result, error in
```
yo1995, line 60: The `evaluateOperation` doesn't seem to be used. Either remove it, or cancel before a new operation and after the operation calls back.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,111 @@
...
+        // Create a portal item with the portal and item ID.
+        let portalItem = AGSPortalItem(portal: .arcGISOnline(withLoginRequired: false), itemID: "14562fced3474190b52d315bc19127f6")
+        // Make a map with the portal item.
+        let map = AGSMap(item: portalItem)
+        // Load the map.
+        map.load() { _ in
+            // Set the visibility of all but the RDT Beats layer to false.
+            map.operationalLayers.forEach { layer in
+                let currentLayer = layer as? AGSLayer
+                if currentLayer?.name == "Crime in the last 60 days" || currentLayer?.name == "Police Stations" {
```
yo1995, line 41: Is it able to use the opposite condition, i.e., (and adjust the comment)

```swift
if let currentLayer = layer as? AGSLayer {
    currentLayer.isVisible = currentLayer.name.contains("RPD Beats")
}
```

Also I'm seeing a copy layer in the data. Should we remove it?

image

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/README.md
```diff
@@ -0,0 +1,58 @@
...
+## How it works
+
+1. Create an `AGSPortalItem` using the portal and ID.
+2. Create an `AGSMap` using the portal item.
+3. Set the visibility of all the layers to false, except for the layer named "RPD Beats  - City_Beats_Border_1128-4500".
+4. Set the `AGSGeoViewTouchDelegate` to the map.
+5. Identify the layers tapped on the map view with `AGSGeoView.identifyLayers(atScreenPoint:tolerance:returnPopupsOnly:completion:)`.
+6. Create the following `AGSArcadeExpression`:
+
+	 ```swift
```
yo1995, line 24: The indentation here is off.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+	 ```swift		
+	 expressionValue = "var crimes = FeatureSetByName($map, 'Crime in the last 60 days');\n"
+     				   "return Count(Intersects($feature, crimes));"
+     ```
+            
+7. Create an `AGSArcadeEvaluator` using the Arcade expression and `ArcadeProfile.FORM_CALCULATION`.
+8. Create a dictionary of profile variables with the following pairs:
+
+	 `{"$feature", identifiedFeature}`
```
yo1995, line 32: As you may tell, the status check complains about the hard tabs used here. Replace them with spaces.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/README.md
```diff
@@ -0,0 +1,58 @@
...
+* AGSArcadeEvaluationResult
+* AGSArcadeEvaluator
+* AGSArcadeExpression
+* AGSArcadeProfile
+* AGSPortal
+* AGSPortalItem
+
+## About the data
+
+This sample uses the [Crimes in Police Beats Sample](https://www.arcgis.com/home/item.html?id=14562fced3474190b52d315bc19127f6) ArcGIS Online Web Map which contains 3 layers for police stations, city beats borders, and crimes in the last 60 days as recorded by the Rochester, NY police department.
```
yo1995, line 50: See my comment above. Can you verify the data contains 3 layers?

yo1995 (commented):
Sometimes when I consecutively identify the features, the evaluation doesn't end. No output in console either. Do you also see this? If so we may need to file an API bug.

https://user-images.githubusercontent.com/9660181/176062520-498034be-5c50-46a4-ae62-bb295aac55bc.mp4

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
vquach2404: The data used is not managed by the Runtime account so I don't think we should alter the data.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
vquach2404: I prefer to leave it here since the other callout adjustments are here as well. Splitting it up makes it feel disorganized IMO 😅

vquach2404:
@mhdostal Hi Mark, please review this since Ting's on vacation. Thanks!

vquach2404:
> Sometimes when I consecutively identify the features, the evaluation doesn't end. No output in console either. Do you also see this? If so we may need to file an API bug.

I made some changes and haven't been able to reproduce this the last few times I tested the sample. Please let me know if anyone else continues to see this behavior.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+        let expressionValue =
+        """
+        var crimes = FeatureSetByName($map, 'Crime in the last 60 days');
+        return Count(Intersects($feature, crimes));
+        """
+        // Create an arcade expression using the string.
+        let expression = AGSArcadeExpression(expression: expressionValue)
+        // Create an arcade evaluator with the arcade expression and an arcade profile.
+        let evaluator = AGSArcadeEvaluator(expression: expression, profile: .formCalculation)
+        guard let map = mapView.map else { return }
```
sbaskaran, line 62: Again I'm curious, is it a pattern samples follow not to have a global variable for map?

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+        let profileVariables = ["$feature": feature, "$map": map]
+        // Get the arcade evaluation result given the previously set profile variables.
+        if evaluateOperation != nil {
+            evaluateOperation?.cancel()
```
sbaskaran, line 66: If we are anyway checking for `nil`, why not unwrap it?

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+        }
+        evaluateOperation = evaluator.evaluate(withProfileVariables: profileVariables) { [weak self] result, error in
+            // Dismiss progress hud.
+            UIApplication.shared.hideProgressHUD()
+            guard let self = self else { return }
+            if let result = result, let crimeCount = result.cast(to: .string) as? String {
```
sbaskaran, line 72: Can this line be combined with the previous `guard` statement

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/README.md
sbaskaran: Ha, until this point, I assumed RPD was Redlands PD :D

sbaskaran (commented):
hey @vquach2404 I left a few questions/comments

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+class QueryFeaturesArcadeExpressionViewController: UIViewController {
+    @IBOutlet var mapView: AGSMapView! {
+        didSet {
+            mapView.map = makeMap()
```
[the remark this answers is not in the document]
vquach2404: Yes, "make" is a common term we use in almost all our samples! You can see an example [here](https://github.com/Esri/arcgis-runtime-samples-ios/blob/67946e13bdda2ca6bbdd3dfe8fee6e868f8b7c32/arcgis-ios-sdk-samples/Maps/Change%20basemap/SwitchBasemapViewController.swift#L22).

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
vquach2404: Yes, we usually don't make map a global variable since it's not needed. I only guarded it in this case because of its later, one-time use.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
vquach2404: It could, although I feel it may be better to keep separate. If I guard `result` and `crimeCount` as well, it would prevent the error from being presented.

vquach2404:
Thanks @sbaskaran! I made some changes and responded to some of your comments.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,119 @@
...
+        let expressionValue =
+        """
+        var crimes = FeatureSetByName($map, 'Crime in the last 60 days');
+        return Count(Intersects($feature, crimes));
+        """
+        // Create an arcade expression using the string.
+        let expression = AGSArcadeExpression(expression: expressionValue)
+        // Create an arcade evaluator with the arcade expression and an arcade profile.
+        let evaluator = AGSArcadeEvaluator(expression: expression, profile: .formCalculation)
+        guard let map = mapView.map else { return }
```
mhdostal, line 62: The `else` part of this also needs to hide the progress HUD.  Alternatively, you could move the `showProgressHUD` to just before the async `evaluator.evaluate` line.

mhdostal (changes requested):

sbaskaran (approved):

mhdostal (approved):

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -106,6 +108,8 @@ extension QueryFeaturesArcadeExpressionViewController: AGSGeoViewTouchDelegate {
             guard let results = results, let self = self else { return }
             // Get the selected feature.
             if let elements = results.first?.geoElements, let identifiedFeature = elements.first as? AGSArcGISFeature {
+                // Disable user interaction after a feature has been identified.
+                self.mapView.isUserInteractionEnabled = false
```
mhdostal, line 112: Was there a specific issue this is trying to address?

If you're concerned about an "identify" operation happing at the same time as the arcade evaluation, you could move the block of code [here](https://github.com/Esri/arcgis-runtime-samples-ios/blob/fda1fb4edb0cf1d64dfa040052f7e0f0b577f415/arcgis-ios-sdk-samples/Display%20information/Query%20features%20with%20Arcade%20expression/QueryFeaturesArcadeExpressionViewController.swift#L64-L66) (in the `evaluateArcadeInCallout` method) to [here](https://github.com/Esri/arcgis-runtime-samples-ios/blob/fda1fb4edb0cf1d64dfa040052f7e0f0b577f415/arcgis-ios-sdk-samples/Display%20information/Query%20features%20with%20Arcade%20expression/QueryFeaturesArcadeExpressionViewController.swift#L106) (geoView(_ geoView: AGSGeoView, didTapAtScreenPoint)

That way, the arcade evaluation is canceled as soon as the user taps again on the map.

vquach2404: I was trying to address this [issue](https://github.com/Esri/arcgis-runtime-samples-ios/pull/1182#pullrequestreview-1020937384). I thought I resolved it after canceling `evaluateOperation` but I tested it again and am experiencing the same issue. I tried out your suggestion but I'm still seeing the same bug but I realize it's similar logic to disabling user interaction on the map. Do you have any insight to this isse?

mhdostal: It would be helpful to see if the trailing closure for `evaluator.evaluate` was called after a call to `evaluateOperation.cancel()`.  I'm not sure if it's supposed to be called, but it seems reasonable to think that it would be called with a "user cancelled" error.  It it's not it may be an error in the SDK.

Otherwise, you *could* put the call to `UIApplication.shared.hideProgressHUD()` after the call to `evaluateOperation.cancel()`.  That would indicate whether the closure was being called after cancel.

vquach2404: @mhdostal Hmm, the closure for `evaluator.evaluate` is still called after canceling `evaluateOperation`. But isn't that expected since `evaluateOperation` only holds the previous `AGSCancelable` so it wouldn't affect the current evaluation?

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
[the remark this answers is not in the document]
yo1995: It's recommended by Phil and is from the Swift API guidelines: https://www.swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage , "Begin names of factory methods with “make”"

yo1995:
As Gela changed the data, let's use the new data in this sample: 539d93de54c7422f88f69bfac2aebf7d

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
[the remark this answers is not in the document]
sbaskaran: Good to know, thanks.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+class QueryFeaturesArcadeExpressionViewController: UIViewController {
+    @IBOutlet var mapView: AGSMapView! {
+        didSet {
+            mapView.map = makeMap()
+            // Set the touch delegate.
+            mapView.touchDelegate = self
+        }
+    }
+    
+    /// The arcade expression evaluation operation.
```
yo1995, line 27:
```suggestion
    /// The Arcade expression evaluation operation.
```
According to the [terminology](https://csreview.esri.com/terminology-guide) guide it should be capitalized.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+    /// Make and load a map.
+    func makeMap() -> AGSMap {
+        // Create a portal item with the portal and item ID.
+        let portalItem = AGSPortalItem(portal: .arcGISOnline(withLoginRequired: false), itemID: "539d93de54c7422f88f69bfac2aebf7d")
+        // Make a map with the portal item.
+        let map = AGSMap(item: portalItem)
+        return map
+    }
+    
+    /// Evaluate the arcade expression for the selected feature at the map point.
```
yo1995, line 39:
```suggestion
    /// Evaluate the Arcade expression for the selected feature at the map point.
```

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+    func evaluateArcadeInCallout(for feature: AGSArcGISFeature, at mapPoint: AGSPoint) {
+        // Instantiate a string containing the arcade expression.
```
yo1995, line 41:
```suggestion
        // Instantiate a string containing the Arcade expression.
```

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        let expressionValue =
+        """
+        var crimes = FeatureSetByName($map, 'Crime in the last 60 days');
+        return Count(Intersects($feature, crimes));
+        """
+        // Create an arcade expression using the string.
```
yo1995, line 47:
```suggestion
        // Create an Arcade expression using the string.
```

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        let expression = AGSArcadeExpression(expression: expressionValue)
+        // Create an arcade evaluator with the arcade expression and an arcade profile.
```
yo1995, line 49:
```suggestion
        // Create an Arcade evaluator with the Arcade expression and an Arcade profile.
```

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        let evaluator = AGSArcadeEvaluator(expression: expression, profile: .formCalculation)
+        guard let map = mapView.map else { return }
+        let profileVariables = ["$feature": feature, "$map": map]
+        // Show progress hud.
+        UIApplication.shared.showProgressHUD(message: "Evaluating")
+        if let evaluateOperation = evaluateOperation {
+            evaluateOperation.cancel()
+        }
+        // Get the arcade evaluation result given the previously set profile variables.
```
yo1995, line 58:
```suggestion
        // Get the Arcade evaluation result given the previously set profile variables.
```

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        // Dismiss any presenting callout.
+        mapView.callout.dismiss()
+        // Identify features at the tapped location.
+        mapView.identifyLayers(atScreenPoint: screenPoint, tolerance: 12, returnPopupsOnly: false) { [weak self] results, error in
+            guard let self = self else { return }
+            // Get the selected feature.
+            if let elements = results?.first?.geoElements, let identifiedFeature = elements.first as? AGSArcGISFeature {
+                // Disable user interaction after a feature has been identified.
+                self.mapView.isUserInteractionEnabled = false
+                // Evaluate the arcade for the given feature.
```
yo1995, line 104:
```suggestion
                // Evaluate the Arcade for the given feature.
```

yo1995 (commented):
The `AGSArcadeEvaluator.evaluate(withProfileVariables:completion:)` method doesn't call back which results in the [issue](https://github.com/Esri/arcgis-runtime-samples-ios/pull/1182#pullrequestreview-1020937384) mentioned above. 
Shall we add a test in iOS SDK to consecutively tap on a same police beat, say 25 times, to see if it hangs?

Other than that and the UI enable/disable discussion, LGTM.

vquach2404:
> The `AGSArcadeEvaluator.evaluate(withProfileVariables:completion:)` method doesn't call back which results in the [issue](https://github.com/Esri/arcgis-runtime-samples-ios/pull/1182#pullrequestreview-1020937384) mentioned above. Shall we add a test in iOS SDK to consecutively tap on a same police beat, say 25 times, to see if it hangs?

Hi @mhdostal @philium can we get your input about testing this in the SDK?

mhdostal:
> > The `AGSArcadeEvaluator.evaluate(withProfileVariables:completion:)` method doesn't call back which results in the [issue](https://github.com/Esri/arcgis-runtime-samples-ios/pull/1182#pullrequestreview-1020937384) mentioned above. Shall we add a test in iOS SDK to consecutively tap on a same police beat, say 25 times, to see if it hangs?
> 
> Hi @mhdostal @philium can we get your input about testing this in the SDK?

@vquach2404 If you can create a stand-alone bit of code which reproduces the problem, that would be best.  Then create an issue in the runtime/cocoa repo.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        let expression = AGSArcadeExpression(expression: expressionValue)
+        // Create an Arcade evaluator with the Arcade expression and an Arcade profile.
+        let evaluator = AGSArcadeEvaluator(expression: expression, profile: .formCalculation)
+        guard let map = mapView.map else { return }
+        let profileVariables = ["$feature": feature, "$map": map]
+        // Show progress hud.
+        UIApplication.shared.showProgressHUD(message: "Evaluating")
+        if let evaluateOperation = evaluateOperation {
+            evaluateOperation.cancel()
+        }
```
mhdostal, line 57: This can be simplified:
```suggestion
        evaluateOperation?.cancel()
```

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        // Get the Arcade evaluation result given the previously set profile variables.
+        evaluateOperation = evaluator.evaluate(withProfileVariables: profileVariables) { [weak self] result, error in
+            guard let self = self else { return }
+            // Dismiss progress hud.
+            UIApplication.shared.hideProgressHUD()
+            // Enable user interaction.
+            self.mapView.isUserInteractionEnabled = true
```
mhdostal, line 64: Is this really necessary?  If an identify or evaluate operation takes a long time, the user is locked out of doing anything with the mapView.

mhdostal (changes requested):

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
vquach2404: I suppose it's not necessary, but I'm hesitant to make changes at the moment until we figure out the bug with the evaluation operation

yo1995:
> @vquach2404 If you can create a stand-alone bit of code which reproduces the problem, that would be best. Then create an issue in the runtime/cocoa repo.

Created at `cocoa/issues/11185`

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,110 @@
...
+        return Count(Intersects($feature, crimes));
+        """
+        // Create an Arcade expression using the string.
+        let expression = AGSArcadeExpression(expression: expressionValue)
+        // Create an Arcade evaluator with the Arcade expression and an Arcade profile.
+        let evaluator = AGSArcadeEvaluator(expression: expression, profile: .formCalculation)
+        guard let map = mapView.map else { return }
+        let profileVariables = ["$feature": feature, "$map": map]
+        // Show progress hud.
+        UIApplication.shared.showProgressHUD(message: "Evaluating")
```
yo1995, line 54: I'd like to write this part like this

```swift
        evaluateOperation?.cancel()
        // Show progress hud.
        UIApplication.shared.showProgressHUD(message: "Evaluating")
        // Get the Arcade evaluation result given the previously set profile variables.
        evaluateOperation = evaluator.evaluate(withProfileVariables: profileVariables) { [weak self] result, error in
            // Dismiss progress hud.
            UIApplication.shared.hideProgressHUD()
            self.evaluateOperation = nil
```
A few reasons:

1. the progress HUD has nothing to do with `self`, and should be shown and dismissed in an as limited as possible scope. Typically I just put it right above (show) and below (hide) the line of the async method.
2. `AGSCancelable?` doesn't turn into nil automatically after it calls back. To avoid the redundant `cancel` operation, always set it to nil in the callback, regardless it succeeds or fails

yo1995 (approved):

vquach2404:
After discussion with Ting, we've concluded that the bug is not due to an error in the sample. There's not much we can do to fix the issue so we'll be moving on with this PR and putting it into verification.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,107 @@
+// Copyright 2021 Esri
```
saratchandrakarumuri, line 1: Wrong year in copy right header

saratchandrakarumuri (commented):
I see that there is an open issue for the identify getting hung, but for me rather than randomly or eventually, it is happening with every second identify. If it is that bad should we be merging it now?

yo1995:
> I see that there is an open issue for the identify getting hung, but for me rather than randomly or eventually, it is happening with every second identify.

That's quite unexpected... cannot repro at my end. Though I'm getting similar hang-rate, approx. once in 5 evaluations.

> If it is that bad should we be merging it now?

I'm hoping it could get resolved before U15, so I feel we can proceed because it's not a samples problem. I'd like to hear others thoughts though.

vquach2404:
> I see that there is an open issue for the identify getting hung, but for me rather than randomly or eventually, it is happening with every second identify. If it is that bad should we be merging it now?

@saratchandrakarumuri @yo1995  Thanks for your input! I just spoke to Trevor about this. We'll put this PR on pause for now until we discuss this during healthcheck on Friday. We're still aiming to get this sample in for U15 but hopefully this can be resolved before we merge the sample.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+        let evaluator = AGSArcadeEvaluator(expression: expression, profile: .formCalculation)
+        guard let map = mapView.map else { return }
+        let profileVariables = ["$feature": feature, "$map": map]
+        evaluateOperation?.cancel()
+        // Show progress hud.
+        UIApplication.shared.showProgressHUD(message: "Evaluating")
+        // Get the Arcade evaluation result given the previously set profile variables.
+        evaluateOperation = evaluator.evaluate(withProfileVariables: profileVariables) { [weak self] result, error in
+            // Dismiss progress hud.
+            UIApplication.shared.hideProgressHUD()
```
yo1995, line 59:
```suggestion
        // Get the Arcade evaluation result given the previously set profile variables.
        evaluateOperation = evaluator.evaluate(withProfileVariables: profileVariables) { [weak self] result, error in
```
If we gonna proceed with the buggy evaluate operation, then we need to not use the HUD so it doesn't block user interaction, and when the evaluation hangs, nothing is showing in UI level.

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
vquach2404, line 1:
```suggestion
// Copyright 2022 Esri
```

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+        didSet {
+            mapView.map = makeMap()
+            // Set the touch delegate.
+            mapView.touchDelegate = self
+        }
+    }
+    
+    /// The Arcade expression evaluation operation.
+    var evaluateOperation: AGSCancelable?
+
```
yo1995, line 29:
```suggestion
    /// The evaluator to evaluate an `AGSArcadeExpression` under a given
    /// `AGSArcadeProfile`.
    var evaluator: AGSArcadeEvaluator?
```

arcgis-ios-sdk-samples/Display information/Query features with Arcade expression/QueryFeaturesArcadeExpressionViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+    /// Make and load a map.
+    func makeMap() -> AGSMap {
+        // Create a portal item with the portal and item ID.
+        let portalItem = AGSPortalItem(portal: .arcGISOnline(withLoginRequired: false), itemID: "539d93de54c7422f88f69bfac2aebf7d")
+        // Make a map with the portal item.
+        let map = AGSMap(item: portalItem)
+        return map
+    }
+    
+    /// Evaluate the Arcade expression for the selected feature at the map point.
```
yo1995, line 39: The problem is that the evaluator needs to be held a strong reference while evaluating, just like every other loadable resource. My misconception was that the cancelable (`evaluateOperation`) is implicitly holding a reference, which is not. Attached is the code I edited that works now.

```swift
    /// Evaluate the Arcade expression for the selected feature at the map point.
    func evaluateArcadeInCallout(for feature: AGSArcGISFeature, at mapPoint: AGSPoint) {
        guard let map = mapView.map else { return }
        evaluateOperation?.cancel()
        // Instantiate a string containing the Arcade expression.
        let expressionValue =
        """
        var crimes = FeatureSetByName($map, 'Crime in the last 60 days');
        return Count(Intersects($feature, crimes));
        """
        // Create an Arcade expression using the string.
        let expression = AGSArcadeExpression(expression: expressionValue)
        // Create an Arcade evaluator with the Arcade expression and an Arcade profile.
        evaluator = AGSArcadeEvaluator(expression: expression, profile: .formCalculation)
        let profileVariables = ["$feature": feature, "$map": map]
        // Show progress hud.
        UIApplication.shared.showProgressHUD(message: "Evaluating")
        // Get the Arcade evaluation result given the previously set profile variables.
        evaluateOperation = evaluator!.evaluate(withProfileVariables: profileVariables) { [weak self] result, error in
            // Dismiss progress hud.
            UIApplication.shared.hideProgressHUD()
            self.evaluateOperation = nil
            self.evaluator = nil
            if let result = result, let crimeCount = result.cast(to: .string) as? String {
                self.mapView.setViewpointCenter(mapPoint)
                // Hide the accessory button.
                self.mapView.callout.isAccessoryButtonHidden = true
                // Set the detail text.
                self.mapView.callout.detail = String(format: "Crimes in the last 60 days: %@", crimeCount)
                // Prompt the callout at the map point.
                self.mapView.callout.show(for: feature, tapLocation: mapPoint, animated: true)
            } else if let error = error {
                // Present an error if needed.
            }
        }
    }
```

yo1995 (changes requested):

yo1995 (approved):

saratchandrakarumuri (approved):
Works well with the changes now

mhdostal (approved):

vquach2404:
Thanks for your help everyone!

PR #1187, opened 2022-07-29
[Fix] Address Xcode 14 build warnings

yo1995:
## Description

This PR fixes more build warnings provided by Xcode 14 beta 4. The nature of the warnings are similar to Swift SV pull 41.

```
Run script build phase 'Run Script' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase.
```

See similar issue here: https://github.com/realm/SwiftLint/issues/4015

### DIFF
--- a/arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
+++ b/arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
@@ -5293,6 +5293,7 @@
 		};
 		3ED029A21B8E5D4D00ACA70D /* ShellScript */ = {
 			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
 			buildActionMask = 2147483647;
 			files = (
 			);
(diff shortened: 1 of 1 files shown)


philium:
Just to confirm, this change is backwards compatible with Xcode 13?

yo1995:
> Just to confirm, this change is backwards compatible with Xcode 13?

Yes basically it means uncheck this box for a build phase and force the build phase to run on each build. For the 3 occurrences

- We have duplicate detection mechanism for downloading Portal Data, so running the script every time doesn't cause a problem
- README copy can be incremental, but since the build phase doesn't specify the Output Files section, it has to be unchecked. Otherwise Xcode 14 would complain
- Similar for SwiftLint phase - see the linked issue above. Until Xcode allow specifying incremental files, the dependency analysis box needs to be unchecked.

image

philium (approved):

saratchandrakarumuri (approved):

PR #1197, opened 2022-08-12
Update bitcode settings

vquach2404:
This PR addresses the build errors with Xcode 14. See `cocoa/issues/11193` for more info.


yo1995 (approved):
I'm not fully aware of the impact of this change - seems to be required by Xcode 14. 
But it is the only change we need to make in the project file - that is what I approve here. 😅 

Edit: I also suggest marking this PR as draft, as we are not including it in the upcoming release next week, building with Xcode 13.

Edit again: with the latest change to default setting, we can merge it now.

philium:
A couple questions:

1. Did you try using the default value (select the setting and press the Delete key)?
2. Did you try uploading a build to App Store Connect after making these changes?

yo1995 (changes requested):
Pressing delete to go to default value will remove the entries in the project file.

philium:
> shows like this in Xcode UI

@yo1995 Is that in Xcode 13 or 14? With that change, I'm hoping they'll be different, with Xcode 13 being Yes and Xcode 14 being No.

yo1995:
> Is that in Xcode 13 or 14? With that change, I'm hoping they'll be different, with Xcode 13 being Yes and Xcode 14 being No.

That is verified. The screenshot above was Xcode 13. See the comparison below by resetting the bitcode config to default.

|Xcode 13|Xcode 14|
|:-:|:-:|
|image|image|

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -5834,7 +5834,7 @@
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_TEAM = "";
 				EMBED_ASSET_PACKS_IN_PRODUCT_BUNDLE = YES;
-				ENABLE_BITCODE = YES;
+				ENABLE_BITCODE = NO;
```
yo1995, line 5837:
```suggestion
```

arcgis-ios-sdk-samples.xcodeproj/project.pbxproj
```diff
@@ -5863,7 +5863,7 @@
...
```
yo1995, line 5866:
```suggestion
```

saratchandrakarumuri (dismissed):

vquach2404:
> 2. Did you try uploading a build to App Store Connect after making these changes?

To answer your other question, I ran into no errors with bitcode set to "No" (Xcode 13.4) while uploading a build to AppStore Connect.

yo1995 (approved):

dg0yal:
> Edit: I also suggest marking this PR as draft, as we are not including it in the upcoming release next week, building with Xcode 13.

This change should not hurt Xcode 13
We should get this into the U15 release so that when users upgrade to xcode 14 in the fall things will continue to work

dg0yal (approved):

